PR #1, opened 2019-05-22
Enhance list API and add support for icon

ruflin:
In this PR the list API was enhanced and now looks as following:

```
[
  {
    "description": "This is the envoyproxy integration.",
    "icon": "/img/envoyproxy-0.0.2.png",
    "name": "envoyproxy",
    "version": "0.0.2"
  },
  {
    "description": "This is the envoyproxy integration with improved features.",
    "icon": "/img/envoyproxy-0.0.5.png",
    "name": "envoyproxy",
    "version": "0.0.5"
  }
]
```

How also the description and the icon path are exposed. The icon will be important to be used to serve the overview page to not have to download all the packages directly.

Each package has an icon path even if it does not contain an icon. In this case, the API returns a default icon. The path shown in the data above can be used to fetch the image.

An additional package `envoyproxy-0.0.5` was added as an example. It's identical to the version 0.0.2 but has a different description and an icon inside.


ruflin:
@jfsiii Opening this as a PR for visibility. Let me know what you think and what additional info we needed.

PR #10, opened 2019-06-03
Add support for multiple images per integration

ruflin:
Until now it was only possible to have an `icon.png` in a package. But having documentation with dashboards example in the future, multiple images per integrations are needed. This adds support for `.png` and `.jpg` images per package.

The path to the images changes because of this and is now `/img/{integration-name}-{version}/{img}`, before it was `/img/{integration-name}-{version}.png`

As an example:

* Before: `/img/envoyproxy-0.0.5.png`
* After: `/img/envoyproxy-0.0.5/icon.png`

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,7 +44,7 @@ func main() {
 	router.HandleFunc("/list", listHandler())
 	router.HandleFunc("/package/{name}", packageHandler())
 	router.HandleFunc("/package/{name}/get", downloadHandler)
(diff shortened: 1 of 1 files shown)


ruflin:
@jfsiii This is a breaking change assuming you hardcoded the previous path. If you took the path already from the `/list` API endpoint you are fine. If you didn't use images at all yet, also no issue.

PR #12, opened 2019-06-04
Make address of service configurable

ruflin:
Before the service was always exposed on `localhost:8080`. This default stays but with this change it is possible to also specify a different address / port:

```
 go run main.go -address="localhost:8081"
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,7 +34,6 @@ func init() {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


ruflin:
jenkins, test this

ruflin:
jenkins, test this

PR #13, opened 2019-06-05
Add support for .tar.gz files and change download path

ruflin:
Previously only .zip files were exposed. Now also .tar.gz files are exposed. It should be decided if long term we only need one format but this will allow us to play around with both formats for now.

This commit also changes the download API. Previously it was: `/package/{package}/get`. This is now changed to `/package/{package}.tar.gz` and `/package/{package}.zip`. It feels more logical to me to have the suffix already in the URL.

To make such changes less breaking for the UI in the future, the download link is also added to the list output:

```
[
  {
    "description": "This is the envoyproxy integration.",
    "download": "/package/envoyproxy-0.0.2.tar.gz",
    "icon": "/img/envoyproxy-0.0.2.png",
    "name": "envoyproxy",
    "version": "0.0.2"
  },
  {
    "description": "This is the envoyproxy integration with improved features.",
    "download": "/package/envoyproxy-0.0.5.tar.gz",
    "icon": "/img/envoyproxy-0.0.5.png",
    "name": "envoyproxy",
    "version": "0.0.5"
  }
]
```

It seems some of the information that is above in the list package output should also be in the detail package output. I will tackle this later.


ruflin:
@jfsiii Let me know what you think about these API changes.

jfsiii:
The URL changes LGTM. Nothing too controversial in supporting multiple formats and giving each a URL. However, if the Manager needs knowledge of how/when to use .gz vs .zip I don't know if the single value in `download` is that helpful. 

If the Manager uses .zip in some situations (for Windows, etc) and .tar.gz for others, it can't blindly follow the `download` value (`"/package/envoyproxy-0.0.5.tar.gz"`). It'll use one of the paths we've defined, effectively ignoring `download`.

ruflin:
I don't think we should long term support 2 formats (if not really needed). This is more temporary to play around with both options. And as soon as we decided on one, we should remove the other and have only one left.

I'm currently leaning in the direction of using `.tar.gz` only. Could you see us hitting issues with it in the future?

jfsiii:
I think the issue with `.tar.gz` is Windows requires 3rd party software to extract it. Hopefully, we can avoid that issue because we're extracting the archive instead of asking the user to do it.

I'm all for supporting both to explore and reducing down to one (`.tar.gz`?) if we can.

re: JS advantages for `.tar.gz` vs `.zip`, I was just saying that you can't unpack `.zip` via _stream_ because the Central Directory is _at the end of the file_. So you need to get the entire file (buffered in memory or saved to disk) before you start extracting. There's some more info at https://github.com/thejoshwolfe/yauzl#no-streaming-unzip-api & https://nervous.io/node/javascript/2017/02/06/npm-unzip/

I don't think that's an issue yet, though. The archives I've seen so far are small enough to hold in memory and we only need them in memory while we're queuing them to be processed/uploaded.

:shipit:

ruflin:
Thanks for the detail. I suggest we go down the following path:

* For now we support both to download
* We stick to `.tar.gz` for download (the one we link) as we both expect that this is the final format.

PR #14, opened 2019-06-07
Add generated example docs for the API

ruflin:
To make working with the API easier and track changes to the API docs, example API docs are generated as part of the tests and compared. The example can be found under `/docs/api`.

To update the docs after and API change, `go test . -generate` needs to be run.


ruflin:
@jfsiii I hope these example docs make it easier to work with the API and also make it very obvious what changed if we modify the output.

jfsiii:
:+1:

PR #15, opened 2019-06-07
Introduce package struct to align API calls output

ruflin:
To make sure the different API calls which expose package information (/list, /package/*) have the same content / naming, the package object is introduced.

The icon path is added to the detail output to have in the detail view all the information which is also on the top level.

Moved handler and to it's own file for better readability. Server now must be started with `go run .` instead of `go run main.go`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,14 +7,10 @@ package main
 import (
 	"archive/zip"
 	"bytes"
-	"encoding/json"
 	"flag"
-	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"net/http"
-	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


ruflin:
@jfsiii I think the important change for you in this PR is that you need to run now `go run .` instead of `go run main.go` to start the service.

main.go
```diff
@@ -207,10 +68,19 @@ func getIntegrationPackages() ([]string, error) {
...
+	Icon        string `yaml:"icon" json:"icon"`
+}
+
+func (p *Package) getIcon() string {
+	return "/img/" + p.Name + "-" + p.Version + "/icon.png"
+}
+
+type Manifest struct {
+	Package     `yaml:",inline" json:",inline"`
 	Requirement struct {
```
jfsiii: Will the Integrations Manager need the `Requirement` info when it calls `/list` (i.e. shows the Integrations list view in Kibana)?

I imagine the manager will either ask for "a list of available integrations which meet this criteria" (`/list?kibanaMin=X&kibanaMax=Y`) or it will ask for "a list of available integrations" (`/list`) and doing work on the client side to filter or indicate to the user it their current setup is compatible.

Right now, we need to call `/package/{name}` for each package to find out if it met the criteria.

Ideally, from Integrations Manager POV, if `/package/{name}` returns `Package` type, then `/list` returns `Package[]` or `Array<Package>`

ruflin: Good point that already in list we need to be able to query / filter on the requirements. Could you open an issue related to this in this repo?

As I expect the information inside `Package` to grow I would suggest `[]Package(subset)` that the list does not contain all the info we get when doing to `/package/{name}` request.

jfsiii: > As I expect the information inside Package to grow I would suggest []Package(subset) that the list does not contain all the info we get when doing to /package/{name} request.

SGTM. That was my inclination as well. We'll iterate on the types in https://github.com/elastic/kibana/issues/38598 and https://github.com/elastic/kibana/blob/feature-integrations-manager/x-pack/plugins/integrations_manager/common/types.ts

ruflin: Thanks for opening the testing issue, let's continue the conversation there. Are you ok with merging as is for now?

PR #17, opened 2019-06-12
Make running as container work by default

ruflin:
The default address the service is listening on is `localhost:8080`. Because of this when running inside a docker container, it cant be accessed. The Dockerfile is updated to expose it by default.


jfsiii:
Tried these changes locally and works for me. :shipit:

PR #20, opened 2019-06-14
Add docs around all supported assets

ruflin:
The ASSETS.md file was added to the docs to document each asset which can exist in a package and its content with the potential issues surrounding it.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -1,0 +1,163 @@
+# Integration package assets
+
+This doc is to describe all the assets available in the integration packages and some details around each asset.
+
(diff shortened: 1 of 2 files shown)


ruflin:
@jfsiii @jasonrhodes I created this documentation to have single source of truths around which assets should exist under which path in a package and on what some of the gotchas around these assets are. Please have a look. I suggest to merge this soonish and then keep iterating on top of it by adding more details when needed.

@roncohen FYI

ruflin:
I merged this to be able to add more details in follow up PR's.

PR #23, opened 2019-06-17
Add link for example registry

ruflin:
And example registry based on the Docker container built from master is running now under http://integrations-registry.app.elstc.co/

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ There are two options to run this. Either the service can be run as a go command
 `go run .`
 
 ### Docker
+**Example**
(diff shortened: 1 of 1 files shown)


ruflin:
@jasonrhodes @jfsiii @roncohen This should simplify testing. 

@jfsiii We should also add something around this to our Kibana README?

jfsiii:
@ruflin I deleted that step in https://github.com/elastic/kibana/pull/39040/commits/4833088fff7daee11cf4fb4df248387a9e6ac3f6

Can add directions back if the registry is flaky, or running a local registry becomes a more common flow.

PR #24, opened 2019-06-17
Add data frame transforms to Asset description

ruflin:


ASSETS.md
```diff
@@ -108,6 +108,12 @@ Elasticsearch [Machine Learning Jobs](https://www.elastic.co/guide/en/elasticsea
 can be created in Elasticsearch assuming ML is enabled. As soon as a job is started, the job creates results. If results
 are around, a Job can't be just removed anymore but also the results must be removed first (more details needed).
 
+### Data Frames
+
+* Asset Path: `elasticsearch/data-frame/*.json`
```
ruflin: @andrewvc I wonder if this should be `data-frame-transform`? Could there be other objects falling under `data-frame` in the future?

andrewvc: Good point, yes. In the [ES API](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/put-data-frame-transform.html) it's `_data_frame/transforms`, so maybe we should separate it with a slash?

ruflin: separating with a slash would mean one more directory. I would prefer `data-frame-transform`. I wonder if there are any other objects planned then transform.

andrewvc: I'm fine with no slash. Not a big difference to me.

ruflin: Ok, lets go with this for now. I'm still not clear "what else" is coming for data frames besides transform in the future.

PR #28, opened 2019-06-19
Only expose most recent integrations version in /list

ruflin:
Until now under `/list` all integrations were exposed, meaning also all versions of an integration. For mysql to versions were exposed. With this change, only the most recent version will be exposed. See https://github.com/elastic/integrations-registry/issues/26 for more discussions.

### DIFF
--- a/docs/api/list.json
+++ b/docs/api/list.json
@@ -1,12 +1,5 @@
 [
   {
-    "description": "This is the example integration.",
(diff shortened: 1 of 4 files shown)


ruflin:
@jfsiii Here is the first change for https://github.com/elastic/integrations-registry/issues/26 I assume this should not have any affect on the integrations manager.

PR #29, opened 2019-06-19
Enhanced example manifest

ruflin:
Extending the manifest used in the testing package as an example.


ruflin:
@hbharding This is also some information we might be able to show in the UI.

PR #30, opened 2019-06-19
Allow in /list API to filter by Kibana version

ruflin:
With this change, it's possible to query for integration packages which only match the requirement of a specific Kibana version. The query will look as following:

```
/list?kibana=6.5.2
```

This returns all packages which are compatible with Kibana 6.5.2 based on their manifest.yml definition.

Additional changes:

* Support for passing `PACKAGES_PATH` to build package command to also rebuild testing packages if needed
* Update example package requirements


main_test.go
```diff
@@ -32,6 +32,8 @@ func TestEndpoints(t *testing.T) {
 	}{
 		{"/", "/", "info.json", infoHandler()},
 		{"/list", "/list", "list.json", listHandler()},
+		{"/list?kibana=6.5.2", "/list", "list-kibana652.json", listHandler()},
```
ruflin: @jfsiii This PR allows this kind of queries. No ranges or similar yet.

ruflin:
@jfsiii Merged already as I need it for other PRs. But please comment here if this should work differently.

jfsiii:
@ruflin lgtm, thanks

PR #32, opened 2019-06-21
Allow to configure the service through config file

ruflin:
A config.yml file is added to the service which can be used moving forward to configure the service.

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,7 +9,6 @@ import (
 	"bytes"
 	"context"
 	"flag"
-	"fmt"
 	"io"
 	"log"
 	"net/http"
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -33,11 +36,31 @@ func init() {
...
+	PackagesPath string `config:"packages.path"`
+	Foo          string `config:"foo"`
+}
+
 func main() {
 	flag.Parse()
 	log.Println("Integrations registry started.")
 	defer log.Println("Integrations registry stopped.")
 
+	cfg, err := ucfgYAML.NewConfigWithFile("config.yml")
```
ruflin: @urso For some reason this line stopped working for the `.` notations if I used `ucfg.PathSep(".")` as option. Did not investigate but somehow assume with the option is should work.

PR #34, opened 2019-06-24
Add integration API endpoint to list packages

ruflin:
Based on the discussion in https://github.com/elastic/integrations-registry/issues/26 it became obvious that there is also API endpoint needed which list all packages for a specific version. This could have been done in the `/list` API endpoint but I did not like this too much as by default only the newest packages are exposed of an integration and then if an integration name is given, all of this integration? This does not seem to be intuitive.

Instead I propose an `/integration/{name}` endpoint which exposes all the package of an integration. The output for `http://localhost:8080/integration/mysql` looks as following:

```
{
  "integration": "mysql",
  "packages": [
    {
      "name": "mysql",
      "version": "1.0.1",
      "description": "Mysql Logs and metrics",
      "icon": "/img/mysql-1.0.1/icon.png",
      "requirement": {
        "kibana": {
          "min": "",
          "max": ""
        }
      }
    },
    {
      "name": "mysql",
      "version": "1.0.2",
      "description": "Mysql Logs and metrics",
      "icon": "/img/mysql-1.0.2/icon.png",
      "requirement": {
        "kibana": {
          "min": "",
          "max": ""
        }
      }
    }
  ]
}
```

If wonder if long term there could also be additional info around an integration inside here? Or should the description of the integration be taken from the most recent integration?

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -11,6 +11,9 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"path/filepath"
+
+	"github.com/elastic/beats/libbeat/common"
 
 	"github.com/blang/semver"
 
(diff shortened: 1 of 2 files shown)


ruflin:
@jfsiii Would be great to get your thoughts on this one.

jfsiii:
I was expecting this to be under `/list` and bias to fewer endpoints but this is fine as well. Or at least no reason to block shipping. We’ll iterate and see.

Mapping out how I thought of `/list`:
 * is a search endpoint and will always a list of integrations. 
 * Without arguments, it returns all integrations, filtered to their most recent version
 * Given an argument for integration name it would return all results for that integration. They supply another argument (e.g. `semver=latest`) if they want to further restrict the results.

Again, happy to iterate with `/integration`, but I don’t think there’s an issue using `/list`

ruflin:
Agree on fewer API endpoints. The part that bothers me is really that a filter can lead to more results then without the filter. Let's assume we have 4 versions of apache and 1 version of mysql. Now fetching `/list` will return 2 results, 1 for apache, 1 for mysql. Fetching now `/list?integration=apache` will return 4 results, 1 for each apache version. This sound counter intuitive to me as I would expect filters to reduce the number of responds.

Perhaps we could add one more query param, something like `all=true` which is off by default. So if someone queries for `integration=apache` he would get 1 result by default, but would have to add `integration=apache&all=true` to get all versions. Or the other way around we could say by default always all results are returned, by when only wanting the latests, the user adds `version=latest`.

As you see, I'm on the fence :-)

ruflin:
I had a quick chat with @jfsiii about this. The main conclusion is, this is something we don't need right now for the integrations manager and is more something to test the integrations registry. I still plan to add it to `/list` endpoint as suggested by @jfsiii in the near term to have it around for testing.

ruflin:
Closing as not needed at the moment.

PR #38, opened 2019-06-25
Add release tag and compatibilty to manifest

ruflin:
The manifest can also contain compatiblity and release information. With this PR this information is added to the example manifest file.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -24,6 +24,10 @@ name: envoyproxy
 description: This is the envoyproxy integration.
 version: 0.0.2
 categories: ["logs", "metrics"]
+# Options are experimental, beta, ga
+release: beta
+compatibility: [1.0.2, 2.0.1]
+
 requirement:
   elasticsearch:
     version.min: 7.0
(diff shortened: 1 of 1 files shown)


ASSETS.md
```diff
@@ -24,6 +24,10 @@ name: envoyproxy
 description: This is the envoyproxy integration.
 version: 0.0.2
 categories: ["logs", "metrics"]
+# Options are experimental, beta, ga
```
ruflin: @hbharding FYI

ASSETS.md
```diff
@@ -38,6 +42,8 @@ requirement:
 
 The definition of the manifest is not complete yet and further details will follow.
 
+Question: How do we handle if only one metricset is in beta but integration is in GA?
```
ruflin: @exekias Not to be answered in this PR, but interested to hear your thoughts.

exekias: that's a good question, I guess we should not add a beta metricset to defaults? It could also be interesting to flag it in the provided configuration (or even the UI)

ruflin: Good idea with the defaults. So it's documented with the specific input that it's in beta so it's going to show up in the UI when the user wants to add it.

PR #39, opened 2019-06-26
Add download stats

ruflin:
Implements a first version of https://github.com/elastic/integrations-registry/issues/31. For now it is logged to the stdout by default and if configured and enabled, also to Elasticsearch.

The mapping for the Elasticsearch index is added under `/dev/stats`. It has to be loaded manually into Elasticsearch.

TODO:

* [ ] Implement config option to enable / disable sending data to Elasticsearch cluster and configuring cluster.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,1 +1,3 @@
 packages.path: "./build/packages"
+
+stats.elasticsearch: ["http://localhost:9200"]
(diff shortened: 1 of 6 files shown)


ruflin:
Currently blocked on https://github.com/elastic/integrations-registry/pull/32 as this also needs the config options.

ruflin:
Closing this for now as it is expected that we get the download stats directly from the statically hosted files on GCP or AWS. These logs we then need to process and it would be nice to still end up with the same structure as proposed here.

PR #40, opened 2019-06-26
Add definitions to the assets

ruflin:
The goal is to have a versioned place with a definiton on what an input, data source etc. are.


ruflin:
Merged for now so we can iterate on it.

mattapperson:
Sorry, I must have missed the request for review on this

PR #47, opened 2019-08-14
Add support for title in package manifest

ruflin:
In the `manifest.yml`, `title: foo` can be set to differentiate from the name. If a title is not set, it is populated with the name param.

Closes https://github.com/elastic/integrations-registry/issues/43


ruflin:
@jfsiii I initially thought best is to keep `title` populated with the name by default if it is empty. But now I have second thoughts and thinking to rather keep it empty if not set. This will allow the UI to make the decision if it should default to the name or apply additional magic on top. If it is set, the UI will not know if it was generated or is actually set. WDYT?

jfsiii:
@ruflin yeah, I think empty is best

docs/api/list-kibana652.json
```diff
@@ -4,6 +4,7 @@
     "download": "/package/example-0.0.5.tar.gz",
     "icon": "/img/example-0.0.5/icon.png",
     "name": "example",
+    "title": "",
```
ruflin: @jfsiii This is the way it will look like at the moment if the title is not set. The field is in the response but with an empty string. Is that what you expected or better the entry does not exist at all?

ruflin:
@jfsiii Merged for now to use this in other integrations so we can start testing it, please still comment on the above question, undecided on the best approach here.

docs/api/list-kibana652.json
jfsiii: My preferences are

1. No `title` entry
1. `title` value of `undefined`
1. `title` value of `""` (empty string)

but any of these is workable

ruflin: Option 1 it is: https://github.com/elastic/integrations-registry/pull/50

jfsiii: The `/list` and `/package` endpoint have slightly different behaviors.

The first result from `curl http://integrations-registry.app.elstc.co/list` has no `title` key:

```json
  {
    "description": "Mysql Logs and metrics",
    "download": "/package/mysql-1.0.2.tar.gz",
    "icon": "/img/mysql-1.0.2/icon.png",
    "name": "mysql",
    "version": "1.0.2"
  },
```


but `curl http://integrations-registry.app.elstc.co/package/mysql-1.0.2` has `title` key and `null` value:

```json
{
  "name": "mysql",
  "title": null,
  "version": "1.0.2",
  "description": "Mysql Logs and metrics",
  "icon": "/img/mysql-1.0.2/icon.png",
  "requirement": {
    "kibana": {
      "version.min": "6.7.0",
      "version.max": "7.6.0"
    }
  }
}
```

jfsiii: Related: I think many of the titles should be updated (e.g. "MySQL" , "PostgreSQL", "HAProxy", "iptables", etc). Should I open an issue, comment here, or something else?

ruflin: Comment here works, will fix it.

ruflin: Here is the fix: https://github.com/elastic/integrations-registry/pull/60 I left out some titles on purpose so you can test on how it works without ;-)

PR #48, opened 2019-08-15
Add CoreDNS example integration

ruflin:
This integration takes the assets from the coredns filebeat and metricbeat module and merges them together.

A few things are special new here:

* It has a long README which we can use to test on what happens with long README's.
* The ingest pipelines contain template parts as they need to be completed during insertion.
* It contains also the module configs for Filebeat and Metricbeat. Not clear if we will need it but probably will have to be removed later again. Right now they are here so we can play around with them.
* Dashboards are for Logs and Metrics.
* Contains a png and an svg icon.


dev/package-examples/coredns-1.0.1/elasticsearch/ingest-pipeline/pipeline-entry.json
```diff
@@ -0,0 +1,101 @@
+{
+  "description": "Pipeline for normalizing Kubernetes coredns logs",
+  "processors": [
+    {
+      "pipeline": {
+          "if": "ctx.message.charAt(0) == (char)(\"{\")",
+          "name": "{< IngestPipeline "pipeline-json" >}"
```
ruflin: @jfsiii It will be interesting to see how we handle these :-) I have a few ideas / suggestions, but lets discuss separately.

jfsiii: Sure thing, just let me know. 

The first obstacle is to make these valid JSON documents. As is, we can't load them to scan for `processors` or `pipeline`.

ruflin: Sounds like we have a chicken / egg problem here then. We need some template language here to indicate it will need replacing and also with what (the name that is given to the ingest pipeline with the name `pipeline-json`). Any suggestions on the template language?

jfsiii: The immediate issue is the double-quotes within double-quotes.

If we escaped them:
```diff
- "name": "{< IngestPipeline "pipeline-json" >}"
+ "name": "{< IngestPipeline \"pipeline-json\" >}"
```

or used single quotes:
```diff
- "name": "{< IngestPipeline "pipeline-json" >}"
+ "name": "{< IngestPipeline 'pipeline-json' >}"
```

or dropped them entirely?:
```diff
- "name": "{< IngestPipeline "pipeline-json" >}"
+ "name": "{< IngestPipeline pipeline-json >}"
```

we'd have a document that we could parse.

From there we can move on to extracting the info. I'm fine with starting with the current "start_token type id end_token" (or whatever it represents) and iterating from there. Especially if it's based on a pattern we use elsewhere and want to keep.

ruflin: The above is what we have today in Filebeat and the conversion is done by Golang. But moving forward we should focus on what is best for JS / TypeScript. It would be nice to use an existing template language. One thing to keep in mind here is that `mustache` is already used by Elasticsearch: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#search-template So if we want to support search templates as part of a package one day, we should not conflict with it.

Happy to modify the ingest pipelines to any of the above proposals for now, wich one is your favorite?

jfsiii: > It would be nice to use an existing template language

💯 I'm not interested in creating a new templating language

> One thing to keep in mind here is that [mustache is already used by Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#search-template):  So if we want to support search templates as part of a package one day, we should not conflict with it.

That's good to know. Let's start a discussion issue or doc which lists the requirements so we can pick an appropriate solution, but I'm fine with Mustache being the default or null hypothesis.

> Happy to modify the ingest pipelines to any of the above proposals for now, which one is your favorite?

Probably the first (escaped quotes), because it's the syntactically correct version of what it seems like you intended to do. If you're trying to encode `{< IngestPipeline "pipeline-json" >}` as JSON, that's the correct output.

ruflin: Discussion issue opened: https://github.com/elastic/integrations-registry/issues/54

ruflin: Opened a PR with the escaping: https://github.com/elastic/integrations-registry/pull/55 What is in here so far is a 1-1 copy from Beats, so whatever we come up with in #54 it will need to be taken into account in move from modules to integrations.

Issue #45, opened 2019-08-06
Discuss icon response restrictions


jfsiii:
## Current
`icon` is a string which returns a relative path to an image (or possibly any asset type?)

## Challenges include
1. No guidance or consistency on image dimensions, format, resolution (portrait, 32x32, image-name@2x, etc)
1. Missing support for the Elastic and Third Party logos from [`EuiIcon`](https://elastic.github.io/eui/#/display/icons) which are the majority of the initial integrations.
1. Issues using `<img>` tags with [`EuiCard`](https://elastic.github.io/eui/#/display/card) (can't use with Horizontal Layouts, essentially forces a single layout/appearance, etc).

## Options include
1. Maintaining status quo
1. Convert `icon` to values compatible with `EuiIcon` (e.g. enum of icon names, string URL to SVG, or string of inline SVG)
1. Add a separate entry or otherwise make a distinction between raster images and SVG (e.g. `iconUrl` & `iconType`)

jfsiii:
I don't think Option 1 is really a viable choice.

Whether we go for 2, 3 or another one, I think we should prioritize (if not exclusively support) SVG. They are quite common and eliminate several categories of challenges with raster images. 

I also think it's consistent with our progressive/iterative approach to support the needs/resources we have now (svg asset icons) and think about other cases later.

ruflin:
+1 on having SVG images as it will eliminate the scale discussion (I hope). Are there downsides of only supporting SVGs?

jfsiii:
The first one which occurs to me is that not every project has an SVG asset. This is more true when we eventually open up to outside contributions.

jfsiii:
We could follow the example from the [Web App Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) have an [`icons`](https://developer.mozilla.org/en-US/docs/Web/Manifest/icons) section which looks like:

```
"icons": [
  {
    "src": "/img/icon.svg",
    "sizes": "16x16 32x32 96x96 128x128 256x256"
   },
  {
    "src": "/img/icon-192.png",
    "type": "image/png",
    "sizes": "192x192"
  },
  {
    "src": "/img/icon-512.png",
    "type": "image/png",
    "sizes": "512x512"
  }
]
```

The `src` values don't matter here, they're just the path to the asset. `type` and `sizes` give the information about it.

[...]

PR #49, opened 2019-08-15
Proposal: List screenshots in manifest

ruflin:
Discussing with Henry some of the designs, it turned out that we will need in the future support for multiple screenshots, potentially even quick intro videos. As these screenshots should also have a description, I propose the following addition to the `manifest.yml` file:

```
screenshots:
- name: overview-logs.png
  title: This shows the overview of the logs dashboard.
- name: overview-metrics.png
  title: Metrics Dashboard.
- name: getting-started.mp4
  title: Getting started with the envoyproxy integration.
```

All the screenshots must be listed in here and will be shown in the order listed. Videos can also be added and the description is in here. The path to each screenshot / video will be based on a convention.

I don't suggest we add a list of screenshots or videos in our first iteration but just have one, but already set the technical foundation for it. This means for now our implementation would look as following:

```
screenshots:
- name: overview-logs.png
  title: This shows the overview of the logs dashboard.
```

This PR does not change yet the existing integrations and implementation of the registry, only the example manifest.


jfsiii:
This also has Challenge 1 from #45. There's no information about dimensions, orientation, etc.

The Web App Manifest I mention in #45 also has a [`screenshots`](https://developer.mozilla.org/en-US/docs/Web/Manifest/screenshots) member which would help address those issues. That happens to be JSON, but the data structure could be YAML.

If we avoid an explicit contract, we should at least provide guidance about what we expect. e.g. landscape orientation, suitable for viewing at WxH, etc.

ruflin:
What happens with the images when someone looks at it on a mobile device for example? Do we expect the integrations to have 2 different sized images? Do we just scale down?

I wonder if we should even have some validation of packages to make sure the images are in the right size to not break the UI. Or is this not something we should worry about?

ruflin:
One other thing that came to my mind related to this change is that multiple images + videos could make the package pretty large. To solve this I would suggest that the default package that is downloaded only contains the icons and not the screenshots / videos (or only 1) and all the others are streamed from the registry by default. To download a complete package to perhaps modify it, someone would have to add a param like `?full=true`.

jfsiii:
> What happens with the images when someone looks at it on a mobile device for example? Do we expect the integrations to have 2 different sized images? Do we just scale down?

I don't think multiple images are a requirement. Worst case we could scale down. However, it _is_ an argument for specifying dimensions in the manifest. The consumer can filter the list by images which are appropriate.

> To download a complete package to perhaps modify it, someone would have to add a param like ?full=true.

Agreed. 

If it gets to where we need more granular control, it's common to use `?include=x,y` or `?exclude=a,b`

ruflin:
So our above discussion translates to the following for the YAML part:
```
screenshots:
- src: /img/overview-logs.png
  title: This shows the overview of the logs dashboard.
  size: 10x10
  type: foo
```

Correct?

ASSETS.md
```diff
@@ -39,6 +39,26 @@ requirement:
...
+screenshots:
+  # The src path is relative from inside the package. The full path will be generated by the server
+  # and exposed through the API.
+- src: /img/overview-logs.png
+  title: This shows the overview of the logs dashboard.
+  # The type does not have to be set explicitly if it's in the file extesions
+  # but the server will extract it an expose it through the API
+  type: image/png
+  # The size of the image could be detect by the server too if needed.
+  # We must come up with a recommended image size.
```
ruflin: @hbharding Do we already have a suggest size for this?

ruflin:
@jfsiii @hbharding Could you have an other look at the suggested definition. Will follow up with a PR for the API endpoint if we agree on the above.

ASSETS.md
```diff
@@ -39,6 +39,26 @@ requirement:
...
+  size: 800x600
+- src: /img/overview-metrics.jpg
+  title: Metrics Dashboard.
+  size: 800x600
+- src: getting-started.mp4
+  title: Getting started with the envoyproxy integration.
+  size: 800x600
```
ruflin: @jfsiii For icons we discuss to make this an array. I would suggest for screenshots we don't for now.

ASSETS.md
```diff
@@ -39,6 +39,26 @@ requirement:
     version.min: 7.1
   filebeat:
     version.min: 7.2
+
+# The order of the items listed here is the order they show up in the package overview.
+screenshots:
+  # The src path is relative from inside the package. The full path will be generated by the server
```
jfsiii: 👍 Not sure if this is a different from the `icon` behavior but I like this explicit statement. Good experience for the integration authors and registry API consumers.

ASSETS.md
```diff
@@ -39,6 +39,26 @@ requirement:
...
+  # and exposed through the API.
+- src: /img/overview-logs.png
+  title: This shows the overview of the logs dashboard.
+  # The type does not have to be set explicitly if it's in the file extesions
+  # but the server will extract it an expose it through the API
```
jfsiii: 👍 Nice to see the API will include this. The UI might not need it now or every time, but it will make things much easier when it is needed.

ruflin: Potentially we can skip it in the first version until we have a consumer. But good to have it in the definition.

ASSETS.md
ruflin: We should make the behaviour in the icon the same I think.

PR #52, opened 2019-08-19
Extract packages on server startup

ruflin:
So far files were read from the `.tar.gz` packages which made the implementations more complex. Instead, all the packages are extracted on startup. This will slow down start up time a bit but afterwards will allow to make all queries directly file based which should speed up things and especially simplifies the implementations.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,14 +5,13 @@
 package main
 
 import (
-	"archive/zip"
-	"bytes"
 	"context"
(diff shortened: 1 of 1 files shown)


ruflin:
With https://github.com/elastic/integrations-registry/issues/53 we potentially move to a fully file based approach. Merging this as it simplifies the existing code.

PR #57, opened 2019-08-21
Rename /list endpoint to /search

ruflin:
In recent discussions it became more obvious to me that our current `/list` endpoint is more a search endpoint for packages. It returns a list of packages based on the search query params. At the moment these queries are versions of the different services but long term it could also be a text search. By default the endpoint returns the list of all packages with the most recent version.

This is a breaking change!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 Endpoints:
 
 * `/`: Info about the registry
-* `/list`: Lis of all available integration packages
+* `/search`: Search for packages. By default returns all the most recent packages available.
 * `/package/{name}`: Info about a package
 * `/package/{name}.tar.gz`: Download a package
 
--- a/handler.go
+++ b/handler.go
@@ -140,7 +140,7 @@ func imgHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprint(w, string(img))
 }
 
-func listHandler() func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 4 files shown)


ruflin:
@jfsiii This is a proposal to make it easier to understand what this endpoint is for.

jfsiii:
:+1: As I mentioned in https://github.com/elastic/integrations-registry/pull/34#issuecomment-504995057 I see it as a search endpoint as well, so I like the name change

The "breaking change" note reminds me that `/` still returns `{"version": "0.0.1"}`. I'll open another ticket to discuss versioning.

I'm fine with this merging whenever you feel and I'll update the manager route

ruflin:
@jfsiii Merged and will deploy it shortly. Would be great if you could do a PR to return functionality.

For the version: +1 on an issue. Keeping it at 0.1.0 at the moment as I would say we are not "stable" yet.

jfsiii:
agreed on not stable but that just means it stays in the 0 major; not that versions don't get updated.

I see it's been deployed. Opening a branch now and will cc: you on it.

jfsiii:
@ruflin I think this means https://github.com/elastic/integrations-registry/blob/master/docs/api/list.json should become https://github.com/elastic/integrations-registry/blob/master/docs/api/search.json

ruflin:
@jfsiii Will take care of renaming these files.

Issue #53, opened 2019-08-21
Scaling the integrations registry service


ruflin:
Long term the availability of the integrations registry service becomes critical for each Kibana installation. This raises the question on how to make sure the registry is always available and easy to scale.

Large parts of the existing API endpoints are static. Data for an existing package never changes and all the package related API endpoints never change. This makes it possible instead of running a service which creates the responses on demand the pregenerate all the content and serve it from static files.

The `/list` endpoint is an exception to the above. This endpoints allows to query the existing packages based on different factors like versions of the components in the stack, existing plugins or even search queries in the future. This leads to millions or billions of different possible combinations. 

# Proposal

The proposal is to pregenerate all the content for the packages and package API endpoints. For the `/list` endpoint an actual service needs to be running that serves the on demand content.

roncohen:
As you indicate, for the all endpoints that are essentially static, i don't think there's much to discuss when it comes to scalability. For the `/list` (soon `/search` https://github.com/elastic/integrations-registry/pull/57) endpoint, i think it's a good question how it will work and how the registry will be deployed/hosted.

It's important to provide high quality search. Letting users optionally use Elasticsearch to back the registry would be interesting.

roncohen:
To address your concrete proposal, I'm not sure we need to generate static content right now. 

Asking people to run a simple go service doesn't seem like a problem, but i could be missing something.

ruflin:
I started to play around with static content for some parts and it actually simplifies the code: https://github.com/elastic/integrations-registry/pull/58 For now it's still a go service that serves the static and the `/search` endpoint. This keeps us all options open.

The more I think about the search endpoint, the more I'm thinking ES as a search backend will even be required as it will make the service itself simpler and more efficient. Each ES instance would be independent and the service would load all the needed that into ES on startup.

As the registry service is completely independent it should be pretty simple to horizontally scale it. The part I don't know how much traffic we will expect as all newer Kibana instances will hit it pretty often.

roncohen:
> The more I think about the search endpoint, the more I'm thinking ES as a search backend will even be required as it will make the service itself simpler and more efficient. Each ES instance would be independent and the service would load all the needed that into ES on startup.

The only problem is see with this is situations where users do not want their Kibana to reach outside their network and use our hosted registry. With ES required, it becomes slightly more complicated to run it, maintain it etc.

ruflin:
[...]

PR #58, opened 2019-08-21
Introduce static file handling

ruflin:
As discussed in #53, some of the files can be generated statically and don't require it's own handler. Pregenerating some of the static content can also simplify the code. To be able to work with both methods in parallel and evaluate this new approach, there is a catchAll handler introduce which is called when no other handler matches. Then it is checked if a file for the request URL exists and if yes, it is served. Like this the deployment model of the services stays exactly the same for now.

To make the above possible, the generated packages are now moved to `public/package` directory. This is done by `mage build`. The side affect of having all packages in there is that also all files inside the package can be directly statically served.

As a first step, the static handler for the `.tar.gz` downloads and `/info` endpoint was removed.


ruflin:
jenkins, test this

ruflin:
@jfsiii What is interesting with this change from the Integrations manager perspective is that now you can fetch any file from inside the package also directly from the web. This should become especially useful for the icons and screenshots. Will open a follow up PR for this as it might changes the current paths.

PR #59, opened 2019-08-21
Remove .zip packages

ruflin:
We discussed earlier that we only need the .tar.gz files for now. To simplify the code, I am removing the zip download handler and related code.


ruflin:
@jfsiii are you ok with this? Do you use .zip still somewhere?

jfsiii:
The Manager isn't using it anywhere at the moment. Or it doesn't have any UI flows where we request a zip.

The only zip-related code is https://github.com/elastic/kibana/blob/c583570f948e4e00f11ecde8de2dcee86d539012/x-pack/legacy/plugins/integrations_manager/server/registry/index.ts#L64 and the `unzipBuffer` function it references.

I think the main (only?) reason for them is installing on Windows. We can remove it now and restore later if we need them without any real impact on the Manager.

ruflin:
Ok, lets remove it for now also from the Kibana code but we should take a note / task to check what happens with the tar.gz code when running Kibana on Windows.

PR #62, opened 2019-08-22
Change path of icons

ruflin:
This changes the current icon path from `/img/{package}-{version}/icon.png` to `/package/{package}-{version}/img/icon.png`. The reasoning is that directly the assets from the package itself are served so no special handlers are needed in the backend. The other breaking change is that in case no icon exists, a 404 is returned instead of the default icon. It's up the integrations manager to decide what to do for a default icon.

With https://github.com/elastic/integrations-registry/issues/45 the API response for icons will be refactored. For now I left the API response the same, meaning it contains the icon path even if no icon exists, but will change that when tackling #45.

### DIFF
--- a/docs/api/package.json
+++ b/docs/api/package.json
@@ -3,7 +3,7 @@
   "title": "Example Integration",
   "version": "1.0.0",
   "description": "This is the example integration",
-  "icon": "/img/example-1.0.0/icon.png",
+  "icon": "/package/example-1.0.0/img/icon.png",
   "requirement": {
     "kibana": {
       "version.min": "7.0.0",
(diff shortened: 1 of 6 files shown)


ruflin:
@jfsiii This is a breaking change.

jfsiii:
Thanks for the ping. The route change doesn't affect the Manager as it just uses whatever path is provided, so as long as the Registry is serving assets from that location all is well.

Will a path be listed if no asset exists? i.e. is it possible the manager would request the icon path provided but the registry would respond with a 404? If there's no icon in the manifest/package/etc, I'd expect the registry to omit the value or otherwise signal it's missing to avoid the roundtrip.

ruflin:
In this PR, it's even there if it is missing so you get a 404. But that is what I will fix in a follow up PR when refactoring the icons part.

PR #63, opened 2019-08-22
Refactor icon structure

### LINKED ISSUE #45: Discuss icon response restrictions

ruflin:
This refactors the structure for icons. The implementation is mostly based on the discussion in #45. With this change, multiple icons can be used and the size can be pre populated. An example output for the CoreDNS package looks as following in the `/search` API endpoint:

```
  {
    "description": "CoreDNS logs and metrics integration.\nThe CoreDNS integrations allows to gather logs and metrics from the CoreDNS DNS server to get better insights.\n",
    "download": "/package/coredns-1.0.1.tar.gz",
    "icons": [
      {
        "src": "/package/coredns-1.0.1/img/icon.png",
        "size": "1800x1800"
      },
      {
        "src": "/package/coredns-1.0.1/img/icon.svg",
        "size": "255x144",
        "type": "image/svg"
      }
    ],
    "name": "coredns",
    "title": "CoreDNS",
    "version": "1.0.1"
  },
```

In the package specific endpoint, it looks like:

```
{
  "name": "coredns",
  "title": "CoreDNS",
  "version": "1.0.1",
  "description": "CoreDNS logs and metrics integration.\nThe CoreDNS integrations allows to gather logs and metrics from the CoreDNS DNS server to get better insights.\n",
  "requirement": {
    "kibana": {
      "version.min": "6.7.0",
      "version.max": "7.6.0"
    }
  },
  "icons": [
    {
      "src": "/package/coredns-1.0.1/img/icon.png",
      "size": "1800x1800"
    },
    {
      "src": "/package/coredns-1.0.1/img/icon.svg",
      "size": "255x144",
      "type": "image/svg"
    }
  ]
}
```

If no icons are available, no `icons` array is exposed.

Closes #45


dev/package-examples/coredns-1.0.1/manifest.yml
```diff
@@ -16,3 +16,10 @@ requirement:
     version.max: 7.6.0
   elasticsearch:
     version.min: 7.0.1
+
+icons:
+- src: "/img/icon.png"
+  size: ["1800x1800"]
+- src: "/img/icon.svg"
+  size: ["255x144"]
```
ruflin: @jfsiii We agreed to make it an array here. But already during the implementation I got it in my test examples twice wrong because I wrote `size: "255x144"`.

I guest most users will not specify a size for the icons, and if they want, WDYT about:

```
- src: "/img/icon.svg"
  size: "1800x1800"
- src: "/img/icon.svg"
  size: "200x200"
```

The same icon can be specified multiple times. This should keep the normal case simpler I hope.

jfsiii: I thought we were using `sizes` (plural) not `size` (singular) https://github.com/elastic/integrations-registry/issues/45#issuecomment-521684522

I wonder if that contributed to your expectation that it was a singular/string value? Perhaps that's also why the Web App Manifest uses a space-delimited string.

I favor the "users over authors over implementors over specifiers over theoretical purity" thinking from the HTML design principles, so I want to be mindful of any pain points we're feeling, but I rather not change just yet.

Mind trying with `sizes` (and optionally the space separated string) for a bit and reporting back?

ruflin: Yes, we had sizes and tried both but did not get happy :-) What I was hoping for initially is that `size: 200x300` and `size: [200x300]` would both work (or `sizes`) to make it for the user as simple as possible.

The above also came from that I wanted to use the same definition as we have for screenshots (which is size). As the two are very close to each other in the manifest, it's comfortable to just copy it over.

Historically we struggle getting icons with the integrations. Having one I assume all of us would be all very happy already, I doubt there will be really many with multiple icons. Because of this I would for now optimise on a single icon. The nice part of using `size` for now is that `sizes` is still available if we need it one day :-)

jfsiii: OK. Have a go with

```
- src: "/img/icon.svg"
  size: "1800x1800"
- src: "/img/icon.svg"
  size: "200x200"
```

I don't think any of these choices have a significant impact on the Manager (user) so let's optimize for authors (you, currently).

ruflin: Ok, will do for now. We will change it when we need if hit a roadblock with it 👍

ruflin:
@jfsiii Did not deploy this yet. Let's sync up today quickly on when we should deploy it. I guess it will break all the icons on your end.

jfsiii:
@ruflin I mentioned in Slack that the Manager is currently ignoring icon paths from the registry. It attempts to find an icon named `logo${packageName}` (like `logoElasticsearch` or `logoPostgres`).

Deploy this whenever is convenient for you.

ruflin:
Just deployed it.

PR #67, opened 2019-08-27
Add longdocs package

ruflin:
This package is intended to test all the varieties  of documentation blocks and length that can exist.


ruflin:
@hbharding @jfsiii I plan to add several "testing packages" to the repository so we can test some of the edge cases and how they look in the UI. Let me know for which edge cases we need some test packages or also add them / modify them directly yourself.

PR #68, opened 2019-08-27
Add Auditd example package

ruflin:
This copies the assets from the Auditd package. This is especially interesting as it contains some sample rules. We must figure out on how we best show this to the users. Also the config is pretty long and has some variables inside for generating different configs.


dev/package-examples/auditd-2.0.4/auditbeat/config/config.yml
```diff
@@ -0,0 +1,45 @@
+{{ if eq .GOOS "linux" -}}
```
ruflin: @ph @andrewkroh @jfsiii @mattapperson This is one more example on where the configs are currently dynamically generated on the edge based on some options or OS. I wonder how you imagine this to work in the future?

ruflin: Agree, Fleet will need to get this meta data form the agent. @ph I leave it up to you on how this is "communicated".

On the integrations side when no Fleet is around, I expect that in the manifest file there will be a list of predefined options for these params and the user needs to select it.

ruflin:
Merging this to have it in as an example to play around with. This doesn't mean we should stop the discussion on this issue.

dev/package-examples/auditd-2.0.4/auditbeat/config/config.yml
```diff
@@ -0,0 +1,45 @@
...
+  {{ if .Reference -}}
+  # The auditd module collects events from the audit framework in the Linux
+  # kernel. You need to specify audit rules for the events that you want to audit.
+  {{ end -}}
+- module: auditd
+  {{ if .Reference -}}
```
ruflin: @andrewkroh Where is this one coming from?

ruflin: Ah, just answered it myself. You combined the reference config and this config into 1 👍

PR #69, opened 2019-08-27
Proposal: How to store the setup / installations docs

ruflin:
WIP

### DIFF
--- a/dev/package-examples/iptables-1.0.4/docs/setup.yml
+++ b/dev/package-examples/iptables-1.0.4/docs/setup.yml
@@ -1,0 +1,62 @@
+# This are the installation steps described as a YAML file. Each package must include its own
+# installation steps.
+
+steps:
+  - filters:
+      cloud: true
+    text: Try in Elastic Cloud, sign up
+
+
+  - filters:
+      cloud: true
+    text: Connect local Kibana instance
+
+  - filters:
+      os.name: macos
+      cloud: false
+      distribution: tar.gz
+    text: >
+
+      Download and install Agent:
+
+      ```
+      curl -L -O https://artifacts.elastic.co/downloads/beats/filebeat/filebeat-{{stack.version}}-darwin-x86_64.tar.gz
+      tar xzvf filebeat-{{stack.version}}-darwin-x86_64.tar.gz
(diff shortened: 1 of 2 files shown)


dev/package-examples/iptables-1.0.4/docs/setup.yml
```diff
@@ -0,0 +1,62 @@
...
+      os.name: [windows, macos]
+      cloud: false
+    text: >
+
+      Edit configuration:
+
+    # This is expected to take the file `input.yml` and replace the variables either
+    # with environment variables set in this file or read `vars` from the manifest
+    # and select the right option based on the variables set here.
+    {{ file /filebeat/input.yml | manifest.yml }}
```
ruflin: @jfsiii Very early brain storming phase. But this is basically a template which refers to an other template with a file as variables. Any recommendations to look at an existing tool for syntax in the JS world?

ruflin: Thanks for the detailed feedback, this is super helpful. I need to look deeper into handlebars. 

I'm starting to think that perhaps we should require the integrations builder already to combine config options and templates into 1 file. It's a bit less abstract but should make parsing it on our end much easier. Will play around with this idea.

dev/package-examples/iptables-1.0.4/docs/setup.yml
```diff
@@ -0,0 +1,62 @@
...
+    text: >
+
+      Download and install
+
+    ```
+    curl -L -O https://artifacts.elastic.co/downloads/beats/filebeat/filebeat-8.0.0-SNAPSHOT-x86_64.rpm
+    sudo rpm -vi filebeat-8.0.0-SNAPSHOT-x86_64.rpm
+    ```
+
+  -  filters:
```
ruflin: As we will strictly separate the configuration from the setup, we should be able to split up this file into 2 parts which should make things simpler.

ruflin:
@nchaulet @jen-huang This is an old PR where I started to think about how we could also have the setup docs in each package to have it also versioned. Now that we only support the agent I suggest we hard code this for now in Kibana. Closing.

PR #71, opened 2019-08-27
File reference in docs

ruflin:
This adds an example on how part of the docs could be generated from existing files. This is especially useful for showing example data or longer configurations.

### DIFF
--- a/dev/package-examples/longdocs-1.0.4/docs/README.md
+++ b/dev/package-examples/longdocs-1.0.4/docs/README.md
@@ -16,7 +16,7 @@ This is a [link](https://github.com/elastic/integrations-registry) inside the do
 The following file shows how an event might look like:
 
 ```
-{{ file /docs/data.json }}
+{{ > ./data.json }}
 ```
 
 ## Template parts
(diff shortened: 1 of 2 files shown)


dev/package-examples/longdocs-1.0.4/docs/README.md
```diff
@@ -11,6 +11,14 @@ This integration is considered to be in _beta_.
 
 This is a [link](https://github.com/elastic/integrations-registry) inside the docs.
 
+## Examples from a file
+
+The following file shows how an event might look like:
+
+```
+{{ file /docs/data.json }}
```
ruflin: @jfsiii Looking for input on this one. It's simpler then what I asked for in https://github.com/elastic/integrations-registry/pull/70 and could be enough to get started.

ruflin: Exactly. Will use this for now. One thing we probably will need in addition is:

```
{{ > {{prefix}}/docs/data.json }}
```
The path on your end will have a prefix and will need replacement I guess. So we have mustache in mustache :-) Suggestions?

ruflin: I went for now with `{{ > ./data.json }}`. The assumption is that all paths are relative to where they are specified.

ruflin:
@skh @jfsiii Merging this example as is for now so we can start to play around with it in the UI. If paths need to be defined differently, let me know.

PR #77, opened 2019-08-29
Add correct headers for .svg files

ruflin:
Closes https://github.com/elastic/integrations-registry/issues/74


jfsiii:
Works!

`` gives

PR #81, opened 2019-08-29
Add /categories endpoint

ruflin:
The `/categories` endpoint is to serve information about how many packages per category exist. An example output looks as following:

```
[
  {
    "id": "logs",
    "title": "Logs",
    "count": 16
  },
  {
    "id": "metrics",
    "title": "Metrics",
    "count": 14
  }
]
```

For now the output does not support any filtering. It just takes all the newest packages and adds them to the categories. The category titles are hardcoded in the registry code for now. Later on we should validate the titles of each package.

The decision was made for the more complex output already know, so we can add additional fields in the future without breaking the API.

In the future I imagine this endpoint supports the same filtering capabilities as the search endpoint.

Closes https://github.com/elastic/integrations-registry/issues/75

### DIFF
--- a/categories.go
+++ b/categories.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/blang/semver"
+
+	"github.com/elastic/integrations-registry/p"
+)
+
+var categoryTitles = map[string]string{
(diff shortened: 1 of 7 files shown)


ruflin:
@jfsiii This endpoint currently does not contain any information about the `All` packages. Would you expect this to be returned too?

PR #82, opened 2019-08-29
Allow to filter by category

ruflin:
Now one of the available category id's can be used to filter the packages by a category. For now this is `logs` and `metrics`, more will be added in the future.

Further changes:

* A new test package was added in this PR to better test the category endpoint.
* Ignore .tar.gz packages in the testdata directory. These are not commited as they change every time they are generated.
* Package output is now sorted


ruflin:
@jfsiii The should be helpful to show the pages which you filter by category.

PR #85, opened 2019-09-03
Translate manifest and input to handlebars template

ruflin:
This PR combines the manifest and input template into 1 document. This allows to have all information in one place and potentially for easier processing.

### DIFF
--- a/dev/package-examples/iptables-1.0.4/filebeat/config/input.yml
+++ b/dev/package-examples/iptables-1.0.4/filebeat/config/input.yml
@@ -1,25 +1,32 @@
-{{ if eq .input "syslog" }}
-
+{{setEmptyVar input "syslog"}}
+{{ ifeq input "syslog" }}
 type: syslog
 protocol.udp:
-  host: "{{.syslog_host}}:{{.syslog_port}}"
+  host: "{{defVal syslog.host localhost}}:{{defVal syslog.post 9001}}"
 
-  {{ else if eq .input "file" }}
 
+{{ elseifeeq input "file" }}
 type: log
 paths:
-  {{ range $i, $path := .paths }}
-- {{$path}}
-  {{ end }}
+-
+  {{!-- If the paths variable is set, it will be used, otherwise falls back to the default config. --}}
+  {{ if paths }}
+  {{#each path}}
+    {{this}}
+  {{/each }}
+  {{ else }}
+    /var/log/iptables.log
+  {{/if}
 exclude_files: [".gz$"]
+{{ /if }}
 
-  {{ end }}
-
-tags: {{.tags}}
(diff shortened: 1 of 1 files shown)


dev/package-examples/iptables-1.0.4/filebeat/config/input.yml
```diff
@@ -1,25 +1,32 @@
-{{ if eq .input "syslog" }}
-
+{{setEmptyVar input "syslog"}}
```
ruflin: This helper would set a variable to the given value if the variable is not set yet.

dev/package-examples/iptables-1.0.4/filebeat/config/input.yml
```diff
@@ -1,25 +1,32 @@
...
-  {{ else if eq .input "file" }}
 
+{{ elseifeeq input "file" }}
 type: log
 paths:
-  {{ range $i, $path := .paths }}
-- {{$path}}
-  {{ end }}
+-
+  {{!-- If the paths variable is set, it will be used, otherwise falls back to the default config. --}}
```
ruflin: I'm realising now, that if we have the set for the default variable, we could skip this and instead set paths if not set yet.

ruflin:
@jsoriano Pinging you on this one as this changes how manifest.yml and input.yml in the modules today work and merges it into 1 file. Would be good to get your feedback on it too.

dev/package-examples/iptables-1.0.4/filebeat/config/input.yml
jsoriano: I'd prefer to have some more explicit way to define the defaults as we had with `manifest.yml`. Before, we had an only way to set variables in the fileset definition, and here I see three (`setEmptyVar`, `defVal` and this custom check).

Also we would still need some logic like the one offered previously by the `manifest.yml` file, where you could set defaults also per operating system, e.g. in the apache module:
```
  - name: paths
    default:
      - /var/log/apache2/access.log*
      - /var/log/apache2/other_vhosts_access.log*
      - /var/log/httpd/access_log*
    os.darwin:
      - /usr/local/var/log/apache2/access_log*
    os.windows:
      - "C:/tools/Apache/httpd-2.*/Apache24/logs/access.log*"
      - "C:/Program Files/Apache Software Foundation/Apache2.*/logs/access.log*"
```

ruflin:
I'm peddling back on this approach here and think we should keep using the var definitions we have at the moment. Long term we need to allow the user to put in some of the variables in the UI. `paths` is one of them. If we have the var definitions in a separate file, we can also directly read from this file which variables the user can configure and how.

What I think should still be the case is that we use handlebars for the templates but variables are defined externally.

PR #88, opened 2019-09-03
Introduce package filtering in /search endpoint

ruflin:
To show which versions of a package exist, the `/search` endpoint must support filtering by a package and then return all versions of this package.

This introduces the parameter `package` with which queries like `/search?package=mysql` can be created.

These results can be used for showing a drop down when upgrading a version to show which other versions exist.

### DIFF
--- a/search.go
+++ b/search.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strings"
 
 	"github.com/blang/semver"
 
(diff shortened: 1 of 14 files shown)


ruflin:
@jfsiii Would be good to get your feedback on this one.

General: Currently still in draft as the code is not as nice yet as I would like.

jfsiii:
@ruflin What you have here now (`/search?package=example` returns response shown in `docs/api/search-package-example.json`) LGTM for creating a dropdown in the UI.

search.go
```diff
@@ -134,19 +159,23 @@ func validKibanaVersion(version *semver.Version, m *util.Manifest) (bool, error)
...
+			keys = append(keys, key+"-"+v)
+		}
 	}
 	sort.Strings(keys)
 
 	var output []map[string]interface{}
 
 	for _, k := range keys {
-		m := packagesList[k]
+		// TODO: this could break if we have integrations with `-` in the name
```
jfsiii: I created [this ticket](https://github.com/elastic/kibana/issues/41253) to change to `name@version` in the Manager for this reason.

ruflin: Good idea, will do the same here.

ruflin:
@jfsiii Yes, exactly. It's the same format as all other outputs form the `/search` endpoint.

I'm going to merge this as is and follow up with a bigger refactoring of the search endpoint.

ruflin:
@jfsiii I added a package called `multiversion` which has 3 versions available. Like this we have in our registry a package with multiple versions for testing.

PR #89, opened 2019-09-03
Adjust description that fields.yml should be used for generation

ruflin:
To keep all the flexibility with the fields.yml also after creating a package, for now I suggest that the integrations managers converts it to index templates. This has the advantage that we do not need to merge all fields.yml into one file in the package and can delay the decision to the installation if one or multiple templates (logs vs metrics) are created.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -103,8 +103,9 @@ manager must be able to overwrite the index pattern.
 
 On the Beats side today the Index Template is generated out of the `fields.yml` files. This allows to give more flexibility
 to generate the correct template for different Elasticsearch version. As we can release integrations packages for different
-version of Elasticsearch independently this is probably not needed anymore. I expect one or multiple fields.yml to be
-in each Integrations Package but leave it to the package creator to create the index template (TBD).
+version of Elasticsearch independently this is probably not needed anymore. I expect fields.yml to stick around as it's a nice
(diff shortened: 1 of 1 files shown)


ruflin:
@jfsiii This would mean that we need to reimplement the logic to understand `fields.yml` in the integrations manager which we have today in Beats. Nothing fancy, just traversing trees.

jfsiii:
@ruflin OK. Can you link me code where the processing is done? Ideally we'd extract it to another lib/service and both consume/use it, but we can duplicate for initially and figure out how best to keep in sync later.

ruflin:
The generation of the Elasticsearch index template can be found here: https://github.com/elastic/beats/blob/master/libbeat/template/template.go Generation of the Kibana index pattern here: https://github.com/elastic/beats/blob/master/libbeat/kibana/index_pattern_generator.go The implementation of the reading fields.yml itself is here: https://github.com/elastic/beats/blob/master/libbeat/mapping/field.go 

It's all in Golang at the moment so sharing is tricky. It's something we rarely update and I expect to even become obsolete in the future. One reason it would be nice having it as a lib is then we could also use it for testing our integrations without directly requiring Kibana but that is something we could figure out later.

@jfsiii It's probably worth that I open an issue in Kibana related to this?

PR #94, opened 2019-09-05
Add infrastructure-ui-source to the Assets

ruflin:
This is an asset which we don't have yet in modules today but could become useful in the future. I'm adding it here for completeness and for having a discussion around it. It is not expected to be supported in our first version.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -213,6 +213,17 @@ or support inheritance / composition like the index templates in Elasticsearch. 
 problem as a user would have to constantly switch between them. The integrations manager could work around this by
 append / updating index pattern. But this will lead to the problem on how to remove these fields again.
 
+### Infrastructure UI Source
+
+* Asset Path: `kibana/infrastructure-ui-source/*.json`
+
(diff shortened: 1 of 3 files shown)


ruflin:
@skh @jfsiii After our discussion yesterday, I had a chat with @weltenwort about how we could integrate. To make the links possible, we need to load a source. But this source could be part of a package. I don't expect that we support this short term as we don't have it yet in any packages / modules as of today, but worth having in mind.

@andresrc @exekias @jsoriano This could be very interesting in the future to have it as part of the integrations. It allows us to neatly integrate with the Logs and Metrics UI, for example which columns should be shown in the Logs UI for an integration or syntax highlighting.

ruflin:
Thanks for the feedback @weltenwort Any chance you could open a Github issue for the Infra UI on this in Kibana and link it here? Would do it directly myself but I think if you phrase it the team will much better understand what is meant by it.

weltenwort:
Sure, I can create it as a proposal.

PR #95, opened 2019-09-10
Add package type to manifest file

ruflin:
Currently the type is not used and the default type is 'integration'. But this gives us the option in the future to also introduce packages which are not integrations, for example for solutions.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -27,6 +27,9 @@ version: 0.0.2
 categories: ["logs", "metrics"]
 # Options are experimental, beta, ga
 release: beta
+# The package type. The options for now are [integration, solution], more type might be added in the future.
+# The default type is integration and will be set if empty.
+type: integration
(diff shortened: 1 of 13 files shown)


ruflin:
@jfsiii @skh This currently has no effect on the UI I assume but probably good to know it is there.

jfsiii:
I see the `docs/api/*` files were updated. Out of curiosity, is that automatic when you build/merge/deploy or do you need to remember to "rebuild"?

This change isn't breaking but it is what I was referring to in https://github.com/elastic/kibana/issues/38598 I really want to work out some coordination which removes the human component -- the need to ping one another.

ruflin:
To update these files I need to run `go test . -generate` as otherwise my tests fail. +1 on having some automation here. A human thing that will also reduce pings will be that we start using the Changelog and do proper versioning. My plan is to start this today.

PR #96, opened 2019-09-10
Update README about API endpoints and query params

ruflin:
Also adding some details about basic architecture / setup.


ruflin:
@skh Would be great to get your feedback on this update. Happy to add more details.

PR #97, opened 2019-09-10
Add Validation of package

ruflin:
To make sure the registry only serves valid packages, packages should be validated during build time. For this a `Package.Validate() error` method was added. For not it is simple and only checks for title and description but more checks should be added in the future.

The packages which were not valid were updated.


ruflin:
@jsoriano I expect we will also be able to use this as part of a validation step in https://github.com/elastic/integrations/pull/3 Might also become part of this "magical" binary ;-)

PR #100, opened 2019-09-10
Validate packages to only use predefined categories

ruflin:
This required to move the categories to util package.


ruflin:
Conflicts resolved, should be ready for a round of reviews.

PR #102, opened 2019-09-11
Prepare release 0.1.0

ruflin:
This is the first release that should be tagged. So far the changelog was not used but that will change now. All development will still happen on top of master.

All files were updated for the this release. Breaking changes might still happen until we hit 1.0 but they will be documented.


ruflin:
I tagged a release so we get in the habit of doing it: https://github.com/elastic/integrations-registry/releases/tag/v0.1.0

PR #104, opened 2019-09-11
Add .dockerignore file for slimmer image

ruflin:
Adding the same files to .dockerignore as we had in the .gitignore files as these are not needed to build the container. This should make sure if a container is built locally but some of the .gitignore files are modified, it still builds to container. With this, less unncessary files should end up in the container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
+.DS_Store
+integrations-registry
+
 .idea
 build
 public/*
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -0,0 +1,3 @@
+.idea
```
jfsiii: Perhaps `.DS_Store` as well?

ruflin: Good point, also added now `integrations-registry`. Seems like I did a bad copy of the .gitignore file.

PR #108, opened 2019-09-19
Add pipeline reference example to coredns

ruflin:
Pipelines are input specific. Currently there is no simple way for the integrations manager to know which pipeline should be reference in an input. This introduces the convention that the pipeline name must be a variable which matches one of the pipeline file names in the package.

On creating the pipeline instance, the integrations manager will change and prefix the name. The above convention makes it possible to convert it in both places. As an example: The package has an ingest pipeline `foo.yml` or `foo.json`, in the input it will reference `foo`. Then the integration manager builds the datasource and calls it `bar-foo`. It will the replace the input variable with `bar-foo`.

This also has consequences on https://github.com/elastic/integrations/pull/3. It means the developer building integrations must ensure that no two pipelines have the same name in one package.

To make sure the name in an input matches a pipeline, later on also a validation step should be added.

### DIFF
--- a/dev/package-examples/coredns-1.0.1/agent/config/input.yml
+++ b/dev/package-examples/coredns-1.0.1/agent/config/input.yml
@@ -1,0 +1,6 @@
+- type: logs
+  paths:
+    - /var/log/coredns.log
+  # This must be identical to the name of the pipeline file (without .yml)
+  pipeline: {{pipeline-entry}}
+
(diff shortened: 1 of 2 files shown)


ruflin:
@skh @jfsiii Would be good if you could have a look at the above explanations to see if that makes sense from a manager perspective.

ruflin:
@jsoriano Pinging you on this one as it's relevant for https://github.com/elastic/integrations/pull/3

ruflin:
Merging for now to have the change in as an example. Overall rethinking of inputs will happen in https://github.com/elastic/integrations-registry/pull/110

PR #110, opened 2019-09-19
Adding concept of input to package

ruflin:
In the future metricbeat / filebeat and the agent will only support inputs. With this inputs become a first class citizen in our stack. An input is basically an agent configuration + an ingest pipeline. At the moment the package content is focused on having a config for the Beat or agent and all pipelines in one place. This complicates 2 things:

* Knowing which ingest pipeline belongs to a specific input
* Building integrations with multiple inputs: https://github.com/elastic/integrations/pulls

Having the concept could simplify things as the package builder must not try to prevent naming configs of ingest pipeline by introducing extra long names. Also it should simplify testing as often testing is focused on inputs. With this all assets related to an input are together.

As part of this PR there is an example on how such an input structure could look like. This should not replace the old place of ingest pipelines. If a user wants to build a package with just a ingest pipeline but not an input, this should also be possible in the future.

The changed structure is described in the ASSET.md file.


dev/package-examples/nginx-1.2.0/input/stubstatus/fields.yml
```diff
@@ -0,0 +1,46 @@
+- name: stubstatus
```
ruflin: This should be the full path here in the future.

dev/package-examples/nginx-1.2.0/input/access/agent/input/input.yml
```diff
@@ -0,0 +1,14 @@
+# This is not an array on purpose to make sure only 1 single input is specified in this file.
```
ruflin: In a meeting, @skh brought up a good point here that we have a double meaning of input here (shows up twice in the path). We didn't reach a conclusion on what this should be called, but `dataset` was mentioned as an option.

@exekias @jsoriano @ph Some thoughts on this?

ruflin: I have been thinking about this a bit more. A dataset is basically a template for an input with all its assets. All inputs with the data set `access.log` look exactly the same in the end. An input can exist multiple times, still it is the same dataset. So I think this fits well here. Will rename.

ruflin: Just pushed a commit with renaming it. One thing I realised is that the agent now does not have an input anymore, but only streams. So the above was renamed to `agent/stream/config.yml`. @ph Does this sound correct?

ASSETS.md
```diff
@@ -259,6 +259,88 @@ but it can be triggered through URL parameters.
...
+**manifest.yml**
+
+Each input must contain a manifest.yml. It contains all information about the input and how to configure it.
+
+```
+# Needs to describe the type of this input. Currently either metric or log
+type: metric
+
+# Each input can be in its own release status
+release: beta
```
ruflin: @hbharding Some inputs can also be in beta. We probably need some design to indicate this on the create data source page where inputs can be enabled / disabled.

ruflin:
To have an example of a dataset in the repository for further discussion, I will merge this PR. There are still open questions around naming (dataset vs input) but the basic structure should stay the same. Having it in the repository will allow the EPM team to start implementing the structure and we get feedback if it works as expected. Also all future changes will be documented.

PR #112, opened 2019-09-24
Move package generation to its own binary

ruflin:
Generating a package to be served by the package registry can be used in different places. This extracts the generation of it to its own binary. The binary can be found under `dev/generator` and be built with `go build .`.

The binary takes 2 params:

* `sourceDir`: Path to the source directory of the packages
* `publicDir`: Public directory where the packages should be copied to and built

This might look like `go run . -sourceDir=../packages -publicDir=./public`

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -1,0 +1,139 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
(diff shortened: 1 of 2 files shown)


ruflin:
@skh I hope this will be useful for you.

PR #115, opened 2019-09-26
Add Docker healthcheck

ruflin:
The healthceck is useful in Docker environments. All the preloading of packages happens before the http Server is started. Because of this the `/` endpoint is a good healthcheck indicator. As the registry is never started in error cases, no special healthcheck endpoint is needed.


ruflin:
@tylerjl Here we go with the healthcheck for the service. Let me know if that is what you expected.

ruflin:
@tylerjl I will update the PR with a note in the docs.

ruflin:
Docs updated.

PR #117, opened 2019-09-26
Clean up changelog

ruflin:
Some changelog entries went missing in recent PRs. They are added now.

The categories of the changelog were also cleaned up and aligned with what is used in Beats.


ruflin:
No template yet, but we can introduce one.

PR #118, opened 2019-09-26
Remove packages.path config and replace it with public_dir

ruflin:
The only thing needed to start the server is having it pointed to the public directory to serve files from there. What the internal name is for the packages directory is an implementation detail.

Some refactoring of the code was needed for this change. Now path variables are passed to the handler instead of using a global variable. This also makes it simpler to test individually instead of modifying global variables.


ruflin:
Rebase was needed as this PR confliced with the cacheTime PR.

PR #120, opened 2019-10-21
Add svg icons from EUI to packages

ruflin:
Lots of pacakges were missing icons. These were added as svg and references in the manifest file. System and traefik are still missing the icons which should be fine.

Closing https://github.com/elastic/integrations-registry/issues/119


ruflin:
@jfsiii All "magic" code to show icons from EUI directly should be removed from EPM.

PR #123, opened 2019-10-21
Remove assets from generated packages

ruflin:
The registry so far contained generated packages. These packages contained all the same assets and were mainly designed to test the overview page of the package manager. But this now caused problems as many dashboards for example had the same ids. Because of this, all the assets are removed from these example packages and only the package with the icons are now left which is enough to show a nice home screen.


jfsiii:
The tradeoff is that this removes the data which drives:

- the assets sidebar on the right column of the details page
Screen Shot 2019-10-21 at 9 08 46 AM

- the assets "panel" in the details page

Screen Shot 2019-10-21 at 9 08 58 AM

ruflin:
Correct, but we still have packages which contain all the assets (and real assets) here: https://github.com/elastic/integrations-registry/tree/master/dev/package-examples Happy to add more here.

jfsiii:
> we still have packages which contain all the assets (and real assets) here: https://github.com/elastic/integrations-registry/tree/master/dev/package-examples 

Ah, ok. That's `auditd`, `coredns`, and `iptables` for now. I just took a look and they have a greater breadth of asset types than the set the others were sharing.

> Happy to add more here

I believe we are missing examples with ILM policies and rollup jobs, but I will confirm and get back to you

ruflin:
@jfsiii Correct. But rollups will not be part of MVP AFAIK.

PR #126, opened 2019-10-22
Add base package

ruflin:
The base package contains all the assets, which are used by all packages. This package is expected to be installed before any other package is installed. EPM should install it automatically and no user interaction should be needed.

For now this only contains the two default ILM policies for logs and metrics. More assets could be added later.

A flag `internal: true` was added to the packages so these packages should not be shown in the overview. By default the internal packages are not exposed in the `/search` endpoint. To also get internal packages, the query param `?internal=true` has to be set. In the `/categories` endpoint, internal packages are not counted.

Further changes:

* Test package for testing `internal: true` added
* README updated with new endpoint.
* Add test for `internal=true` query param

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ The `/search` API endpoint has few additional query parameters. More might be ad
 * category: Filters the package by the given category. Available categories can be seend when going to `/categories` endpoint.
 * package: Filters by a specific package name, for example `mysql`. In contrast to the other endpoints, it will return
   by default all versions of this package.
+* internal: This can be set to true, to also list internal packages. This is set to `false` by default.
   
 The different query parameters above can be combined, so `?package=mysql&kibana=7.3.0` will return all mysql package versions
 which are compatible with `7.3.0`.
--- a/categories.go
+++ b/categories.go
@@ -33,6 +33,11 @@ func categoriesHandler(packagesBasePath, cacheTime string) func(w http.ResponseW
 		// Get unique list of newest packages
 		for _, p := range packages {
 
+			// Skip internal packages
+			if p.Internal {
+				continue
+			}
+
 			// Check if the version exists and if it should be added or not.
 			if pp, ok := packageList[p.Name]; ok {
 				// If the package in the list is newer, do nothing. Otherwise delete and later add the new one.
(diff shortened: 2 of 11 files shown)


ruflin:
@ph @mattapperson @skh @jfsiii @exekias I'm collecting all the default assets in this package for now. It will probably make sense to ship this package or just the assets directly with Kibana.

ph:
sound good

dev/package-examples/default-1.0.0/manifest.yml
```diff
@@ -0,0 +1,23 @@
...
+title: Default
+description: >
+  The default package contains assets which are needed for the basic setup of the stack.
+
+  It contains the default ILM policies.
+version: 1.0.0
+categories: []
+release: ga
+# The user should not see this package and not be able to install it
+visible: false
```
ruflin: @jfsiii @skh Any thoughts on a param naming for packages which are internal and should not be shown in the UI?

jfsiii: `private` and `internal` are also options

jfsiii:
🚲 🏠: I take "default" to mean "use this value unless another value is given". e.g. mutually exclusive `A` OR `B`. IIUC, we'll install will be "union of the `default` and `apache` packages".

Will _everything_ in this package _always_ be installed? If so, would "core" or "base" or something similar also work? If not, perhaps default _s_ (plural)? i.e. it's not the default package (the one we install absent other options) it's the package which holds the _defaults_ for assets?

Do not need to decide this now, and happy to talk in person vs typing it out. Just trying to get clarification and writing down what I thought when I first read it.

ruflin:
+1 on `core` or `base`, makes more sense in this case. These asses build the foundation (base) for everything else.

dev/package-examples/default-1.0.0/manifest.yml
ruflin: went for `internal: true`

ruflin:
Ok, this should be ready for an other round of review. Query params and tests added.

ruflin:
@jfsiii Let me know if the query params work for you as expect. Basically the output is still the same as now and internal packages have to be requested with `internal:true` if needed.

PR #129, opened 2019-10-22
Update import syntax to match Handlebars

jfsiii:
From https://github.com/wycats/handlebars.js#compatibility

>Handlebars does not allow space between the opening `{{` and a command character such as `#`, `/` or `>`. The command character must immediately follow the braces, so for example `{{> partial }}` is allowed but `{{ > partial }}` is not.


ruflin:
Good to know.

ruflin:
@jfsiii Directly merged it, hope that is ok.

PR #130, opened 2019-10-22
Fix test data generation

ruflin:
When updating the generation process of assets, the test packages were not kept up to date. Some refactoring was needed to make it available again. Now it is always run as part of `mage check` so it is run on all tests. Like this, the test packages are also always up-to-date.

Some of the testdata had to be updated.


ruflin:
Merging as soon as CI is green as I need this for https://github.com/elastic/integrations-registry/pull/128 No functional changes in this PR.

PR #131, opened 2019-10-23
Convert packages to new dataset structure

ruflin:
This converts the existing example packages to the new structure with a directory for each dataset/input. Not all assets are converted to 100% and will be converted to the proper format as soon as the work on EPM progressed and further feature requests come in.


dev/package-examples/auditd-2.0.4/dataset/auditd/agent/stream/stream.yml
```diff
@@ -0,0 +1,49 @@
+type: auditd/auditd
```
ruflin: @ph What is the type for auditbeat modules?

ph: @ruflin https://docs.google.com/spreadsheets/d/1jv_zQIlhv88BYXcpr3i2mo2MVB0OpPUPMYqH8JeSUjU/edit#gid=1351342325

In the above case audit/auditd

ruflin: thanks, adjusted.

PR #132, opened 2019-10-23
Prepare renaming of integration-registry to package-registry

ruflin:
We decided some time ago to rename the integration-registry to package-registry as it contains more then just integrations. This is a first step to rename the service by adjusting the docs. A second PR will be needed when renaming the repository to adjust import golang paths.

Part of https://github.com/elastic/integrations-registry/issues/113

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -1,6 +1,6 @@
-# Integration package assets
+# Package assets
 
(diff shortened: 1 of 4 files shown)


ruflin:
jenkins, test this

PR #134, opened 2019-10-23
Adjust API output for Kibana compatibility version

ruflin:
Previously the json output for the compatible Kibana version looked as following:

```
"kibana": {
  "version.min": "7.0.0"
}
```

As this causes troubles on the JS side, it is changed to:

```
"kibana": {
  "version": {
    "min": "7.0.0"
  }
}
```

This should make it easier to parse.

To make it for package creators still possible to use the short notation in the manifest file, go-ucfg is used to read the manifest file so we get this functioanlity for free.

An additional change which happened in this PR is that if a max or min version are empty, it is not part of the API output anymore.

Closes https://github.com/elastic/integrations-registry/issues/44


ruflin:
@jfsiii @skh This might be a breaking change for EPM.

jfsiii:
@ruflin It is a breaking change but, practically, is very easy to update in the Manager.

This is the type of change/breakage I was thinking of with https://github.com/elastic/kibana/issues/38598. I'd love some way to identify (and fail, alert, etc) when the response from the registry no longer validates.

ruflin:
Opened https://github.com/elastic/kibana/pull/49172 on the Kibana side.

PR #137, opened 2019-10-28
[WIP] Add a script to automatically import integrations from the Beats repo

exekias:
This PR adds a script to import integrations from Metricbeat and Filebeat. Running `import_beats` will walk beats repo, create the proper manifests, and copy over assets for each module.

image

Having the repo checked out in the same root folder as this repo, run to update `dev/package-beats` folder.
```
go run dev/import_beats/ -beatsDir ../beats -publicDir dev/package-beats
```

This is still a work in progress effort with many things TODO:

- [ ] Generate + copy `fields.yml`
- [ ] Come up with a plan for docs
- [ ] Dashboard screenshots
- [ ] Icons
- [ ] Input configs
- [ ] ...and more

Reviews will probably want to focus on `import_beats/main.go` file, as the rest are generated.

### DIFF
--- a/dev/import_beats/main.go
+++ b/dev/import_beats/main.go
@@ -1,0 +1,217 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/elastic/integrations-registry/util"
+)
+
+// BEAT_CATEGORIES maps beats to integration categories
+var BEAT_CATEGORIES = map[string]string{
+	"metricbeat":        "metrics",
+	"x-pack/metricbeat": "metrics",
+	"filebeat":          "logs",
+	"x-pack/filebeat":   "logs",
+}
+
+var IGNORE = map[string]interface{}{
+	"apache2": nil, // was renamed to apache
+}
+
+func main() {
+	// Beats repo directory
+	var beatsDir string
+	// Target public directory where the generated packages should end up in
+	var publicDir string
+
+	flag.StringVar(&beatsDir, "beatsDir", "", "Path to the beats repository")
+	flag.StringVar(&publicDir, "publicDir", "", "Path to the public directory ")
+	flag.Parse()
+
+	if beatsDir == "" || publicDir == "" {
+		log.Fatal("beatsDir and publicDir must be set")
+	}
+
+	integrations := NewIntegrations()
+	for beat, category := range BEAT_CATEGORIES {
+		dirs, err := ioutil.ReadDir(filepath.Join(beatsDir, beat, "module"))
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		for _, module := range dirs {
+			if !module.IsDir() {
+				continue
+			}
+
+			if _, ok := IGNORE[module.Name()]; ok {
+				continue
+			}
+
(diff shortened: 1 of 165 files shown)


dev/import_beats/main.go
```diff
@@ -0,0 +1,219 @@
...
+		// dashboards
+		srcDashboardPath := filepath.Join(integration.Path, "_meta/kibana/7/dashboard")
+		fmt.Println(srcDashboardPath)
+		srcDashboards, err := ioutil.ReadDir(srcDashboardPath)
+		if err != nil && !os.IsNotExist(err) {
+			log.Fatal(err)
+		}
+
+		if len(srcDashboards) > 0 {
+			dstDashboardPath := filepath.Join(path, "kibana", "dashboard")
```
ruflin: We should discuss what is best for you as the desitination dashboard / visualization etc. format. Also should it be encoded or not?

magefile.go
```diff
@@ -31,7 +31,7 @@ var (
 
 	publicDir    = "./public"
 	buildDir     = "./build"
-	packagePaths = []string{"./dev/package-generated/", "./dev/package-examples/"}
+	packagePaths = []string{"./dev/package-generated/", "./dev/package-examples/", "./dev/package-beats/"}
```
ruflin: Perfect for now.

dev/import_beats/main.go
exekias: That's a good question, so far it has been good for development and keeping things in a repo. I would be +1 on keeping them as a pretty json

dev/import_beats/main.go
```diff
@@ -0,0 +1,219 @@
...
+			},
+			Path: path,
+		}
+		i.list[meta.Key] = integration
+	}
+
+	integration.Categories = append(integration.Categories, category)
+
+	// TODO come up with a general description
+	integration.Description = meta.Description
```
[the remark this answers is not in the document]
exekias: yes, I think we will need to do this for many things, will try to figure out what's missing

dev/import_beats/main.go
ruflin: I definitively think we should keep it as pretty json for the development of integrations. What about inside the packages?

exekias: I guess as part of the publish process we can do any transformation that is needed. This probably depends on what's best for Kibana?

ruflin: I would argue Kibana wants to do as little work as possible so getting already the "right" format would be beneficial form them I would argue.

One thing to take into account here is Kibana will potentially move to the decoded format in the future ...

ruflin: My thinking at the moment:

* Decoded in the integrations repo where you build it
* Decoded in the unzippped packages we have here at the moment
* Encoded inside to public dir.

ruflin:
@exekias To move this PR forward, following suggestion. We take a subset of the script you have above that does the following:

* Create the correct directories from the modules for the packages
* Creates a manifest (not necessarly complete)
* Ignore this directory from now from the build process but check in the files

The above should allow us to make many small iterations on the script, like adding datasets, completing manifests etc. and always see the changes it makes. Also not adding it directly to the package build process will mean we can play around with it, without affecting the current packages.

WDYT?

exekias:
Yes! I think we should retake this effort (maybe even from scratch :innocent:). I agree with a plan like that

exekias:
I have opened https://github.com/elastic/package-registry/issues/221 to formalize this effort, feel free to update it. Closing this PR for now, as it still requires some work to be ready

PR #138, opened 2019-10-29
Rename integrations-registry to package-registry

ruflin:
The repository was renamed to package-registry and with it all the paths inside as needed.


ruflin:
Merged this even though CI was not run. CI needs updating to match the new repo. Same is true for the docker image build. Following up with both.

PR #139, opened 2019-10-29
Add version ranges instead of min / max version

ruflin:
Currently it is only possible to specify a min and max version as compatibility. This has its limitations as the especially the max version is not always known in advance but all versions of the minor should be compatible.

Implementing this change I realised we should only support ranges or min / max but mixing them gets confusing. Having just one string to define the full range also becomes powerful. For example 6.8 was released after 7.0 and it could be that a package is compatible with 6.8 and 7.2. This is not possible to represent at the moment.

My proposal in this PR is to remove min / max and instead use a `version` or `versions` string.

```
requirement:
  kibana:
    versions: ">=7.1.0 <8.0.0"
```

The supported formats by the library I use can be found here: https://github.com/blang/semver#ranges

### DIFF
--- a/docs/api/package.json
+++ b/docs/api/package.json
@@ -11,9 +11,7 @@
   ],
   "requirement": {
     "kibana": {
-      "version": {
-        "min": "7.0.0"
-      }
+      "versions": "\u003e=7.0.0"
     }
   },
   "screenshots": [
--- a/testdata/package/foo-1.0.0/index.json
+++ b/testdata/package/foo-1.0.0/index.json
@@ -9,9 +9,7 @@
   ],
   "requirement": {
     "kibana": {
-      "version": {
-        "min": "7.0.0"
-      }
+      "versions": "\u003e=7.0.0"
     }
   },
   "assets": [
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -35,14 +35,11 @@ os.platform: [darwin, freebsd, linux, macos, openbsd, windows]
 
 requirement:
   elasticsearch:
-    version.min: 7.0
-    version.max: 7
+    versions: >7.0
   kibana:
-    version.min: 7.0
-  metricbeat:
-    version.min: 7.1
-  filebeat:
-    version.min: 7.2
+    versions: >7.0
+  agent:
+    versions: >7.1
 
 # The order of the items listed here is the order they show up in the package overview.
 screenshots:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* Change `requirements.kibana.version.min/max` to `requirements.kibana.versions: {semver-range}`
+
 ### Bugfixes
 
 ### Added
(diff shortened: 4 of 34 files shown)


ruflin:
@exekias @jfsiii Would be great to get your feedback on this one and what your preference for the config naming is.

@hbharding This would also have an affect on the current design as we don't have min / max anymore.

ruflin:
jenkins, test this

ruflin:
jenkins, test this

exekias:
This sounds good to me, as long as Kibana is able to parse it :)

jfsiii:
I like using a `string` for a `semver` expression.

It is possible that we could still represent those ranges to the user as a min & max since those expressions aren't obvious. https://semver.npmjs.com is one example to show _all_ matching versions, but we could only show min & max.

Seems like the registry or the UI could take all known versions, then filter them for those which match the expression, then show the low & high value.

We don't have to do this, but I want to point out that it's possible to both keep the original UI and change the way we specify values.

ruflin:
@exekias Kibana does not need to parse it. The way it works is that Kibana sends with each request its version to the registry and the registry only returns the packages which are compatible. 

@jfsiii For the min / max: I think there are cases where there are multiple min versions, like `6.8` and `7.2`. Do you think it would be ok to show the actual expressions to users in our first release? We can still go "all in" and show all the compatible versions like NPM does.

ruflin:
@jfsiii @exekias WDYT think about the name of the config: `*.kibana.versions: ...`?

exekias:
> @exekias Kibana does not need to parse it. The way it works is that Kibana sends with each request its version to the registry and the registry only returns the packages which are compatible.

Awesome! didn't know that. What about the rest of apps in the stack, I understand we will use this same format for agent version, for instance?

The config name SGTM

ruflin:
The same will apply for agent, elasticsearch etc. Kibana sends as part of the query these versions and only gets the compatible package. But so far only Kibana version check is implemented.

exekias:
Interesting, how will kibana know my agent version when I'm installing a new integration (without fleet)?

ruflin:
@exekias Good question, it doesn't. We will at least show the user which versions are compatible and he can decide, if he can use this version of the package.

testdata/package/foo-1.0.0/index.json
```diff
@@ -9,9 +9,7 @@
   ],
   "requirement": {
     "kibana": {
-      "version": {
-        "min": "7.0.0"
-      }
+      "versions": "\u003e=7.0.0"
```
ruflin: Hm, that should not look like that :-(

ruflin:
@jfsiii The example for the versions is Elasticsearch 6.8 and 7.2. The version 6.8 was released after 7.0 and contains some features which are only in 7.1 or 7.2. It is an edge case but thought worth bringing up.

If @hbharding agrees, I would suggest we go with implementation speed for now and improve it over time.

docs/api/package.json
```diff
@@ -11,9 +11,7 @@
   ],
   "requirement": {
     "kibana": {
-      "version": {
-        "min": "7.0.0"
-      }
+      "versions": "\u003e=7.0.0"
```
ruflin: @jfsiii This is automatically encoded during marshalling. I don't think this should cause any issue in the UI.

jfsiii: I think you're right.

```
> node -p 'console.log("\u003e=7.0.0")'
>=7.0.0
```

ruflin: @jfsiii Any with this, I just learned a nice new trick on how to quickly execute some js code.

PR #142, opened 2019-10-30
Get rid of the last include commands

ruflin:
Some `::include` commands were left over from the asciidoc conversion. This PR removes these.


ruflin:
@neptunian After merging and deploying this, no includes should be left over.

PR #143, opened 2019-10-30
Remove file include syntax from Markdown

ruflin:
We decided to go with pure Markdown for now for simplicity reason and not allow users to include other files. Because of this I am removing these examples.


ruflin:
@exekias This change means, it is up to the builder of the package to make sure all docs are included in one README file together.

@hbharding You brought up yesterday the point around what we could specify in the manifest file to have it generated. I think one of these things are the input docs. My thinking is that each input should be able to have its own documentation file. If there is one, we should include it automatically. (in the future).

PR #144, opened 2019-10-31
Discuss: What do we do with meta fields

ruflin:
NOT READY FOR COMMENTING YET!

This PR is to discuss what we should do best with meta fields.

* Not all ECS, some ECS
* Each should define on its own
* But Agent processors can add unexpected fields
* How do we keep in sync?
* Do we have a global template? logs-*, metrics-*


ruflin:
Closing this as a potential new feature called Alias templates which lets us compose templates could solve this.

PR #146, opened 2019-11-11
Add missing assets from datasets

ruflin:
The current code was only going 4 levels deep when looking for assets. But since we introduce datasets more level exists. This code makes sure when the package output is generated it goes into all directories.

Tests were updated with the new package format to also confirm this change.


util/package.go
```diff
@@ -182,22 +182,19 @@ func (p *Package) LoadAssets(packagePath string) (err error) {
...
-		return err
-	}
-	assets = append(assets, a...)
-
-	a, err = filepath.Glob("*/*/*")
-	if err != nil {
-		return err
+	var assets []string
+	var pattern = "*"
+	// Iterates 6 levels deep through the tree to find assets
```
skh: Will there be a time when 6 levels aren't deep enough? Is there a way to solve this recursively once and for all?

ruflin: I'm sure it will happen. There is a library that does it but I was hesitant to include it here: https://github.com/bmatcuk/doublestar I'm just thinking of an alternative recursive implementation now that I mention it which "should" always work. Will update.

ruflin:
@skh I just pushed a new commit to make it recursive. Let me know what you think.

PR #147, opened 2019-11-12
Add format_version to manifest file

ruflin:
The field `format_version` is added to each manifest to indicate which package version format is used. This will become useful if in case one day we will make breaking changes to the structure and can then on the EPM side make a smart decision on how to process the package.

Additional changes:

* All the config reading had to be updated to `config` from `yaml` as it uses ucfg. This should have done before but was missing. The problem only showed up now because of different names of variable and field name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add `/health` and `/health?ready=1` endpoint for healthcheck. [#151](https://github.com/elastic/integrations-registry/pull/151)
 * Add `default` config to dataset manifest. [#148](https://github.com/elastic/integrations-registry/pull/148)
 * Update Golang version to 1.13.4. [#159](https://github.com/elastic/integrations-registry/pull/159)
+* Add missing assets from datasets. [#146](https://github.com/elastic/integrations-registry/pull/146)
+* Add `format_version` to define the package format.
 
 ### Deprecated
 
--- a/dev/package-examples/auditd-2.0.4/manifest.yml
+++ b/dev/package-examples/auditd-2.0.4/manifest.yml
@@ -1,3 +1,5 @@
+format_version: 1.0.0
+
 name: auditd
 title: Auditd
 description: Auditd integration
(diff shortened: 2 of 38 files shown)


ruflin:
@exekias This should be added to you conversion script.

CHANGELOG.md
```diff
@@ -10,8 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Add missing assets from datasets. [#146](https://github.com/elastic/integrations-registry/pull/146)
```
ruflin: This was forgotten in an other PR.

dev/package-examples/auditd-2.0.4/manifest.yml
```diff
@@ -1,3 +1,5 @@
+package_version: 1.0.0
```
exekias: why not using just version? Like we do with `name` instead of `package_name`

ruflin: version is already taken: https://github.com/elastic/package-registry/pull/147/files/4eb902e5e7f2a83334bb2d8d54b76de823d929b0#diff-a2506afb478bf1deffc60f4eb9b4ea26R9

There are 2 versions:
* Version of the package, incremented with each addition
* Version of the package structure, incremented only when there is a change in the package structure (rarely hopefully). So far in Beats we never had to change it: https://github.com/elastic/beats/blob/master/filebeat/module/mongodb/log/manifest.yml#L1

exekias: ah gotcha :+1:

ruflin: well, I'm pretty sure that confusion will pop up again ;-) Suggestion for a better more intuitive name?

ruflin: @exekias WDYT about `package_format` or `package_format_version`?

exekias: both look good to me, also `format_version` as package is implicit

ruflin: +1 on `format_version`, will go with this one.

PR #148, opened 2019-11-12
Add default config to dataset

ruflin:
Not all datasets are enabled by default. Having a config per dataset should allow the UI to decide which ones should be enabled by default. The default should be false.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -276,6 +276,9 @@ type: metric
 # Each input can be in its own release status
 release: beta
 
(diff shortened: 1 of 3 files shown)


ruflin:
@exekias Would be great to get your feedback on this. Not sure about the naming.

ruflin:
@exekias I assume your 👍 means you are ok with the name? Lets go with it then for now.

exekias:
Yes, sorry for not saying that explicitly :)

Issue #150, opened 2019-11-13
Add package with ingest pipelines in yaml format


skh:
Add a sample package that contains at least two ingest pipelines in yaml format, one referencing the other.

PR #152, opened 2019-11-14
Add package with ingest pipelines in YAML

ruflin:
This adds the yamlpipeline package with 3 ingest pipelines in yaml.

Closes https://github.com/elastic/package-registry/issues/150

### DIFF
--- a/dev/package-examples/yamlpipeline-1.0.0/dataset/log/elasticsearch/ingest-pipeline/pipeline-entry.yml
+++ b/dev/package-examples/yamlpipeline-1.0.0/dataset/log/elasticsearch/ingest-pipeline/pipeline-entry.yml
@@ -1,0 +1,113 @@
+---
+description: Pipeline for normalizing Kubernetes CoreDNS logs.
+processors:
+  - pipeline:
+      if: ctx.message.charAt(0) == (char)("{")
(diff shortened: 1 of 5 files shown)


dev/package-examples/yamlpipeline-1.0.0/dataset/log/elasticsearch/ingest-pipeline/pipeline-entry.yml
```diff
@@ -0,0 +1,113 @@
+---
+description: Pipeline for normalizing Kubernetes CoreDNS logs.
+processors:
+  - pipeline:
+      if: ctx.message.charAt(0) == (char)("{")
+      name: '{{IngestPipeline "pipeline-json" }}'
+  - pipeline:
+      if: ctx.message.charAt(0) != (char)("{")
+      name: '{{IngestPipeline "pipeline-plaintext" }}'
```
ruflin: @skh We should still decide if we really want to use `{{IngestPipeline` as our identifier or should use something differently. In Beats we used `{<`.

skh: Then we should probably use `{<` too, because then the pipelines can be used from both projects. We could even do _both_.

PR #158, opened 2019-11-20
Fix typos 'pipline'->'pipeline'.

skh:
Fixes a few typos.


ruflin:
Thanks for the fix, will try to deploy it as soon as the container is built.

PR #165, opened 2019-11-26
Update test data generation with 2 directories

ruflin:
During packaging assets are modified. Because of it must be tested if the input and output is as expected. So far the generated data was written into the same directory as the test packages. Now this does not work anymore because the files are modified. This PR moves the generated data for the testdata into a public directory. This will also help with having nice diffs on what is the new content that is generated.


ruflin:
@jfsiii This also now makes the tests for the generated content of the Kibana assets obsolete as we will see any change directly here.

testdata/public/package/example-0.0.2/elasticsearch/ingest-pipeline/pipeline-entry.json
```diff
@@ -0,0 +1,42 @@
+{
+  "description": "Pipeline for normalizing envoyproxy logs",
+  "processors": [
+    {
+      "pipeline": {
+          "if": "ctx.message.charAt(0) != (char)(\"{\")",
+          "name": "{< IngestPipeline "pipeline-plaintext" >}"
```
jfsiii: Note the red background on this line and others. This is not a valid JSON file/string.

I'll try to find the issue (in Beats repo?) where we discussed this, but this isn't like our recent "will it parse?" or "how do we embed JSON strings inside other JSON strings?". This file will not pass validation on https://jsonlint.com/ or similar tools. That's what the GitHub UI is indicating with the red background

ruflin: Yes, we changed the syntax of this to: https://github.com/elastic/package-registry/blob/master/dev/package-examples/coredns-1.0.1/dataset/log/elasticsearch/ingest-pipeline/pipeline-entry.json#L7

This one didn't get changed as it was lost in the testdata. @skh will bring the topic up again in case we need it differently.

I will open a PR to adjust the above.

skh: This will be rewritten during the ingest pipeline rewriting step. It should be valid JSON afterwards.

The result of the rewrite will look like the output files in https://github.com/elastic/kibana/tree/feature-ingest/x-pack/legacy/plugins/epm/server/lib/tests/ingest_pipelines

Both `{{...}}` and `{<...>}` are currently being accepted, see 

https://github.com/elastic/kibana/blob/feature-ingest/x-pack/legacy/plugins/epm/server/lib/ingest_pipelines.ts#L20-L23

ruflin: @skh Thanks for the details 👍

PR #169, opened 2019-11-27
Prefix package download url with /epr/{package-name}

ruflin:
Prefixing it with `/epr/{package-name}` simplifies the download stats rules. Only the `.tar.gz` packages are move to a different url. An example of a download path is: `/epr/apache/apache-1.0.1.tar.gz`.

The new download path is available in the package output. It is a breaking change but I would expect that from a package manager point of view it should just work as we rely on the download field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Change `requirements.kibana.version.min/max` to `requirements.kibana.versions: {semver-range}`
 * Encode Kibana objects during packaging. [#157](https://github.com/elastic/integrations-registry/pull/157)
+* Prefix package download url with `/epr/{package-name}`.
 
 ### Bugfixes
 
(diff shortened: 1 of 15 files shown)


ruflin:
This is currently blocked by https://github.com/elastic/kibana/pull/51881

ruflin:
@JoshMock Working on this, I realised when we already change this, we could go one step further. Instead of having `/epr/mysql-1.0.1.tar.gz` we could use `/epr/mysql/mysql-1.0.1.tar.gz` instead. On the registry side this means that all `.tar.gz` version for one package are in one directory. This is also more similar to what we have in our current download, for example for Elasticsearch: `https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.4.2-linux-x86_64.tar.gz`

ruflin:
@jfsiii Would be great to get your take on the above too

JoshMock:
That makes total sense to me. Let me know if you go that route so I can adjust expectations on the log-parsing side.

jfsiii:
@ruflin Works for EPM. The location for downloads is stored in one place in the code, so it's not a problem to update that. And when https://github.com/elastic/package-registry/pull/174 lands, EPM won't "know" anything about the location so EPR can change the location anytime it wants without affecting EPM.

ruflin:
@JoshMock Will merge this in a bit. The new download urls look as following:

```
/epr/apache/apache-1.0.1.tar.gz
```

PR #170, opened 2019-11-27
Create basic system package

ruflin:
This system package currently only contains the configurations for the agent. This should be sufficient to set it up.


dev/package-examples/system-0.9.0/dataset/cpu/agent/input/input.yml
```diff
@@ -0,0 +1,4 @@
+- type: metric/system
```
ruflin: @ph I keep coming back that this here should be an input and not a stream. Reason is that a stream combines input + output, but this one here is only input configuration. How do you call it on the code side in the agent?

PR #174, opened 2019-12-02
Add download and path urls

ruflin:
At the moment, only the `/search` endpoint contains the download url. This is now also added to the package info. In addition a field `path` is added which is the path to the package info. This should help with breaking changes in the future so EPM can rely on this information to find the info about a package instead of having to hard code it.


ruflin:
@jfsiii I share the exact same sentiment around naming and potentially having a better structure. Putting it in an object could be helpful, need to think more about this in the context of all fields. 

For now will merge it and I kind of hope that if we add more fields it will become clear where things belong.

PR #177, opened 2019-12-05
Add Makefile with Docker publishing targets

JoshMock:
Adds a Make target to build and publish images to Elastic's container registry so I can take advantage of it with an upcoming CI job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+TAG := $(shell date +'%Y-%m-%d')-$(shell git log --pretty=format:%h -1)
+REPO := package-registry:$(TAG)
+DOCKER_PATH := push.docker.elastic.co/ruflin/$(REPO)
+
+build:
+	docker build -t $(REPO) .
+
+publish: build
+	docker tag $(REPO) $(DOCKER_PATH)
+	docker push $(DOCKER_PATH)
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 TAG := $(shell date +'%Y-%m-%d')-$(shell git log --pretty=format:%h -1)
 REPO := package-registry:$(TAG)
-DOCKER_PATH := push.docker.elastic.co/ruflin/$(REPO)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,10 @@
+TAG := $(shell date +'%Y-%m-%d')-$(shell git log --pretty=format:%h -1)
+REPO := package-registry:$(TAG)
+DOCKER_PATH := push.docker.elastic.co/employees/ruflin/$(REPO)
```
ruflin: Any suggestion on what path we could use here to not tied this to me? I see elasticsearch is under `docker.elastic.co/elasticsearch/elasticsearch:7.5.0` so I wonder if we should do something like `docker.elastic.co/package-registry/package-registry:{TAG}`.

JoshMock:
Ahh, got it. Didn't notice you were using `mage` as a stand-in for `make`.

Makefile
JoshMock: It depends: if this is allowed to be a public image, we could put it in any public namespace you think is appropriate; if it should remain private, we can spin up a single private namespace for it, or we can pick an existing private namespace where other related things already live.

Let me know what you'd prefer and we can move it accordingly.

ruflin: This is not only allowed to be a public image, it must be a public image :-) What is your preferred namespace?

JoshMock: That makes things easier for sure! I noticed you're using `beats-ci` for automation, but I'm inclined to say that the `beats` namespace is probably not a right fit. Unless [one of these](https://container-library.elastic.co/) listed on the homepage doesn't look like a good fit to you, we can always create a public `package-registry` namespace.

ruflin: Yes, it is just in `beats` because that was easiest for me to get it in quickly. Happy to move to any other CI group, suggestions?

I don't think any of the listed groups will fit the registry. Lets create a new one.

JoshMock: Great. I'll open a ticket and tag you on it.

PR #180, opened 2019-12-10
Fix nginx dashboards and remove ML viz

ruflin:
The dashboard was just copied over as a bulk object from nginx which doesnt work with EPM. In additiona, non encoded visualizations from ML were copied over. These were removed.


jfsiii:
My apologies for merging. I saw the approval and was eager to see the fix in EPM. However, it still needs to be deployed so me merging didn't change anything from EPM POV.

ruflin:
New registry version should be available in a few minutes.

PR #182, opened 2019-12-16
Cleanup nginx package

ruflin:
This cleans up the nginx package by adding the missing fields like title, ingest pipeline etc.

In addition, the field title is made a requirement to make sure it is always set.

The ecs.yml field is added to the logs part to have the ecs fields around for now. Long term the fields that are actually used have to be added.


dev/package-examples/nginx-1.2.0/dataset/error/manifest.yml
```diff
@@ -1,4 +1,6 @@
 title: Nginx Error Logs
+type: logs
+ingest_pipline: pipeline
```
exekias: would it make sense to have a default name for the main pipeline? it would make this setting optional I guess

ruflin: yes, we could definitively do that, for example call it `default.json`. And then on package building, it checks if it is there. +1 on this. Could you file an issue for it?

exekias: https://github.com/elastic/package-registry/issues/183

util/package.go
```diff
@@ -68,11 +68,11 @@ type Image struct {
...
 
 type DataSet struct {
-	Title          string                   `config:"title" json:"title"`
+	Title          string                   `config:"title" json:"title" validate:"required"`
 	Name           string                   `config:"name" json:"name"`
 	Release        string                   `config:"release" json:"release"`
-	Type           string                   `config:"type" json:"type"`
-	IngestPipeline string                   `config:"ingest_pipeline" config:"ingest_pipeline" json:"ingest_pipeline"`
+	Type           string                   `config:"type" json:"type" validate:"required"`
+	IngestPipeline string                   `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty"`
```
jfsiii: We'll need to update the `Dataset` type in EPM so [`ingest_pipeline`](https://github.com/elastic/kibana/blob/f51a5a935161358d53761c2b651996874e4e336d/x-pack/legacy/plugins/epm/common/types.ts#L123) is optional `ingest_pipeline?`

ruflin: @jfsiii Good catch, can you take care of this?

PR #184, opened 2019-12-17
Rename stream files to input in agent

ruflin:
A stream consists of an input and an output. What we have configured here is only the input section. Because of this I renamed it to input (again).


ruflin:
@ph @skh FYI

ruflin:
Going to merge this as I need it for the development on the Kibana side to generate the config.

PR #188, opened 2019-12-17
Validate dataset to check for pipeline config

ruflin:
This PR checks a dataset manifest to make sure the config is as expected. If no pipeline is specified, it checks if pipeline files are there. If yes, it checks if a `default.json` or `default.yml` are there and takes it as the ingest pipeline. If files are there with different names but no default, it is an invalid dataset and an error is throw. The goal is to make sure, only valid packages are shipped.

Additional changes:

* Move dataset code to its own file
* Use `Validate` magic from go-ucfg
* Introduce package to test for default pipeline behaviour
* Add check that if no Kibana version is specified, it returns true by default

### DIFF
--- a/testdata/public/package/default-pipeline-0.0.2/index.json
+++ b/testdata/public/package/default-pipeline-0.0.2/index.json
@@ -1,0 +1,30 @@
+{
+  "name": "default-pipeline",
+  "title": "Default pipeline Integration",
+  "version": "0.0.2",
+  "description": "Tests if no pipeline is set, it defaults to the default one",
+  "type": "integration",
+  "categories": [
+    "logs"
+  ],
+  "requirement": {
+    "kibana": {}
+  },
+  "assets": [
+    "/package/default-pipeline-0.0.2/manifest.yml",
+    "/package/default-pipeline-0.0.2/dataset/foo/manifest.yml",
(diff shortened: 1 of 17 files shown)


testdata/public/package/default-pipeline-0.0.2/index.json
```diff
@@ -0,0 +1,30 @@
...
+    "/package/default-pipeline-0.0.2/dataset/foo/elasticsearch/ingest-pipeline/default.json"
+  ],
+  "format_version": "1.0.0",
+  "datasets": [
+    {
+      "title": "Foo",
+      "name": "foo",
+      "release": "beta",
+      "type": "logs",
+      "ingest_pipeline": "default"
```
ruflin: @skh This PR made me think: Should I return here `default.json` or `default.yml`. This work could be done automatically by the registry for you. It is only in the manifest where we simplify things for the user.

skh:
> Should I return here default.json or default.yml

It's not necessary, and we would cut off the extension from the name anyway.

ruflin:
Ok, will leave it then for now 👍

ruflin:
@skh Thanks for the testing. Please keep bring up things you would like to see validated on a dataset / package so we can implement it to make sure EPM does not have to deal with edge cases.

PR #189, opened 2019-12-17
Add package name to dataset

ruflin:
The package name in the dataset is useful in EPM. For this reason it is added here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add dataset array to package info endpoint. [#162](https://github.com/elastic/integrations-registry/pull/162)
 * Add path field to search and package info endpoint. [#174](https://github.com/elastic/integrations-registry/pull/174)
 * Add download field to package info endpoint. [#174](https://github.com/elastic/integrations-registry/pull/174)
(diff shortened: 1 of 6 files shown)


ruflin:
@jfsiii I have the suspicion that later on `package.version` will also become useful, not only the name. For now I only added the name, but later we could convert it to:

```
{ 
  "package": {
    "name": "foo",
    "version": "1.2.3"
  }
}
```

ruflin:
@jfsiii Could you help me to get the necessary changes ready on the EPM side if you are good with the change?

jfsiii:
@ruflin sure thing. Sorry for the lag. Thanks for the ping.

We already have [`Dataset.package`](https://github.com/elastic/kibana/blob/c154afc02b181daf9f0337edb4a2a5b8a79744cd/x-pack/legacy/plugins/epm/common/types.ts#L133) so I think the only work is removing the code that manually inserts it. I'll post a PR soon.

ruflin:
Merged. Will redeploy after the demo today ;-)

jfsiii:
I just remembered that @neptunian has a PR that changes this to `packageName`. https://github.com/elastic/kibana/pull/54037

So that should either be reverted to `package` or the registry should use `packageName` so there's agreement

ruflin:
Good point, lets stick to `package` for now if @neptunian is ok with it.

Issue #195, opened 2020-01-10
Generate/publish Typescript types


jfsiii:
EPM maintains TS types for the Registry API responses https://github.com/elastic/kibana/blob/0393fea11c57713f6a9523ef463ff03decb912d2/x-pack/legacy/plugins/epm/common/types.ts#L41-L61 but it requires an individual to watch the Registry commits and be sure to update them on the EPM side. This is burdensome and unreliable. See https://github.com/elastic/kibana/pull/52285 which added several properties which were missing entirely and marked others as optional which had been marked as required.

Perhaps EPR could generate TS types as part of a build/release process like it does with the JSON docs from https://github.com/elastic/package-registry/pull/14

There are a few projects which do this sort of thing. Some I found are

- https://github.com/OneOfOne/struct2ts
- https://github.com/tkrajina/typescriptify-golang-structs
- https://github.com/zmitry/go2typings

I tried `struct2ts` and pointed it at the `util.Package` struct and it produced something quite useful (although there was [one issue](https://github.com/OneOfOne/struct2ts/issues/24) with the output)

```bash
struct2ts -i -m -H util.Package
```

```typescript
// this file was automatically generated, DO NOT EDIT
// classes
// struct2ts:github.com/elastic/package-registry/util.PackageRequirementKibana
interface PackageRequirementKibana {
	versions?: string;
}

// struct2ts:github.com/elastic/package-registry/util.PackageRequirement
interface PackageRequirement {
	kibana: PackageRequirementKibana;
}

// struct2ts:github.com/elastic/package-registry/util.PackageImage
interface PackageImage {
	src?: string;
	title?: string;
	size?: string;
	type?: string;
}

// struct2ts:github.com/elastic/package-registry/util.PackageDataSet
interface PackageDataSet {
	title: string;
	name: string;
	release: string;
	type: string;
	ingest_pipeline?: string;
	vars?: map[string]interface {}[];
	package: string;
}

// struct2ts:github.com/elastic/package-registry/util.Package
interface Package {
	name: string;
	title?: string | null;
	version: string;
	readme?: string | null;
	description: string;
	type: string;
	categories: string[];
	requirement: PackageRequirement;
	screenshots?: PackageImage[];
	icons?: PackageImage[];
	assets?: string[];
	internal?: boolean;
	format_version: string;
	datasets?: PackageDataSet[];
	download: string;
	path: string;
}

// exports
export {
	PackageRequirementKibana,
	PackageRequirement,
	PackageImage,
[...]

PR #196, opened 2020-01-15
Create TS types in `mage build` #195

jfsiii:
Uses https://github.com/OneOfOne/struct2ts to generate `types.d.ts` file from `util.Package` struct.

Added the minimum features required to support consuming the TS types in EPM. We can build on this with a proper build/release/publish process, new/improved types, etc.

Will add more notes inline


types.js
```diff
@@ -0,0 +1,2 @@
+"use strict";
```
jfsiii: Added the bare minimum required to get EPM to be able to `import '@elastic/package-registry'.`

This is the result when did things [The Right Way](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) locally and avoids bringing in TypeScript, running `tsc --declaration ...`, etc each time.

We can move to that later. For now, new golang structs will still be written to `types.d.ts` and available to consumers just by running `mage build`

package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@elastic/package-registry",
+  "version": "0.2.0",
+  "description": "Elastic Package Registry (EPR)",
+  "repository": "git@github.com:elastic/package-registry.git",
+  "author": " Nicolas Ruflin <spam@ruflin.com>",
+  "license": "Elastic-License",
```
jfsiii: `npm`/`yarn` print a warning because this is not available in the [SPDX License List](https://spdx.org/licenses/). Perhaps we could create a PR to be included. [Mongo's SSPL is in the list](https://spdx.org/licenses/SSPL-1.0.html)

package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@elastic/package-registry",
+  "version": "0.2.0",
```
jfsiii: Ideally this would be linked to the `version` in `BuildRootFile` so they're always the same

package.json
```diff
@@ -0,0 +1,10 @@
...
+  "description": "Elastic Package Registry (EPR)",
+  "repository": "git@github.com:elastic/package-registry.git",
+  "author": " Nicolas Ruflin <spam@ruflin.com>",
+  "license": "Elastic-License",
+  "types": "./types.d.ts",
```
jfsiii: This is the file written by `BuildTypesFile`. It's currently possible to update the magefile and break the package but I left that linking for a later phase.

package.json
```diff
@@ -0,0 +1,10 @@
+{
```
ruflin: Do we need this on the top level? I would prefer to not have generated files in the root.

package.json
ruflin: Or we make the definition itself Apache licensed?

types.d.ts
```diff
@@ -0,0 +1,59 @@
+// this file was automatically generated, DO NOT EDIT
```
ruflin: I don't like the location of this file. Could we put it with package under `npm` or `js` or a similar directory?

types.js
ruflin: Also going into a sub directory?

package.json
jfsiii: Yes, `package.json` needs to be at the root. 

I don't think this is a generated file. Or at least it doesn't change as the result of a `mage` command, for example.

package.json
jfsiii: Maybe when/if we split types out to their own project/repo. Or we should change the name/files. This is called `@elastic/package-registry` and actually brings in all the other repo files so the Elastic license is appropriate for what it's currently doing.

types.d.ts
jfsiii: I initially put it in `docs` because I forgot the second part of this comment. I just moved it to a new `types` directory.

types.js
jfsiii: same as https://github.com/elastic/package-registry/pull/196#discussion_r367435696

package.json
ruflin: 👍

package.json
ruflin: We probably will have to generate it. Or does it allow linking here?

package.json
ruflin: having package.json at the root SGTM

package.json
```diff
@@ -0,0 +1,10 @@
...
+  "name": "@elastic/package-registry",
+  "version": "0.2.0",
+  "description": "Elastic Package Registry (EPR)",
+  "repository": "git@github.com:elastic/package-registry.git",
+  "author": " Nicolas Ruflin <spam@ruflin.com>",
+  "license": "Elastic-License",
+  "types": "./types/types.d.ts",
+  "main": "./types/types.js"
```
ruflin: I see we have the path specified here. So if we ever move it around in the repo, we can just update it here 👍

package.json
jfsiii: What I was imagining was BuildRootFile would read package.json to get the version.

package.json would be managed with [`npm version`](https://docs.npmjs.com/cli/version) command or something similar. 

The root file would remain synced and the build script doesn't have to take on knowledge of building another artifact. 

Version would gain a source of truth that can be read by other tools instead of being inside a script.

ruflin: +1 on having it in a file. I wonder if `package.json` is the right place for it considering it is a Golang tool, but in the end it is just JSON, so ... Will figure it out as soon as we automate it.

ruflin:
@jfsiii What should we do with this PR? Should we close it for now or get it in?

ph:
@jfsiii @ruflin should we close this?

ruflin:
This ties into https://github.com/elastic/kibana/pull/54955#issuecomment-633823711 Same suggestion, close it but have an issue in Kibana that tracks this idea and also references both PR's.

PR #202, opened 2020-01-23
Work in progress: First pass of of endpoint package (phase 0)

james-elastic:
This is a work in progress.

This is using the phase 0 data sets and schemas.  I'm looking for general feedback to make sure the package structure and layouts look correct and general comments.


dev/package-examples/endpoint-0.0.1/elasticsearch/index-template/endpoint-index-template.json
```diff
@@ -0,0 +1,1663 @@
+{
```
ruflin: We generate the index template in the package manager from the `fields.yml` files. This has the advantage that we can generate the template and the Kibana index pattern for it. And YAML allows to add a bit more comments.

dev/package-examples/endpoint-0.0.1/kibana/dashboard/endpoint-dashboard-2.json
```diff
@@ -0,0 +1 @@
...   \"type\": \"linear\"\n        },\n        \"show\": true,\n        \"style\": {},\n        \"title\": {\n          \"text\": \"Event Count\"\n        },\n        \"type\": \"value\"\n      }\n    ]\n  },\n  \"aggs\": [\n    {\n      \"id\": \"1\",\n      \"enabled\": true,\n      \"type\": \"count\",\n      \"schema\": \"metric\",\n      \"params\": {\n        \"customLabel\": \"Event Count\"\n      }\n    },\n    {\n      \"id\": \"2\",\n      \"enabled\": true,\n      \"type\": \"terms\",\n      \"schema\": \"segment\",\n      \"params\": {\n        \"field\": \"host.name\",\n        \"order\": \"desc\",\n        \"size\": 5,\n        \"orderBy\": \"1\",\n        \"otherBucket\": false,\n        \"otherBucketLabel\": \"Other\",\n        \"missingBucket\": false,\n        \"missingBucketLabel\": \"Missing\",\n        \"customLabel\": \"Host Name\"\n      }\n    }\n  ]\n}"},"id":"445f5c50-f669-11e9-8344-2f4cf656fceb","migrationVersion":{"visualization":"7.2.0"},"references":[{"id":"endgame-dashboard-index-pattern","name":"kibanaSavedObjectMeta.searchSourceJSON.index","type":"index-pattern"}],"type":"visualization","updated_at":"2019-10-25T16:56:28.618Z","version":"WzE2NCwxXQ=="}
```
ruflin: We store all the saved objects "decoded" for better versioning. I need to share the tool through which I decode it.

dev/package-examples/endpoint-0.0.1/manifest.yml
```diff
@@ -0,0 +1,17 @@
+format_version: 1.0.0
+name: endpoint
+title: Elastic Endpoint
```
tonymeehan: Are we reusing this package for Phase 1? If so, then this is probably the right name. But if we expect these will break in Phase 1, then it might be important for us to consider whether we adjust the name so that it's clear it's for Elastic Endpoint Security.

ruflin: I expect this to be the same package which is shipped also for v1, so we only have 1 endpoint package.

tonymeehan:
I just saw the EPM mocks from the Observability update but didn't see a filter for Endpoint. Where will this show up on the Available Packages page for EPM?

ruflin:
@tonymeehan Should we introduce a category "endpoint"?

dev/package-examples/endpoint-0.0.1/elasticsearch/index-template/endpoint-index-template.json
jonathan-buttner: @ruflin @jen-huang I'm having some trouble finding the tool that handles parsing the fields.yml files. Could you point me to where that tool lives?

dev/package-examples/endpoint-0.0.1/kibana/dashboard/endpoint-dashboard-2.json
jonathan-buttner: By "decoded" do you mean in regular json format as opposed to ndjson?
Looks like these files: https://github.com/elastic/package-registry/blob/master/dev/package-examples/nginx-1.2.0/kibana/dashboard/metricbeat-nginx-overview.json are in regular json format.

dev/package-examples/endpoint-0.0.1/elasticsearch/index-template/endpoint-index-template.json
jonathan-buttner: Sonja mentioned these places:

currently, generation of index templates happens here https://github.com/elastic/kibana/blob/feature-ingest/x-pack/legacy/plugins/epm/server/lib/elasticsearch/template/install.ts , and of index patterns here: https://github.com/elastic/kibana/blob/feature-ingest/x-pack/legacy/plugins/epm/server/lib/kibana/index_pattern/install.ts

dev/package-examples/endpoint-0.0.1/manifest.yml
jonathan-buttner: @tonymeehan I believe this package is just an example to get discussion going. I don't think we'll have a package for the phase 0 work. I'm working on the phase 1 package.

tonymeehan:
@ruflin Good question, I'm curious to hear what @bfishel or @lindseypoli think about that.

bfishel:
> @ruflin Good question, I'm curious to hear what @bfishel or @lindseypoli think about that.

I believe this might be a thing (I like it!) but @hbharding will know for sure

dev/package-examples/endpoint-0.0.1/manifest.yml
james-elastic: @tonymeehan yeah this was to start conversations

ruflin:
@james-elastic @jonathan-buttner I think you asked in the past what kind of tooling I use for the dashboards to export them. I shared the code here: https://github.com/elastic/package-registry/pull/214 I still needs cleanup but hope this gives some idea on how I created some of the current packages.

ruflin:
@james-elastic @jonathan-buttner Should we close this PR in favour of https://github.com/elastic/package-registry/pull/223 ?

jonathan-buttner:
@ruflin yep, sorry about that. Closing.

PR #211, opened 2020-02-11
Validate dataset name to not contain -

ruflin:
Because of the indexing strategy, the dataset name is not allowed to contain a -

### DIFF
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type DataSet struct {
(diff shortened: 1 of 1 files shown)


skh:
To test this, I tried (separately):

- in the `system-0.9.0` package, rename the directory `dataset/cpu` to `dataset/cpu-test`
- in the `system-0.9.0` package, in `dataset/cpu/manifest.yml`, add a line `name: cpu-test`.

After both steps, run `mage build`, run `go run .` and check the output http://localhost:8080/package/system-0.9.0

In both cases, the output of the registry contained
```
"datasets": [
    {
      "title": "System CPU",
      "name": "cpu-test",
      "release": "beta",
      "type": "metrics",
      "package": "system"
    },
    ...
  ]
```

Please note the dataset name with the `-` in it.

I haven't looked at the code. Am I testing this incorrectly?

ruflin:
@skh Yes, you tested this correctly. It helps putting the params in the right order on my end 🤦‍♂ https://github.com/elastic/package-registry/pull/211/commits/155d88606a41bc97482dad21dc5da210d0f93a43

ruflin:
Thanks for testing all these cases manually. We should come up with a way in the future to also test invalid package examples. At the moment the test framework does not allow this :-(

PR #212, opened 2020-02-11
Introduce datasources in package to configure inputs and streams

ruflin:
In https://github.com/elastic/beats/pull/15940 datasources, inputs and streams are introduced into the agent config. To make it possible to configure these in the UI and through the API, some changes to the manifest definitions of a package and datasets are needed.

**Package manifest**

Each package must specify the datasources it supports with the supported inputs inside. So far all the packages only support one datasource but I want to keep the door open for this to potentially change in the future. It also makes it possible to have the manifest config of a datasource be identical to the config which ends up in the agent config.

The package manifest datasource definition looks as following (nginx example):

```
datasources:
  -
    # Do we need a name for the data source?
    name: nginx

    # List of inputs this datasource supports
    inputs:
      -
        # An id can be given, in case the type used here is not unique
        # This is for selection in the stream
        # id: nginx
        type: metrics/nginx

        # Common configuration options for this input
        vars:
          - name: hosts
            description: Nginx hosts
            default:
              ["http://127.0.0.1"]
            # All the config options that are required should be shown in the UI
            required: true
          - name: period
            description: "Collection period. Valid values: 10s, 5m, 2h"
            default: "10s"
          - name: username
            type: text
          - name: password
            # This is the html input type?
            type: password

      -
        type: logs

        # Common configuration options for this input
        vars:

      -
        type: syslog

        # Common configuration options for this input
        vars:

```

Inside the datasource, the supported inputs are specified with the common variables across all streams which use a certain input. In the UI I expect that we show the `required` configs by default and all the others are under "Advanced" or similar.

**Dataset manifest**

With the datasources and inputs defined on the package level, each dataset can specify which inputs it supports. Most datasets will only support one input for now. For the nginx metrics this looks as following:

```
inputs:
  - type: "metric/nginx"

    # Only the variables have to be repeated that are not specified as part of the input
    vars:
      # All variables are specified in the input already
```

As an example with supporting multiple inputs, we have the nginx error logs:

```
inputs:
  - type: log
    vars:
      - name: paths
        required: true
        default:
          - /var/log/nginx/error.log*
        os.darwin:
          - /usr/local/var/log/nginx/error.log*
        os.windows:
          - c:/programdata/nginx/logs/error.log*

  - type: syslog
    vars:
      # Are udp and tcp syslog input two different inputs?
      - name: protocol.udp.host
        required: true
        default:
          - "localhost:9000"
```

The log and syslog input are supported (not the case today, just an example). One the dataset level also all additional variables for this dataset are specified. The ones already specified on the input level in the package don't have to be specified again.

**Stream definition**

Now that the dataset has its supported inputs and variables defined, the stream can be defined. The stream defines which input it uses from the dataset and its configuration variables. Here an example for nginx metrics:

```
input: metrics/nginx
metricsets: ["stubstatus"]
period: {{period}}
enabled: true

hosts: {{hosts}}

{{#if username}}
username: "{{username}}"
{{/if}}
{{#if password}}
password: "{{password}}"
{{/if}}
```

During creation time of the stream config the variables from the datasource inputs and local variables from the dataset are filled in.

A stream definition could also support multiple inputs as seen in the following example:

```

{{#if input == log}}
input: log

{{#each paths}}
paths: "{{this}}"
{{/each}}
exclude_files: [".gz$"]

processors:
  - add_locale: ~
{{/if}}

{{#if input == syslog}}
input: syslog

{{/if}}
```

**Further changes**

* Rename `agent/input` to `agent/stream` as a stream is configured there.

### DIFF
--- a/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
@@ -8,18 +8,22 @@ type: logs
 # The ingest pipeline which should be used.
 ingest_pipeline: default
 
-vars:
-  - name: paths
-    # Should we define this as array? How will the UI best make sense of it?
-    type: textarea
-    default:
-      - /var/log/nginx/access.log*
-    # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
-    # This would need to be based on a predefined definition on what can be filtered on
-    os.darwin:
-      - /usr/local/var/log/nginx/access.log*
-    os.windows:
-      - c:/programdata/nginx/logs/*access.log*
+# List of supported inputs
+inputs:
+  - type: log
+    vars:
+      - name: paths
+        required: true
+        # Should we define this as array? How will the UI best make sense of it?
+        type: textarea
+        default:
+          - /var/log/nginx/access.log*
+        # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
+        # This would need to be based on a predefined definition on what can be filtered on
+        os.darwin:
+          - /usr/local/var/log/nginx/access.log*
+        os.windows:
+          - c:/programdata/nginx/logs/*access.log*
 
 requirements:
   elasticsearch.processors:
--- a/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
@@ -15,7 +15,8 @@ inputs:
       - name: paths
         required: true
         # Should we define this as array? How will the UI best make sense of it?
-        type: textarea
+        type: text
+        multi: true
         default:
           - /var/log/nginx/access.log*
         # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
--- a/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
@@ -15,6 +15,7 @@ inputs:
       - name: paths
         required: true
         # Should we define this as array? How will the UI best make sense of it?
+        description: Paths to the nginx access log file.
         type: text
         multi: true
         default:
@@ -22,8 +23,10 @@ inputs:
         # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
         # This would need to be based on a predefined definition on what can be filtered on
         os.darwin:
+          default:
           - /usr/local/var/log/nginx/access.log*
         os.windows:
+          default:
           - c:/programdata/nginx/logs/*access.log*
 
 requirements:
--- a/dev/package-examples/nginx-1.2.0/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/manifest.yml
@@ -24,7 +24,7 @@ datasources:
         # An id can be given, in case the type used here is not unique
         # This is for selection in the stream
         # id: nginx
-        type: metrics/nginx
+        type: nginx/metrics
 
         # Common configuration options for this input
         vars:
--- a/dev/package-examples/nginx-1.2.0/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/manifest.yml
@@ -34,9 +34,12 @@ datasources:
               ["http://127.0.0.1"]
             # All the config options that are required should be shown in the UI
             required: true
+            multi: true
+            type: text
           - name: period
             description: "Collection period. Valid values: 10s, 5m, 2h"
             default: "10s"
+            type: duration
           - name: username
             type: text
           - name: password
--- a/dev/package-examples/nginx-1.2.0/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/manifest.yml
@@ -12,3 +12,45 @@ requirement:
     versions: ">7.1.0 <7.6.0"
   elasticsearch:
     versions: ">7.0.1"
+
+datasources:
+  -
+    # Do we need a name for the data source?
+    name: nginx
+
+    # List of inputs this datasource supports
+    inputs:
+      -
+        # An id can be given, in case the type used here is not unique
+        # This is for selection in the stream
+        # id: nginx
+        type: metrics/nginx
+
+        # Common configuration options for this input
+        vars:
+          - name: hosts
+            description: Nginx hosts
+            default:
+              ["http://127.0.0.1"]
+            # All the config options that are required should be shown in the UI
+            required: true
+          - name: period
+            description: "Collection period. Valid values: 10s, 5m, 2h"
+            default: "10s"
+          - name: username
+            type: text
+          - name: password
+            # This is the html input type?
+            type: password
+
+      -
+        type: logs
+
+        # Common configuration options for this input
+        vars:
+
+      -
+        type: syslog
+
+        # Common configuration options for this input
+        vars:
--- a/dev/package-examples/nginx-1.2.0/manifest.yml
+++ b/dev/package-examples/nginx-1.2.0/manifest.yml
@@ -25,6 +25,7 @@ datasources:
         # This is for selection in the stream
         # id: nginx
         type: nginx/metrics
+        descrition: Collecting metrics for nginx.
 
         # Common configuration options for this input
         vars:
@@ -48,6 +49,7 @@ datasources:
 
       -
         type: logs
+        description: Collect nginx logs.
 
         # Common configuration options for this input
         vars:
--- a/dev/package-examples/nginx-1.2.0/dataset/access/agent/stream/stream.yml
+++ b/dev/package-examples/nginx-1.2.0/dataset/access/agent/stream/stream.yml
@@ -1,6 +1,5 @@
-# This is not an array on purpose to make sure only 1 single input is specified in this file.
 
-type: log
+input: log
 paths:
 {{#each paths}}
   - "{{this}}"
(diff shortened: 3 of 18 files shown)


ruflin:
@skh Could you give some feedback if the above info the the manifests and stream is enough to build a full config?
@jen-huang Will this info be sufficient to build the UI on it?
@ph Applying what we discussed in the config.

skh:
> Could you give some feedback if the above info the the manifests and stream is enough to build a full config?

Where can I find out what a full, correct agent config must look like?

ruflin:
@skh https://github.com/elastic/beats/pull/15940

ph:
@ruflin format sound good, a few things:

- Make sure you rename the type for metrics where the service is before metrics like this: docker/metrics.
- As we discussed over zoom for simplification, lets make sure we do not have a field defined at the level of the input that can be redefined at the stream level.

dev/package-examples/nginx-1.2.0/manifest.yml
```diff
@@ -12,3 +12,45 @@ requirement:
...
+        type: metrics/nginx
+
+        # Common configuration options for this input
+        vars:
+          - name: hosts
+            description: Nginx hosts
+            default:
+              ["http://127.0.0.1"]
+            # All the config options that are required should be shown in the UI
+            required: true
```
jen-huang: I wold like to see `type` defined explicitly for all vars. And if we like the `multi` suggestion in my previous comment, these should be added:
```yml
type: text
multi: true
```

dev/package-examples/nginx-1.2.0/manifest.yml
```diff
@@ -12,3 +12,45 @@ requirement:
...
+          - name: period
+            description: "Collection period. Valid values: 10s, 5m, 2h"
+            default: "10s"
```
jen-huang: Add `type: duration` or `type: text`?

dev/package-examples/nginx-1.2.0/manifest.yml
```diff
@@ -12,3 +12,45 @@ requirement:
...
+          - name: username
+            type: text
+          - name: password
+            # This is the html input type?
+            type: password
```
jen-huang: I think we need to decide all possible values for `type` field. They do not need to be HTML types, though. Here are the possible types I currently see (or can foresee being useful): 
```
text -> maps to UI text input
number -> maps to UI number input
boolean -> maps to UI checkbox
duration -> maps to UI text input
  this would let us do client-side validation but maybe overkill for now?
url -> maps to UI text input
  this would let us do client-side validation but maybe overkill for now?
```
In the case of password vars, I would use `text` for them as they'll show up as simple text inputs in the UI.

dev/package-examples/nginx-1.2.0/manifest.yml
ph: +1 for a duration type, we used integer priviously and it was not that great. :)

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
```diff
@@ -8,18 +8,22 @@ type: logs
...
+# List of supported inputs
+inputs:
+  - type: log
+    vars:
+      - name: paths
+        required: true
+        # Should we define this as array? How will the UI best make sense of it?
+        type: textarea
+        default:
+          - /var/log/nginx/access.log*
```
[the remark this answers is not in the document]
ruflin: Yes, there can be multiple default paths. Changed it do your suggestion.

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
```diff
@@ -8,18 +8,22 @@ type: logs
...
+        # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
+        # This would need to be based on a predefined definition on what can be filtered on
+        os.darwin:
+          - /usr/local/var/log/nginx/access.log*
+        os.windows:
+          - c:/programdata/nginx/logs/*access.log*
```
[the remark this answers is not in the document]
ruflin: Yes, good point. They are platform specific defaults. As we should not mix array and object I wonder how the yaml should look like:

        default:
           all:
              - /var/log/nginx/access.log*
        # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
        # This would need to be based on a predefined definition on what can be filtered on
          os.darwin:
            - /usr/local/var/log/nginx/access.log*
          os.windows:
            - c:/programdata/nginx/logs/*access.log*
```

I don't like the all. Perhaps we find a better name? And should we call it `defaults` instead of `default`?

File: dev/package-examples/nginx-1.2.0/manifest.yml [hunk 6]
```diff
@@ -12,3 +12,45 @@ requirement:
     versions: ">7.1.0 <7.6.0"
   elasticsearch:
     versions: ">7.0.1"
+
+datasources:
+  -
+    # Do we need a name for the data source?
+    name: nginx
```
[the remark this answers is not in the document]
ruflin: You mean prefixing the id? Because the id needs to be unique.

dev/package-examples/nginx-1.2.0/manifest.yml
ruflin: Agree. Adding.

dev/package-examples/nginx-1.2.0/manifest.yml
ruflin: I like duration if that works for the UI.

dev/package-examples/nginx-1.2.0/manifest.yml
ruflin: I like the above proposed types. I was thinking of password instead of text to not have to show clear text passwords directly on the screen when users put them in.

ruflin: BTW: If we have these predefined values for type we can also validate it already on the package side to make sure nothing odd shows up in the package itself.

ph: For the _URL_ types we have to be careful when we do this and we do it, in some modules we do use compound scheme (not sure if this is the exact term?) Like "http+npipe://./pipe/custom" see https://github.com/elastic/beats/blob/d75261469b5c2e675d7d331b86bfc8599d797ddc/metricbeat/mb/parse/url_test.go for a few examples.

ruflin: I see url support as a stretch goal. In case of doubt we can always fall back to just string. Lets keep it simple for now.

jen-huang: We can add granular text types (duration, url, etc) but it doesn't mean that we need to do validation right now. Validation can be added later on from all fronts (UI, agent, package) once we align on rules.

jen-huang: Also ++ on keeping password type. This give the UI information to decide when to show clear text inputs and when to obfuscate.

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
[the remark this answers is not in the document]
jen-huang: I'm not fond of `all` either, and it seems silly to do `default.default` 🙂 Thinking about this again, what if we just nest `default` under each platform? I don't know if we'll want to add other platform-specific configuration for vars, but this schema could support that too since it essentially mirrors all the top-level vars properties. As an example, maybe `paths` is not required for `windows` for whatever reason:

```yml
        default:
            - /var/log/nginx/access.log*
        os.darwin:
          default:
            - /usr/local/var/log/nginx/access.log*
        os.windows:
          required: false
          default:
            - c:/programdata/nginx/logs/*access.log*
```

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
```diff
@@ -8,18 +8,23 @@ type: logs
...
-      - /var/log/nginx/access.log*
-    # I suggest to use ECS fields for this config options here: https://github.com/elastic/ecs/blob/master/schemas/os.yml
-    # This would need to be based on a predefined definition on what can be filtered on
-    os.darwin:
-      - /usr/local/var/log/nginx/access.log*
-    os.windows:
-      - c:/programdata/nginx/logs/*access.log*
+# List of supported inputs
+inputs:
+  - type: log
```
jen-huang: @hbharding suggested today that it would be great to have descriptions in the UI. Could we add `description` prop for every input, under its `type`? I think these descriptions should only be a sentence or two, max. Not sure how we will localize these though..

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
```diff
@@ -8,18 +8,23 @@ type: logs
...
+    vars:
+      - name: paths
```
jen-huang: Similar to my previous comment, it would be great to have `description` prop per var too. This would be really short descriptions that give the user info about what each var is used for.

dev/package-examples/nginx-1.2.0/dataset/access/manifest.yml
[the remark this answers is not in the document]
ruflin: LGTM. So in summary: All configs from the top level could also be nested under each platform. Lets try it.

What I suggest for now, lets ignore platform at least on the UI side.

dev/package-examples/nginx-1.2.0/manifest.yml
ruflin: As soon as we started the UI implementation and are happy with the format, we should open a PR here with the exact specs.

ph:
@ruflin I am ok to merge this and iterate.

ruflin:
I merged this change as it seems we are mostly happy with what we have here and will follow up with more changes for further discussions.

PR #214, opened 2020-02-17
Dashboard exporting tool

ruflin:
To build the current packages, I used some tooling code on my end to select the correct saved objects from Kibana and decode them. In this PR I want to share the code I use.

The code is a bit messy as it is copied out of the project I initially built in Golang to load assets but now all happens in EPM. This simplifies the code here but still needs clean.

The code needs in input yaml file with all the assets defined (apache.yml as an example) and then a target dir to be specified which is normally the package. The command then looks like the following:

```
go run . -target=/path/to/my/package -input="./package-registry/dev/tools/kibana-exporter/exports/apache.yml"
```

Publishing as a draft as needs cleanup.


dev/tools/kibana-exporter/kibana.go
```diff
@@ -0,0 +1,221 @@
...
+	configType        = "config"
+	timelionSheetType = "timelion-sheet"
+
+	// Queries
+	// /api/saved_objects/_find?type=dashboard&search=id:AV4REOpp5NkDleZmzKkE-ecs
+)
+
+var (
+	// TODO: Hardcoded at the moment, should be a param
+	Host = "elastic:changeme@localhost:5601"
```
ruflin: Needs to be configurable

go.sum
```diff
@@ -1,23 +1,114 @@
+github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
```
ruflin: Dependencies should not be added on this level

mtojek:
@ruflin  I think this one can be closed in favor of https://github.com/elastic/package-registry/pull/247

ruflin:
I think later on we will need some tooling around exporting dashboards which were built in Kibana but no need to tackle this now.

PR #216, opened 2020-02-18
Add code to expose datasource configs in API

ruflin:
In https://github.com/elastic/package-registry/pull/212 the datasource structure was introduced in the manifests but not exposed through the APIs yet. This PR changes this exposing all the fields. So far it only exposes the fields and no validation is done yet. This needs to be added in the future.

A test package was added to see the output of datasource configs and potential changes to it.


ruflin:
@jfsiii I assume shipping this will not break the Kibana code but we miss the info. Could you take care of updating the types on the Kibana end?

PR #217, opened 2020-02-18
Complete System package with default metricsets

ruflin:
The system package is one of the most complex packages. This PR completes the system package with all the metricsets and filesets which are enabled in our system modules.

Further changes:

* Remove old generated system package
* Improve error handling of building packages


ruflin:
I directly merged this as it only extends an existing package.

PR #218, opened 2020-02-18
Update logs package to support datasources

ruflin:
My initial plan was to remove the logs package as the default setup for logs-*-* is already done by alias templates. But I realised there are still two cases where we can use the logs package:

* Update the logs-*-* template to a never version
* Use the logs datasource as our generic data source template

This means, this datasource allows us to support "custom datasources" with a package without having to spend additional efforts on it. All the logs will be shipped for now to logs-generic-{namespace} but at least users can select their own paths.


ruflin:
@jen-huang @hbharding Curious to hear what you think about the above.

PR #223, opened 2020-02-21
Elastic Endpoint Package

jonathan-buttner:
This PR adds the elastic endpoint package that has datasets for events and metadata and the fields.yml for both.

I created the `security` category.

If we want a index pattern created for `events-*` we'll  need to merge this PR as well: https://github.com/elastic/kibana/pull/58908


ruflin:
For the categories: I guess we will only have security or endpoint? I would suggest to go for now with "Security". Are the events you are shipping now ending up under `endpoint-*` or `event-*`?

jonathan-buttner:
> For the categories: I guess we will only have security or endpoint? I would suggest to go for now with "Security".

Sounds good! I'll switch it to Security

> Are the events you are shipping now ending up under `endpoint-*` or `event-*`?

Based on @james-elastic last comment here: https://github.com/elastic/endpoint-app-team/issues/102#issuecomment-588290461 I think we're going to go with `event-*` so I'll make that change as well 👍

dev/package-examples/endpoint-0.0.1/dataset/endpoint.events/fields/fields.yml
```diff
@@ -0,0 +1,3807 @@
...
+      ignore_above: 1024
+      description: File type (file, dir, or symlink).
+      example: file
+    - name: uid
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: The user ID (UID) or security identifier (SID) of the file owner.
+      example: "1001"
+- name: file_classification
```
jonathan-buttner: I haven't pulled in @marshallmain latest changes to the alert schema. Once he merges the PR, I'll pull in the latest and regenerate the fields.yml and either put up a new PR or update this one.

util/package.go
```diff
@@ -21,8 +21,9 @@ import (
 const defaultType = "integration"
 
 var CategoryTitles = map[string]string{
-	"logs":    "Logs",
-	"metrics": "Metrics",
+	"logs":     "Logs",
+	"metrics":  "Metrics",
+	"security": "Security",
```
jonathan-buttner: @ruflin Should I also add `events` here?

dev/package-examples/endpoint-0.0.1/dataset/endpoint.events/manifest.yml
```diff
@@ -0,0 +1,6 @@
+title: Endpoint Events
```
ruflin: Not for this file, but for the path. I think this should be just `events` and not `endpoint.events`. I assume the problem you tried to solve with this is solved in https://github.com/elastic/package-registry/pull/176

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
```diff
@@ -0,0 +1,200 @@
+- name: "@timestamp"
```
ruflin: Same comment as above for the path.

dev/package-examples/endpoint-0.0.1/manifest.yml
```diff
@@ -0,0 +1,18 @@
...
+name: endpoint
+title: Elastic Endpoint
+description: This is the Elastic Endpoint package.
+version: 0.0.1
+categories: ["security"]
+# Options are experimental, beta, ga
+release: beta
+# The package type. The options for now are [integration, solution], more type might be added in the future.
+# The default type is integration and will be set if empty.
+type: integration
```
ruflin: I would assume the type here would be `solution`?

util/package.go
ruflin: No, this is really only about the categories shown in the UI.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
ruflin: I remember also a discussion around endpoint only using one index? But having 2 datasets would mean 2 indices?

util/package.go
jonathan-buttner: ok sounds good 👍

dev/package-examples/endpoint-0.0.1/dataset/endpoint.events/manifest.yml
jonathan-buttner: Yeah I was going off the code that's currently in `feature-ingest` https://github.com/elastic/kibana/blob/feature-ingest/x-pack/plugins/ingest_manager/server/services/epm/elasticsearch/index.ts#L13

So if I understand issue https://github.com/elastic/package-registry/pull/176 correctly there's going to be a field in the manifest.yml of a dataset and I can set the `path` there? and then the code I linked above will look for `dataset.path` instead of doing `${dataset.type}-${dataset.name}` ?

I think the `path` would be `endpoint.events`?

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
jonathan-buttner: Gotcha I'll change the dataset name.

So there will be one index for events and alerts (events) and then another index for the metadata of an endpoint for now.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.events/manifest.yml
ruflin: If you call the directory `endpoint` you should not have to do anything as by default it assume path and id are identical. In this case the `dataset.id == endpoint`. The type of the data is defined as `events` so you should get directly `events-endpoint-*`.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
ruflin: Got it, then this makes sense. 

This raises the question how the metadata index will be called. `events-endpoint.metadata-*`? If yes, the name here is correct.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.events/manifest.yml
jonathan-buttner: ah ok, yeah I think the reason I chose `endpoint.events` is because I was worried that someone looking at the directories wouldn't immediately know what `dataset/endpoint` but I also didn't want to do `dataset/events` because the name would be `events-events`. I'll change it to `endpoint` and if people are confused we can change it later.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
jonathan-buttner: So this one will actually be `metrics-endpoint.metadata-*` (the type is `metrics`, this is information about endpoints and their status etc) Do you think `endpoint.metadata` is reasonable name? I figure `metrics-metadata` wasn't super clear on what it was referring to, that's why I add the `endpoint` in there.

If we still with `endpoint.metadata` would it be better to be consistent in the dataset directory and leave the events one as `endpoint.events`?

ruflin:
I created https://github.com/elastic/package-registry/pull/228 to also have a base template in place for events-*. Should not affect this PR but thought worth to mention here.

dev/package-examples/endpoint-0.0.1/dataset/endpoint.metadata/fields/fields.yml
ruflin: Agree that we can't really use `metrics-metadata` as this is too generic and not clear it would belong to endpoint. 

By default we will have that the dataset id = {package.name}.{directory.name}. So in the nginx case we have access, error and stubstatus directories. The problem above is that you only have `events-endpoint` which means the above does not work as the package is already endpoint. You will have to set id in any case. I would probably call the directory è
`events`, and the other one `metadata`.

jonathan-buttner: Ah ok I see. I'll rename the directories and set the `id` field. Should I rebase off of https://github.com/elastic/package-registry/pull/176 ?

ruflin: I suggest lets get this in as is and then adjust it again as soon as #176 lands.

jonathan-buttner: Ok sounds good, I moved the datasets to `events` and `metadata`. I'm good to merge. @james-elastic could I get a 👍

PR #224, opened 2020-02-24
Add basic logs and metrics template

ruflin:
For logs and metrics indices, we will provide default templates. This makes sure, that all indices which are created in logs-*-* and metrics-*-* follow the convention.

In this PR, a first example of these templates is provided. In total there are 4 templates, one for mapping and one for settings for each. The content of the templates itself is very basic at the moment. It contains a few fields from ECS and defaulting to keyword for all fields otherwise if it is a string. More work must be put into these templates in the future.

These templates are a bit different from the other fields.yml as they are installed globally and don't depend on a data source. Because of this they are in the top level Elasticsearch directory and directly in the json format. From an installation perspective in the package manager my expectations are as following:

* EPM takes the name of the file and the content and puts it directly into Elasticsearch (no processing)

### DIFF
--- a/dev/package-examples/base-1.0.0/elasticsearch/index-template/logs-mapping.json
+++ b/dev/package-examples/base-1.0.0/elasticsearch/index-template/logs-mapping.json
@@ -1,0 +1,134 @@
+{
+    "order": 1,
+    "index_patterns": [
+        "logs-*-*"
+    ],
+    "mappings": {
+        "_meta": {
+        },
+        "dynamic_templates": [
+            {
+                "strings_as_keyword": {
+                    "mapping": {
+                        "ignore_above": 1024,
+                        "type": "keyword"
+                    },
+                    "match_mapping_type": "string"
+                }
+            }
+        ],
+        "date_detection": false,
+        "properties": {
+            "agent": {
+                "properties": {
+                    "hostname": {
+                        "ignore_above": 1024,
+                        "type": "keyword"
+                    },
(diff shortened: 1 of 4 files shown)


ruflin:
@neptunian @jfsiii This should make it possible to get some basics in place based on a package. Long term, it should end up in ES directly. I assume the above will need some additional logic in EPM to deal with it.

ruflin:
@jonathan-buttner This is something we could also use as a "trick" to install a global template for endpoint for now during development. So the endpoint package contains the fields.yml for the datasources but a global template with settings as an example.

jonathan-buttner:
> @jonathan-buttner This is something we could also use as a "trick" to install a global template for endpoint for now during development. So the endpoint package contains the fields.yml for the datasources but a global template with settings as an example.

Ah ok cool. How do you think we'll create the complete global templates json files in the future? We could probably define a subset file that includes certain fields from ecs core and create some custom ecs files: https://github.com/elastic/ecs/pull/737 that way if we need to add additional fields to the global template we can do it in a repeatable way.

ruflin:
I think the mapping part of the global template we should also start to build based on fields.yml in the future. This here is just a shortcut on my end for now.

What fields the global mapping will contain is a very interesting conversation we must have. It should be as little as possible and as much as needed. I think we end up with a subset of ECS fields and the fields might different for metrics-*, logs-* and events-*. Checking your above ECS link, seems like we are on the same page.

PR #227, opened 2020-03-02
Rename dataset inputs to streams

ruflin:
The manifest on the dataset level defines the streams which use an input. Because of this naming it streams instead of input makes probably more sense.


ruflin:
@jen-huang Opened this based on your feedback.

ruflin:
The overall plan is to get the streams also mapped under the inputs in the data source. But lets get this PR in first and follow up with a PR to duplicate the data there.

PR #228, opened 2020-03-03
Add events template to base package

ruflin:
Endpoint will use the events-* prefix. To make sure things work out of the box also with events-*-* this adds mappings and settings for the events prefix to the base package.

### DIFF
--- a/dev/package-examples/base-1.0.0/elasticsearch/ilm-policy/events-default.json
+++ b/dev/package-examples/base-1.0.0/elasticsearch/ilm-policy/events-default.json
@@ -1,0 +1,15 @@
+{
+    "policy": {
+        "phases": {
+            "hot": {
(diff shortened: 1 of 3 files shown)


neptunian:
This will work with the PR open that handles the elasticsearch assets for ilm and index templates (https://github.com/elastic/kibana/pull/58869).  Though, I don't see how this structure is useful for index templates v2.  It seems like we should have a "base" template which can refer to template components like settings and mappings. these would follow in the v2 template component style:

```
{
    "template": {
       "settings": {...},
    },
    "version": 0-∞
}
```

And we'd keep `order`, `index-patterns`, and `aliases` in the "base", non component template

ruflin:
As soon as the new templates land, I will update the base package.

PR #229, opened 2020-03-03
change mapping to mappings for index template names

neptunian:
this doesn't effect anything, but changed so that mappings is also plural as per the documentation, so the name is as expected.


ruflin:
jenkins, test this

neptunian:
retest

PR #230, opened 2020-03-03
Add streams as part of Datasource / Inputs

ruflin:
The streams are defined in the dataset and reference an input. The input is defined in the datasource. For the UI to not have to combine to two together, the API already provides the datasource with all inputs and streams listed inside.

This change also extended the test packages to have better examples for this change.

Note: This PR gets the job done. Code could be cleaned up later.

### DIFF
--- a/testdata/public/package/datasources-1.0.0/index.json
+++ b/testdata/public/package/datasources-1.0.0/index.json
@@ -12,20 +12,57 @@
   },
   "assets": [
     "/package/datasources-1.0.0/manifest.yml",
-    "/package/datasources-1.0.0/dataset/example/manifest.yml"
+    "/package/datasources-1.0.0/dataset/examplelog1/manifest.yml",
+    "/package/datasources-1.0.0/dataset/examplelog2/manifest.yml",
+    "/package/datasources-1.0.0/dataset/examplemetric/manifest.yml"
   ],
   "format_version": "1.0.0",
   "datasets": [
     {
       "title": "Example dataset with inputs",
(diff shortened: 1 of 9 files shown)


hbharding:
Thanks these are helpful examples to look at!

jen-huang:
@jfsiii Would be great if you can look into the changes needed on EPM side of things. I'm already actively working on agent config/datasource schemas & related APIs.

testdata/public/package/datasources-1.0.0/index.json
```diff
@@ -87,11 +147,84 @@
               "type": "password"
             }
           ],
-          "description": "Collecting metrics for nginx."
+          "description": "Collecting metrics for nginx.",
```
jen-huang: I approved too fast.. 🙂 During our call earlier we also discussed the need for `dataset` property, i.e. this line in agent config spec: https://github.com/elastic/beats/blob/feature-ingest/x-pack/agent/docs/agent_configuration_example.yml#L50
I can't find it in this PR?

ruflin: I added the dataset in the most recent commit. Unfortunately it is not the full name yet as it depends on https://github.com/elastic/package-registry/pull/176 

I'm still debating on the name here but I assume if we change it from foo to bar, the change on your end should be pretty straightforward?

jen-huang: yeah, doesn't matter what the property name is, as long as it's there

Issue #221, opened 2020-02-20
Generate integrations from Beats modules


exekias:
As Beats module development and packages registry will coexist for a while, it is in our interest to avoid duplicating the development process, to avoid human errors and inconsistencies. The aim of this issue is to automate as much as possible the generation of registry packages based on the Beats repo.

To scope this down we can initially focus on:

- ~Create an initial script that fills the directories an basic manifest.yml, reading from both Metricbeat and Filebeat modules~ (done)
- Keep iterating over the script to add support for importing as many aspects as possible from the original modules, including:
  - [x] fields.yml
  - [x] Dashboard screenshots
  - [x] Dashboards and visualizations
  - [x] Icons
  - [x] Docs
  - [x] Input configs (manifest.yml)
  - [x] Streams (stream.yml)

We can initially focus on a small list of packages, and grow it later. In some cases we may need to update the Beats repo to include some metadata that is only useful for packages registry. For example, there are no icons in the beats repo, but we can add them for the shake of this process.

We may also find that automating the import process becomes too difficult for some pieces, we can leave some steps as manual, trying to keep these to a minimum.

There is an, now outdated, initial try at this: https://github.com/elastic/package-registry/pull/137

exekias:
cc @urso @ruflin feel free to comment / update

mtojek:
Hi, I've started researching this area, browsing docs, etc., in general - preparing for development here.

Is there any model package that can be considered as template/archetype/skeleton?
Is the "nginx" package the best candidate? https://github.com/elastic/package-registry/tree/master/dev/package-examples/nginx-1.2.0

mtojek:
Hi, I've started researching this area, browsing docs, etc., in general - preparing for development here.

Is there any model package that can be considered as template/archetype/skeleton?
Is the "nginx" package the best candidate? https://github.com/elastic/package-registry/tree/master/dev/package-examples/nginx-1.2.0

mtojek:
Hi, I've started researching this area, browsing docs, etc., in general - preparing for development here.

Is there any model package that can be considered as template/archetype/skeleton?
Is the "nginx" package the best candidate? https://github.com/elastic/package-registry/tree/master/dev/package-examples/nginx-1.2.0

mtojek:
Hi, I've started researching this area, browsing docs, etc., in general - preparing for development here.

Is there any model package that can be considered as template/archetype/skeleton?
[...]

PR #231, opened 2020-03-04
Add basic "import-beats" scripts

mtojek:
This PR creates an initial script that fills the directories an basic manifest.yml.

How to run:
```bash
$ go run dev/import-beats/*.go -beatsDir=../beats -outputDir=dev/package-beats
```

Sample generated manifest.yml:
```
format_version: 1.0.0
name: ibmmq
version: 0.0.1
description: ""
type: integration
categories:
- logs
- metrics
requirement:
  kibana:
    versions: ""
download: ""
path: ""
```

Missing data will be filled in during next iterations (need to open fields.yml).

### DIFF
--- a/dev/import-beats/packages.go
+++ b/dev/import-beats/packages.go
@@ -1,0 +1,73 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package main
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+
+	"github.com/elastic/package-registry/util"
+)
+
+type packageMap map[string]*util.Package
+
+func (p packageMap) loadFromSource(beatsDir, beatName, packageType string) error {
+	path := filepath.Join(beatsDir, beatName, "module")
+	moduleDirs, err := ioutil.ReadDir(path)
+	if err != nil {
+		return errors.Wrapf(err, "cannot read directory '%s'", path)
+	}
+
+	for _, moduleDir := range moduleDirs {
+		if !moduleDir.IsDir() {
+			continue
+		}
+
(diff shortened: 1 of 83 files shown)


dev/package-beats/aerospike-0.0.1/manifest.yml
```diff
@@ -0,0 +1,12 @@
...
+name: aerospike
+version: 0.0.1
+description: ""
+type: integration
+categories:
+- metrics
+requirement:
+  kibana:
+    versions: ""
+download: ""
```
ruflin: We should find a way that download and path don't show up here. They are not needed in the manifest.

util/package.go
```diff
@@ -26,22 +26,23 @@ var CategoryTitles = map[string]string{
 }
 
 type Package struct {
+	FormatVersion string `config:"format_version" json:"format_version" yaml:"format_version"`
+
 	Name          string  `config:"name" json:"name"`
-	Title         *string `config:"title,omitempty" json:"title,omitempty"`
+	Title         *string `config:"title,omitempty" json:"title,omitempty" yaml:"title,omitempty"`
```
ruflin: Your change made me realise we don't have a license field in here but it is something we have in the manifest. Will open a PR.

dev/import-beats/packages.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	for _, moduleDir := range moduleDirs {
+		if !moduleDir.IsDir() {
+			continue
+		}
+
+		log.Printf("Visit '%s:%s'\n", beatName, moduleDir.Name())
+
+		_, ok := p[moduleDir.Name()]
```
ruflin: Perhaps worth a comment that this is needed because in packages logs and metrics modules are combined.

dev/package-beats/aerospike-0.0.1/manifest.yml
mtojek: At the moment I can mark them with `omitempty` to simply skip them if empty. If this is insufficient I can clean their content before writing to file.

dev/import-beats/packages.go
mtojek: Comment left.

PR #235, opened 2020-03-05
Move package directories to its own subdirectory

ruflin:
As we had by now 3 package directories, I moved them under their own directory in packages


ruflin:
@mtojek This means you need to adjust your command.

mtojek:
> @mtojek This means you need to adjust your command.

At the moment, it's executed like this:

```bash
$ go run dev/import-beats/*.go -beatsDir=../beats -outputDir=dev/package-beats
```
(no need to apply special adjustments, but we can think about putting this under a dedicated mage step)

ruflin:
I fixed the test but need to clean up this one. I think its a bad test as it depends on an actual package instead of the test packages. Will follow up on this separately.

PR #241, opened 2020-03-10
Update the system package

ruflin:
Add all existing metricsets and filesets to the system package with all config options. This should be a good example case on how all the configs can be shown in the UI. Also cleaning up some incorrect / missing configs in the datasets.


ruflin:
Merging directly as it is only a change to the packages and not the service itself.

jfsiii:
I'm not sure if there's any impact, but I believe this is a new version of `system`, right? I don't know if our caching of the prior 0.9.0 will cause any issues getting this 0.9.0 but it would when we're live.

ruflin:
@jfsiii As soon as we are live, keeping the version will not be acceptable. For now it will be in the cache for max 24h if you use a url with the CDN.

PR #242, opened 2020-03-10
Cleanup nginx package and add additional fields

ruflin:
Additional fields are added to complete the UI. See screenshot below for some more details / description.

Screenshot 2020-03-10 at 09 40 37

Syslog input is removed as it was only used as an example.

### DIFF
--- a/dev/packages/example/nginx-1.2.0/dataset/stubstatus/manifest.yml
+++ b/dev/packages/example/nginx-1.2.0/dataset/stubstatus/manifest.yml
@@ -12,7 +12,10 @@ release: beta
 
 streams:
   - input: "nginx/metrics"
-
-    # Only the variables have to be repeated that are not specified as part of the input
+    title: Nginx stubstatus metrics
+    description: Collecting [stubstatus metrics](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) from the nginx service.
     vars:
-      # All variables are specified in the input already
+      - name: period
+        description: "Collection period. Valid values: 10s, 5m, 2h"
+        default: "10s"
+        type: duration
--- a/dev/packages/example/nginx-1.2.0/manifest.yml
+++ b/dev/packages/example/nginx-1.2.0/manifest.yml
@@ -27,37 +27,34 @@ datasources:
         # This is for selection in the stream
         # id: nginx
         type: nginx/metrics
-        description: Collecting metrics for nginx.
+        # Title should never have an ending dot.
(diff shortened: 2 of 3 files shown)


ruflin:
@jen-huang Let me know what you think about the above.

dev/packages/example/nginx-1.2.0/dataset/stubstatus/manifest.yml
```diff
@@ -12,7 +12,10 @@ release: beta
 
 streams:
   - input: "nginx/metrics"
-
-    # Only the variables have to be repeated that are not specified as part of the input
+    title: Nginx stubstatus metrics
+    description: Collecting [stubstatus metrics](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) from the nginx service.
```
jen-huang: Is allowing markdown on all `description` fields a business rule we should implement to allow the UI to parse it correctly? That would be nice for even var descriptions too, for linking to external docs

dev/packages/example/nginx-1.2.0/manifest.yml
```diff
@@ -27,37 +27,34 @@ datasources:
         # This is for selection in the stream
         # id: nginx
         type: nginx/metrics
-        description: Collecting metrics for nginx.
+        # Title should never have an ending dot.
+        title: Collecting metrics from nginx hosts
+        description: Collectiong stubstatus metrics from the nginx hosts.
```
jen-huang:
```suggestion
        title: Collect metrics from nginx hosts
        description: Collect stubstatus metrics from nginx hosts.
```

copy correction to match with title and description copy of `logs` input (L58-59)

jen-huang:
Typing and UI adjustments done here, including parsing `description` fields as markdown: https://github.com/elastic/kibana/pull/59822/commits/f34234291ed862e52eb72144368aa393b2c90afd

dev/packages/example/nginx-1.2.0/dataset/stubstatus/manifest.yml
ruflin: TBH this markdown here was coming from some copy / paste but I really like the idea of allowing markdown in the description for all description fields. Looking at your implementation it seems it is not very complex to allow this.

ruflin:
@jen-huang I added your suggestion and will merge this PR. Will follow up with a PR on the registry side to document these fields on how they should be used and also the markdown part.

PR #243, opened 2020-03-10
RFC: Have datasource type to link to correct solution

ruflin:
Some of the datasources will not be configured by the ingest manager UI. An example for this are Endpoint or Uptime. The UI still shows the datasource entry and must know which solution to link to when someone selects such a datasource. For this, a field `solution: endpoint` was introduced inside a datasource.

### DIFF
--- a/dev/packages/example/endpoint-0.0.1/manifest.yml
+++ b/dev/packages/example/endpoint-0.0.1/manifest.yml
@@ -11,6 +11,19 @@ release: beta
 type: solution
 license: basic
 
+
+
+datasources:
+  -
+    name: endpoint
+    title: Endpoint data source
+    description: Interact with the endpoint.
+
+    # This tells the UI that for configuration, it must link to a specific solution
+    # Only solution packages can contain this field.
+    solution: endpoint
+
+
 requirement:
   elasticsearch:
     versions: ">7.4.0"
--- a/util/package.go
+++ b/util/package.go
@@ -53,7 +53,8 @@ type Datasource struct {
 	Name        string  `config:"name" json:"name" validate:"required"`
 	Title       string  `config:"title" json:"title" validate:"required"`
 	Description string  `config:"description" json:"description" validate:"required"`
-	Inputs      []Input `config:"inputs" json:"inputs" validate:"required"`
+	Solution    string  `config:"solution" json:"solution" `
+	Inputs      []Input `config:"inputs" json:"inputs"`
 }
 
 type Requirement struct {


dev/packages/example/endpoint-0.0.1/manifest.yml
```diff
@@ -11,6 +11,19 @@ release: beta
...
+
+datasources:
+  -
+    name: endpoint
+    title: Endpoint data source
+    description: Interact with the endpoint.
+
+    # This tells the UI that for configuration, it must link to a specific solution
+    # Only solution packages can contain this field.
+    solution: endpoint
```
ruflin: @jen-huang Will this help to link to endpoint when someone clicks and endpoint datasource? Or selects endpoint as a datasource?

dev/packages/example/endpoint-0.0.1/manifest.yml
```diff
@@ -11,6 +11,19 @@ release: beta
 type: solution
 license: basic
 
+
```
ruflin: @james-elastic I start to challenge if the package itself is a solution or an integrations package and the difference is only below in the datasource.

dev/packages/example/endpoint-0.0.1/manifest.yml
jen-huang: ++ I like this field. We can treat the value as the app in Kibana to link to (`/app/{solution}`), or hardcode a solution<->app url mapping in the code. I'm not sure if this is enough for what to display in Step 2 of the data source wizard though. It will currently display "Nothing to configure" and if the user confirms the addition, it will add a datasource with an empty `inputs` array

ph: I like the name for that, good idea to add a dedicated field for it.

ph: it could be also a function call or a handler that a solution register to us, we can then pass the content of the datasource for disambiguation, this is where packages or other fields could be used to display the appropriate UI.

ruflin: @jen-huang I think the user should never end up on page 2 of the wizard but be redirected to endpoint for step 2.

ruflin:
Merged for now, we can keep iterating on it. But having it in will allows us to play around with it in the UI. 

I think it would be nice if endpoint would offer us an endpoint where we can send the user to with the config and he will directly end up redirected to the correct screen.

PR #245, opened 2020-03-11
Make src in image required

ruflin:


util/package.go
```diff
@@ -71,7 +71,7 @@ type Version struct {
 }
 
 type Image struct {
-	Src   string `config:"src" json:"src,omitempty"`
+	Src   string `config:"src" json:"src,omitempty" validate:"required"`
```
jfsiii: Since it's required, is `omitempty` needed any longer?
```suggestion
	Src   string `config:"src" json:"src" validate:"required"`
```

ruflin: Agree.

PR #246, opened 2020-03-11
Add icons to aws, log and nginx package

ruflin:


dev/packages/example/log-0.9.0/img/icon.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ruflin: @hbharding I'm using this as the logs icon for now (one from EUI) to have one, but it would be nice to have an actual log icon for this package.

PR #247, opened 2020-03-11
Import "beats" dashboards 

mtojek:
This PR extends "import-beats" script to pull dashboards and visualizations.

_I think we can iterate on top (index names, etc.) of this to come a final version of Kibana artifacts._


dev/import-beats/common.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// put associates the specified value with the specified key. If the map
+// previously contained a mapping for the key, the old value is replaced and
+// returned. The key can be expressed in dot-notation (e.g. x.y) to put a value
+// into a nested map.
+//
+// If you need insert keys containing dots then you must use bracket notation
+// to insert values (e.g. m[key] = value).
+func (m mapStr) put(key string, value interface{}) (interface{}, error) {
```
ruflin: @urso We should really put mapStr in its own packages. I also had other uses in the past for it.

dev/import-beats/kibana.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func (km *kibanaMigrator) migrateDashboardFile(dashboardFile []byte) ([]byte, error) {
+	dashboardFile, err := prepareDashboardFile(dashboardFile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "preparing file failed")
+	}
+
+	request, err := http.NewRequest("POST",
+		fmt.Sprintf("http://%s/api/kibana/dashboards/import?force=true", km.hostPort),
```
ruflin: What is force doing here? Just curious

dev/import-beats/kibana.go
```diff
@@ -0,0 +1,220 @@
...
+	return ms, nil
+}
+
+func createKibanaContent(kibanaMigrator *kibanaMigrator, modulePath string) (kibanaContent, error) {
+	moduleDashboardPath := path.Join(modulePath, "_meta", "kibana", "7", "dashboard")
+	moduleDashboards, err := ioutil.ReadDir(moduleDashboardPath)
+	if os.IsNotExist(err) {
+		log.Printf("\tno dashboards present, skipped (modulePath: %s)", modulePath)
+		return kibanaContent{}, nil
+	} else if err != nil {
```
ruflin: Nit: Problably no else needed as you return in the if clause above.

dev/import-beats/kibana.go
```diff
@@ -0,0 +1,220 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "marshalling object failed")
+		}
+
+		id, err := object.getValue("id")
+		if err != nil {
+			return nil, errors.Wrapf(err, "retrieving id failed")
+		}
+
+		extracted[id.(string)+".json"] = data
```
ruflin: The id of the saved object becomes the name? I like that.

dev/packages/beats/activemq-0.0.1/kibana/dashboard/26434790-1464-11ea-8fd8-030a13064883.json
```diff
@@ -0,0 +1,33 @@
+{
+    "attributes": {
+        "description": "This dashboard shows application logs collected by the ActiveMQ filebeat module.",
+        "hits": 0,
+        "kibanaSavedObjectMeta": {
+            "searchSourceJSON": "{\"filter\":[],\"query\":{\"language\":\"kuery\",\"query\":\"\"}}"
+        },
+        "optionsJSON": "{\"hidePanelTitles\":false,\"useMargins\":true}",
```
ruflin: The JSON strings should be decoded inside the package. I think you have all the logic for it already in place for the encoding, otherwise see https://github.com/elastic/package-registry/pull/214

dev/import-beats/kibana.go
mtojek: Query parameters

```
force
(Optional, boolean) Overwrite any existing objects on ID conflict.
```

mtojek:
> LGTM. Comments can be in follow up PRs.
> 
> If I understand it correctly, you load the dashboards into KB and then pull them out again?

I think there is no need to pull them out. Once you POST them to Kibana, in the response you receive migrated/converted objects.

dev/import-beats/kibana.go
mtojek: That is true :)

dev/packages/beats/activemq-0.0.1/kibana/dashboard/26434790-1464-11ea-8fd8-030a13064883.json
mtojek: Hmm... if you want to upload them to Kibana, you need to encode them again. I was wondering if it's not better to keep them encoded in packages as well. WDYT?

dev/import-beats/kibana.go

ruflin:
That you get it in the response was the piece I was missing. This simplifies things!

dev/packages/beats/activemq-0.0.1/kibana/dashboard/26434790-1464-11ea-8fd8-030a13064883.json
ruflin: The reason we decode them is for versioning purpose. If you change anything small int he visualisation and then get a PR with the change, it is almost impossible to figure out what change. That is the reason we switched to the decoded format in Beats some while ago and I would like also to use it here. My hope is that in the future Kibana directly supports the decoded option.

The encoding currently happens in the build step.

mtojek: Ah, you're right! Versioning is a killer argument in this thread.

I've just pushed a change that decodes fields.

PR #248, opened 2020-03-11
Update reference package

ruflin:
Add details to the reference package discussed in https://github.com/elastic/package-registry/pull/242


ruflin:
@mtojek Some comments in this PR might help you to get some more context on what fields are used for in the manifests.

PR #251, opened 2020-03-12
Support all Kibana object types

mtojek:
This PR adjusts the "import-beats" script to support all object types.


testdata/public/package/example-1.0.0/index.json
```diff
@@ -79,6 +79,7 @@
       "name": "logs",
       "title": "Logs datasource",
       "description": "Datasource for your log files.",
+      "solution": "",
```
ruflin: I need to fix this on my end, should be skipped if empty.

mtojek: Not needed, I fixed it here.

PR #252, opened 2020-03-12
Cleanup log package and add system package dashboards

ruflin:
* Use generic dataset in log package
* Remove empty ingest pipeline
* Add dashboards to system package


ruflin:
@mtojek Thanks for the dashboards ;-)

PR #254, opened 2020-03-12
Make dataset configurable for log dataset

ruflin:
This would allow a user to configure multiple log datasources and send the data to different indices. One of the challenges with this is that we need to verify that dataset only contains valid chars. Other question is if we should allow it or rather have users building packages.

### DIFF
--- a/dev/packages/example/log-0.9.0/dataset/log/manifest.yml
+++ b/dev/packages/example/log-0.9.0/dataset/log/manifest.yml
@@ -24,4 +24,3 @@ streams:
           Set the name for your dataset. Changing the dataset will send the data to a different index.
           You can't use `-` in the name of a dataset and only valid characters for [Elasticsearch index names](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html).
         type: text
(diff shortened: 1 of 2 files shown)


ruflin:
@jen-huang This currently crashes Kibana as soon as I want to look at advanced options.

dev/packages/example/log-0.9.0/dataset/log/manifest.yml
```diff
@@ -16,3 +16,12 @@ streams:
         type: text
         multi: true
 
+      - name: dataset
+        required: true
+        default: generic
```
jen-huang: I think it's crashing because it's set as `multi: true` but the default value doesn't seem to be an array

ruflin: multi is not needed here. I removed it.

ruflin: +1 on having validation for it.

ruflin:
@ph I merged the above to test it further. Curious if it works as expected with the agent.

PR #255, opened 2020-03-12
Add icons to packages

mtojek:
This PR extends "import-beats" script to collect package icons.

It needs to determine the size of an icon (SVG type).


dev/packages/beats/activemq-0.0.1/img/logo_activemq.svg
```diff
@@ -0,0 +1,31 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 128 128">
```
ruflin: I wonder if it would simplify the code if you would always use logo.svg and just add this to the manifest.

PR #256, opened 2020-03-12
Fix: rename indices

mtojek:
This PR renames indices: metricbeat-*, filebeat-* to: metric-*, log-*.


dev/import-beats/kibana.go
```diff
@@ -75,9 +75,14 @@ func (km *kibanaMigrator) migrateDashboardFile(dashboardFile []byte) ([]byte, er
 	return saved, nil
 }
 
+
 func prepareDashboardFile(dashboardFile []byte) ([]byte, error) {
 	var documents kibanaDocuments
 
+	// Rename indices (metricbeat, filebeat)
+	dashboardFile = bytes.ReplaceAll(dashboardFile, []byte(`metricbeat-*`), []byte(`metric-*`))
```
ruflin: This should be logs-* and metrics-*

PR #257, opened 2020-03-12
Reduce cache time to 10min

ruflin:
We are currently still in development and not always update the version of the packages. Because of this we use a much shorter cache time at the moment.


ruflin:
@nchaulet This should reduce the problem in the future.

PR #260, opened 2020-03-12
[Endpoint] Adding image and fixing inputs

jonathan-buttner:
This PR adds an image and some inputs so the kibana UI doesn't fail when installing the package.


jonathan-buttner:
Thanks for the review, yeah I agree, I had to do this because the UI was throwing an error here: https://github.com/elastic/kibana/blob/master/x-pack/plugins/ingest_manager/public/applications/ingest_manager/sections/agent_config/create_datasource_page/step_configure_datasource.tsx#L188

@jfsiii @jen-huang I had to add this to the manifest:

```
    inputs:
      - type: metrics
      - type: events
```
https://github.com/elastic/package-registry/pull/260/files#diff-381ba165569ba08ab0eb43fc21b4cb66R22

otherwise the I think the UI was failing to find the `packageInfo.datasources[0].inputs`. Would it make sense to add inputs to the check as well?
Something like:
`packageInfo.datasources && packageInfo.datasources[0] && packageInfo.datasources[0].input`

jen-huang:
@jonathan-buttner I'll add some additional safeguards here. Thanks!

jen-huang:
PR that adds more null checks: https://github.com/elastic/kibana/pull/60426

jonathan-buttner:
Awesome thanks @jen-huang I'm testing it right now and then I'll 👍

PR #266, opened 2020-03-16
Add show_user flag to AWS and nginx

ruflin:
The show_user flag indicates config options which should be shown to the user by default in the UI. For now the changes has been applied on the input level for AWS and nginx access logs on the stream level.


jen-huang:
Added support on UI side in https://github.com/elastic/kibana/pull/60338

PR #270, opened 2020-03-17
import-beats: copy ingest pipelines

mtojek:
This PR adds support for ingest pipelines to "import-beats" script.


dev/packages/beats/activemq-0.0.1/dataset/audit/elasticsearch/ingest-pipeline/pipeline.yml
```diff
@@ -0,0 +1,32 @@
+---
```
ruflin: if the pipeline is called `default.yml` or `default.json` we automatically fall back to it without any additional config in the manifest. Perhaps worth renaming it.

mtojek: Let's consider these two cases:

1. Logstash

```
../beats/filebeat/module/logstash/slowlog/ingest/pipeline-json.yml
../beats/filebeat/module/logstash/slowlog/ingest/pipeline-plain.yml
```
Which one should called `default.json`? Should I rename it only if there is a single pipeline?

2. Elasticsearch

```
../beats/filebeat/module/elasticsearch/slowlog/ingest/pipeline-json.yml
../beats/filebeat/module/elasticsearch/slowlog/ingest/pipeline-plaintext.yml
../beats/filebeat/module/elasticsearch/slowlog/ingest/pipeline.yml
```

Which one should be called `default.yml`?

mtojek:
> How will you deal with the case where there is more then 1 ingest pipeline?

In Elasticseach filebeat module, `slowlog` manifest.yml:

```
module_version: 1.0

var:
  - name: paths
    default:
      - /var/log/elasticsearch/*_index_search_slowlog.log
      - /var/log/elasticsearch/*_index_indexing_slowlog.log
      - /var/log/elasticsearch/*_index_search_slowlog.json
      - /var/log/elasticsearch/*_index_indexing_slowlog.json
    os.darwin:
      - /usr/local/var/lib/elasticsearch/*_index_search_slowlog.log
      - /usr/local/var/lib/elasticsearch/*_index_indexing_slowlog.log
      - /usr/local/var/lib/elasticsearch/*_index_search_slowlog.json
      - /usr/local/var/lib/elasticsearch/*_index_indexing_slowlog.json
    os.windows:
      - c:/ProgramData/Elastic/Elasticsearch/logs/*_index_search_slowlog.log
      - c:/ProgramData/Elastic/Elasticsearch/logs/*_index_indexing_slowlog.log
      - c:/ProgramData/Elastic/Elasticsearch/logs/*_index_search_slowlog.json
      - c:/ProgramData/Elastic/Elasticsearch/logs/*_index_indexing_slowlog.json

ingest_pipeline:
  - ingest/pipeline.yml
  - ingest/pipeline-plaintext.yml
  - ingest/pipeline-json.yml
input: config/slowlog.yml
```

Ingest pipeline is an array. 

Let me compare with EPR's: `packages/example/coredns/dataset/log/manifest.yml` - ingest_pipeline - single value.

Do you have any preferences against selecting the default pipeline? Name? first in directory list?

dev/packages/beats/activemq-0.0.1/dataset/audit/elasticsearch/ingest-pipeline/pipeline.yml
ruflin: In the case of Elasticsearch, it is the `pipeline.yml` file as this is the one doing the routing. I had to dig into the LS one to see why it even works. It turns out this is an anti pattern from moving forward. It decides the ingest pipeline based on the config. Instead there should always be 1 pipeline. The LS module should be modified to use the ES pattern, otherwise it will not work with the packages.

@ph @ycombinator @exekias We should change the LS module.

ruflin:
EPM can have multiple pipelines, but just one is referenced in the manifest. The others are referenced inside the ingest pipelines itself.

mtojek:
> EPM can have multiple pipelines, but just one is referenced in the manifest. The others are referenced inside the ingest pipelines itself.

I will implement the decision logic for selecting right `ingest_pipeline` while working on the next issue (input config). Hope that works for you.

dev/packages/beats/activemq-0.0.1/dataset/audit/elasticsearch/ingest-pipeline/pipeline.yml
ycombinator: Yes, this has been on the cards for a while anyway: https://github.com/elastic/beats/issues/9964.

mtojek: Fixed.

mtojek:
> Ignore this comment. I still need the logic to determine which file should be named `default.json` / `default.yml`.

Fixed.

dev/packages/beats/activemq-0.0.1/dataset/audit/elasticsearch/ingest-pipeline/pipeline.yml
ycombinator: > Fixed.

Hey @mtojek, just to clarify, we still need elastic/beats#9964 to remain open, right?

mtojek: > > Fixed.
> 
> Hey @mtojek, just to clarify, we still need [elastic/beats#9964](https://github.com/elastic/beats/issues/9964) to remain open, right?

Yes, would be great to fix this.

ruflin: @ycombinator Yes, it now even becomes a must to change to migrate to packages.

ycombinator: Thanks for clarifying. @ruflin other than "ASAP" 😄 what's the timeline you need this to be fixed in?

PR #282, opened 2020-03-19
Add misssing field definitions for system.process

ruflin:
The wrong fields were listed in the dataset.


ruflin:
Adding as draft as some things seem to be still off.

PR #286, opened 2020-03-23
Remove generated packages and enabled beats packages

ruflin:
In the beginning, there were a few generated packages to fill the screen in EPM. With the Beats module migration script kicked off, we now have more then enough packages avaiable. Because of this the existing generated packages are removed and the Beats packages are enabled during the check and build process.

There are some example packages which overlap with the Beats packages like system or nginx and it is at the moment important that we keep using the manual built packages for testing because they contain all the details. This is not a problem as all the Beats packages have version 0.0.1. As long as the manual built packages have a newer version, things will just work.


ruflin:
@mtojek If we merge this, it means moving forward all the things that the migration script does are required to be valid. Not sure if this could become a problem in some cases or is a feature.

@nchaulet I tested if system and nginx package can still be installed as expected with this. It seems to work but please double check for your demo.

mtojek:
> @mtojek If we merge this, it means moving forward all the things that the migration script does are required to be valid. Not sure if this could become a problem in some cases or is a feature.

Not a problem at all. I think we can involve this in the CI process to make sure we're green.

nchaulet:
@ruflin Yes everything work as expected for my demo, (I also installed a local package registry just in case something broke), the nginx log dashboard is still not showing data

PR #289, opened 2020-03-23
import-beats: Add metric input vars

mtojek:
This PR extends "import-beats" with support for metric input vars.

_Let's start with this generated output and mark all the configuration options that should be fixed._

### DIFF
--- a/dev/import-beats/streams.go
+++ b/dev/import-beats/streams.go
@@ -5,9 +5,13 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
+	"log"
+	"os"
 	"path"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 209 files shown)


dev/import-beats/streams.go
```diff
@@ -75,10 +85,109 @@ func wrapVariablesWithDefault(mwvs manifestWithVars) manifestWithVars {
 	return withDefaults
 }
 
-func createMetricStreams() ([]util.Stream, error) {
+func createMetricStreams(modulePath, moduleName, datasetName string) ([]util.Stream, error) {
```
ruflin: Some function comments on these methods could be helpful to explain reviewers and future contributors, on what actually happens inside here.

mtojek: Fixed.

PR #290, opened 2020-03-23
Remove currentDir code from collect assets

ruflin:
The goal is to remove all directory changes in the generator code. This should make the code more predictable and makes it possible to run the generation in parallel.

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -113,11 +113,6 @@ func buildPackage(packagesBasePath string, p util.Package) error {
 	if err != nil {
 		return err
 	}
-	err = os.Chdir(packagesBasePath)
-	if err != nil {
-		return err
-	}
-	defer os.Chdir(currentPath)
 
 	// Checks if the package is valid
 	err = p.Validate()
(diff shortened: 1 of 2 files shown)


dev/generator/main.go
```diff
@@ -113,23 +113,26 @@ func buildPackage(packagesBasePath string, p util.Package) error {
...
-	}
-	defer os.Chdir(currentPath)
 
 	// Checks if the package is valid
 	err = p.Validate()
 	if err != nil {
 		return fmt.Errorf("Invalid package: %s: %s", p.GetPath(), err)
 	}
 
+	p.BasePath = currentPath + "/" + packagesBasePath + "/" + p.GetPath()
```
mtojek: nit: path.Join

dev/generator/main.go
```diff
@@ -113,23 +113,26 @@ func buildPackage(packagesBasePath string, p util.Package) error {
...
+
 	err = p.LoadAssets(p.GetPath())
 	if err != nil {
 		return err
 	}
 
+	err = os.Chdir(packagesBasePath)
```
mtojek: Not blocking, but I'm not convinced if the app should change directory.

ruflin: This is exactly what I'm trying to get rid off. Following up with more PRs.

dev/generator/main.go
ruflin: Will do a refactoring for all the path.Join at once, more to come.

PR #293, opened 2020-03-24
Remove more usage of os.Chdir on saved objects

ruflin:
Removes the usage of os.Chdir for saved objects. Now only the tar.gz part is left. Needs some more fiddling here to still have the same result.


ruflin:
Closing in favor of https://github.com/elastic/package-registry/pull/294

PR #296, opened 2020-03-24
Add stream fields to each dataset

ruflin:
During the generation of the datasets, a stream.yml file is created with the stream fields inside. These are required for each dataset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Allow to set cache times through config. [#271](https://github.com/elastic/integrations-registry/pull/271)
 * Make README.md file a required file for a package. [#287](https://github.com/elastic/integrations-registry/pull/287)
+*  Add stream fields to each dataset [#296](https://github.com/elastic/integrations-registry/pull/296)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


ruflin:
This PR requires https://github.com/elastic/package-registry/pull/295 to be merged first.

@neptunian @skh This adds the stream fields to each dataset. With this, EPM does not have to do anything special. I checked and it looks like template generation and index pattern generation works with the fields. On the index pattern side: Are constant_keyword fields just treated as keyword or does KB have support for it already?

neptunian:
@ruflin for index patterns there is no support for the `constant_keyword` field.  if the type is unknown, it should not be set, so the field would have [no type set](https://github.com/elastic/beats/blob/master/libbeat/kibana/fields_transformer_test.go#L208).  For mappings, if it doesn't recognize the type, it just passes it through.  I created an issue to support it here https://github.com/elastic/kibana/issues/61112 .  [Keywords are converted to a string for index patterns](https://github.com/elastic/beats/blob/master/libbeat/kibana/fields_transformer_test.go#L203), I imagine we will do the same for constant_keyword?

ruflin:
@neptunian Is it ok to merge this PR as I think it does not break anything at the moment?

Moving the other conversation to https://github.com/elastic/kibana/issues/61112

neptunian:
yes it wont break anything

neptunian:
yes, it won't break anything.

PR #297, opened 2020-03-24
[Endpoint] Bumping the version to 1.0.0 and updating the mappings

jonathan-buttner:
Bumping the version and updating the mappings to include malware alerts information.


jonathan-buttner:
Thanks @ruflin I got a "verbal" from marshal

PR #298, opened 2020-03-24
Refactor code base to use (file)path.Join 

ruflin:
This is a refactoring of the packages and dataset code base. Changes:

* All the joined file paths are replaced by `filepath.Join`
* All the joined paths are replaced by `path.Join`
* basePath passed to package and dataset includes the name of the package / dataset

This change is only refactoring, should not have any effect on the registry service itself.

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -52,7 +52,7 @@ func main() {
 
 func Build(sourceDir, publicDir string) error {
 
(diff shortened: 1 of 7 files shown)


ruflin:
@mtojek It is just refactoring for now and I didn't encounter any issues. But I think its important that we make the diff in the code for code and url paths as soon as someone wants to contribute with a Windows machine.

PR #299, opened 2020-03-25
Refactor: Cleanup constructor of dataset object

ruflin:
The creation of the dataset object was leaked into the package code. This is now moved into its own constructor.

### DIFF
--- a/util/package.go
+++ b/util/package.go
@@ -321,7 +321,7 @@ func (p *Package) LoadDataSets(packagePath string) error {
 			return err
 		}
 
-		// Iterate through all datatsources and inputs to find the matching streams and add them to the output.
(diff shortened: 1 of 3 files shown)


util/package.go
```diff
@@ -304,46 +305,23 @@ func (p *Package) GetDatasets() ([]string, error) {
...
-			d.ID = p.Name + "." + dataset
-		}
-
-		if d.Release == "" {
-			d.Release = "beta"
+			return err
 		}
 
-		// Iterate through all datatsource and inputs to find the matching streams and add them to the output.
+		// Iterate through all datatsources and inputs to find the matching streams and add them to the output.
```
mtojek: nit: typo

ruflin: fixed

PR #300, opened 2020-03-25
Change package path from {packagename}-{version} to {packagename}/{version}

ruflin:
The current package path is `/package/{packagename}-{version}`. EPM will need some info about which packages exist. To make it possible to populate some generic info about a package, the path is changed to `/package/{packagename}/{version}`. No additional info is added yet but it gives you the option to add it.

This PR makes all changes in the code that are needed on the packaging side to make the above happen. Also all test files are updated.

This is a breaking change for EPM.


neptunian:
I think we should create an issue for EPM to make sure it can all be fixed locally, first.
Here are some issues i found:
https://github.com/elastic/kibana/blob/master/x-pack/plugins/ingest_manager/server/services/epm/registry/index.ts#L102
https://github.com/elastic/kibana/blob/master/x-pack/plugins/ingest_manager/server/services/epm/packages/assets.ts#L38

ruflin:
@neptunian Could you take a lead on that and then we can merge and deploy close together?

dev/generator/main.go
```diff
@@ -75,9 +76,9 @@ func Build(sourceDir, publicDir string) error {
 
 // CopyPackage copies the files of a package to the public directory
 func CopyPackage(src, dst string) error {
-	fmt.Println(">> Copy package: " + src)
+	log.Println(">> Copy package: " + src)
 	os.MkdirAll(dst, 0755)
-	err := sh.RunV("cp", "-a", src, dst)
+	err := sh.RunV("rsync", "-a", src, dst)
```
ruflin: @kuisathaverat I switched to `rsync` to have better cross platform consistency. But it is not yet on CI. Is there an easy way to get it there?

neptunian:
@jfsiii I am thinking we need to change the EPM api to mirror this change

jfsiii:
@neptunian I don't know if we need to (I believe the two are and should be independent) but we can. There is already a ticket for using name@version for EPM. We could implement that or change it to mirror this.

neptunian:
@ruflin what exactly is `{package}/{packageName}` endpoint going to return?  Will it be similar to search?package=system&all=true?

ruflin:
@neptunian At the moment it will return nothing. This change only gives us the opportunity to discuss further on how we will use this endpoint. If it turns out, we don't need it at all this is ok, but I would still like to do the change as if we realise later we need it, we can't make this change anymore as it is a breaking change.

About the content it could return: I'm thinking of a much more compact version than what we have in the `/search` endpoint which only returns all the available versions potentially with compatiblity.

Related to the comment from @jfsiii: Agree that we don't need to tie this change to the internal EPM API. This can change on its own pace and see there if it makes to change it at all.

ruflin:
@neptunian README updated.

neptunian:
@ruflin i meant this line: https://github.com/elastic/package-registry/pull/300/commits/65ade9921092f17623dacdb82d8a8b5f435dd59b#diff-04c6e90faac2675aa89e2176d2eec7d8R12

README.md
```diff
@@ -20,8 +20,7 @@ The `/search` API endpoint has few additional query parameters. More might be ad
   a package requires 7.4, the package will not be returned or an older compatible package will be shown.
   By default this endpoint always returns only the newest compatible package.
 * category: Filters the package by the given category. Available categories can be seend when going to `/categories` endpoint.
-* package: Filters by a specific package name, for example `mysql`. In contrast to the other endpoints, it will return
-  by default all versions of this package.
+* package: Filters by a specific package name, for example `mysql`. It will return the most recent version of the package.
```
neptunian: i think you meant to put this in the search PR.  I was referring to updating the API endpoint path one line 12

ruflin: @neptunian argh, sorry. Now did the right adjustement.

ruflin:
Comments from @jfsiii were fixed in https://github.com/elastic/package-registry/pull/312

PR #301, opened 2020-03-25
Change default for /search endpoint with package filter

ruflin:
So far when a `/search?package=foo` query was run, by default all versions of the package found were returned. This was in contrast to that by default only the most recent version of a package is returned. To make the API less surprising the behavior was changed that by default only the most recent package is returned.

As there might be still a need to return all packages, and additional param `all=true` was added. This can now be used in combination with all query params. To get the same result as before, `/search?package=foo&all=true` has to be run. By default all is false.


neptunian:
Can you update the README.md to `/search`'s new behavior.

jfsiii:
What about a query/filter param which takes a semver range? Like `versions` which we have elsewhere in the API. So `versions=*`, `versions=7.x`, etc

ruflin:
@jfsiii I'm happy to implement some of the suggestion above. I think what I want to first fix in this PR is the surprising behaviour that when just running `/search` it returns 1 item per package and when adding a package filter, it returns all. I don't think this is intuitive. I would like to keep the default behaviour of only the newest. If we don't need the all at the moment, I can also remove it from the PR and just go with the query change.

PR #302, opened 2020-03-25
Add agent stream configuration

mtojek:
This PR adds agent stream configuration for logs and metrics input.

Also, a README file for script was prepared.


mtojek:
@ruflin I'm not sure if I transformed the config properly (especially "input", "type" variables). It would be grate if you could go through it.

mtojek:
Hm.. I'm not sure about the quality of import here. Looking at the o365/audit (which is a complex one), I'm not impressed. I'm considering rewriting this routine.

Also, there is a strange pipeline.js file inside?

mtojek:
Ok, refactored. I think it looks better (more clear).

dev/packages/beats/zeek-0.0.1/dataset/capture_loss/agent/stream/stream.yml
```diff
@@ -0,0 +1,16 @@
+input: log
+paths:
+{{#each paths}}}}
+  - {{this}}
+{{/each}}
+exclude_files: [".gz$"]
+tags: {{this.tags}}
```
ruflin: Should this be just `tags`?

mtojek: "this" has been added as a workaround for "range":

```
{{#each animals}}}}
  - {{this.species}}
     {{this.size}}
     {{this.name}}
{{/each}}
```

It should cover all placeholders for variable, not analyzing if the used variable is inside the loop or not.

PR #303, opened 2020-03-25
Fix generated files with stream.yml

ruflin:
It seems when the stream.yml PR was merged, some of the test files were not updated. This PR fixes this.


ruflin:
Directly merged as it is only about updating generated files and bringing CI back to green.

PR #305, opened 2020-03-27
Fill package requirement

mtojek:
This PR adjusts `import-beats` to check package requirements for Elastic dependencies - Kibana, Elasticsearch.

Also, basic spelling correction has been added to adjust MySQL.


dev/import-beats/spelling.go
```diff
@@ -0,0 +1,16 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package main
+
+import (
+	"strings"
+)
+
+var spellingReplacer = strings.NewReplacer(
+	"mysql", "MySQL", "Mysql", "MySQL")
```
exekias: this doesn't belong here. I think we need to fix it in the source data, which is a perfectly fine thing to do

ruflin:
I don't think we should do any spelling corrections as part of the migration and rather fix this in the modules itself.

mtojek:
@exekias @ruflin spelling correction and `strings.Title()` has been removed.

util/package.go
```diff
@@ -260,6 +261,13 @@ func (p *Package) Validate() error {
 		return fmt.Errorf("no description set")
 	}
 
+	if p.Requirement.Elasticsearch.Versions != "" {
+		_, err := semver.ParseRange(p.Requirement.Elasticsearch.Versions)
+		if err != nil {
+			return fmt.Errorf("invalid elasticsesrch versions: %s, %s", p.Requirement.Elasticsearch.Versions, err)
```
ruflin: @mtojek Completely missed that change. Typo here.

Could you leave a quick comments in PR's if you also changed the registry logic. Sometimes its hard to find in the massive diff and Github has the tendency to collapse all the things at the end.

mtojek: https://github.com/elastic/package-registry/pull/306

I apologize for the confusion. I will mention all PR changes next time.

ruflin: Don't worry, I should have reviewed more carefully ;-)

PR #311, opened 2020-03-30
Replace vpc with vpn, natgateway and transitgateway in aws dataset

kaiyan-sheng:
This PR is to remove vpc from aws dataset, add vpn, natgateway and transitgateway instead because of the same change in beats repo https://github.com/elastic/beats/pull/16892. 

Also I changed some dataset from `ga` to `beta` in manifest.yml to match what we currently have in Beats repo.


ruflin:
Thanks @kaiyan-sheng

Issue #278, opened 2020-03-18
Template functions inside README file


ruflin:
Today the README.md file is a pure markdown file. All the data has to be added manually to it. As can be seen in https://github.com/elastic/package-registry/pull/276 also the fields.yml are part of it to document the fields. My idea here is that allowing to use some templates / functions inside the README. These are executed during the build stage, so static files are served.

An example could look as following:

```
Below are the described fields:

{{fields dataset=nginx.access}}

```

The above would create the table for the nginx.access dataset with all the fields inside. The template code is just an example.

Which template language we should use here? Golang templates would be easy to use as the language already comes with it but it is specific to Golang. That the build process is written in Golang is an implementation detail. I would rather use something generic like moustache or handlebars. In our pipelines and config files we already use handlebars, so this would be the most obvious one. We should only use 1 template language in all our package files to keep the mental overhead for package creators low.

PR #313, opened 2020-03-31
Add mysql docs

mtojek:
Changes:
* write basic MySQL README file (output: https://github.com/mtojek/package-registry/blob/add-mysql-docs/dev/packages/beats/mysql-0.0.1/docs/README.md)
* use nginx README file as template
* implement template function to generated exported fields
* apply exported fields to MySQL, nginx READMEs
* fix typo: resouce
* replace default empty README file with single word "TODO" (smaller empty READMEs) 

Resolves: https://github.com/elastic/package-registry/issues/278


dev/packages/beats/nginx-0.0.1/docs/README.md
```diff
@@ -1,15 +1,127 @@
...
+
+### Access Logs
+
+Access logs collects the nginx access logs.
+
+**Exported fields**
+
+| Field | Description | Type |
+|---|---|---|
+| nginx.access.agent | Alias for field "user_agent.original" | alias |
```
ruflin: What about all the ECS fields that are used as part of the access dataset? They should also be documented here.

dev/packages/beats/mysql-0.0.1/dataset/status/manifest.yml
```diff
@@ -8,7 +8,13 @@ streams:
   - default:
     - root:secret@tcp(127.0.0.1:3306)/
     name: hosts
+  - default: secret
```
ruflin: I think we are missing here all the additional fields which are need to make the UI nicer. See https://github.com/elastic/package-registry/blob/master/dev/packages/example/nginx-1.2.0/dataset/access/manifest.yml#L17 for an example.

dev/packages/beats/mysql-0.0.1/dataset/galera_status/manifest.yml
```diff
@@ -4,5 +4,17 @@ release: beta
 type: metrics
 streams:
 - input: mysql/metrics
+  vars:
```
ruflin: Same like below for the UI part

dev/import-beats-resources/nginx/docs/README.md
```diff
@@ -0,0 +1,80 @@
...
+        }
+    },
+    "service": {
+        "address": "127.0.0.1",
+        "type": "nginx"
+    }
+}
+```
+
+{{fields "stubstatus"}}
```
ruflin: Good to see such a feature. I wonder if this should become part of the overall build step instead of the import part. I'm good having it here at the moment.

dev/import-beats-resources/nginx/docs/README.md
```diff
@@ -0,0 +1,80 @@
...
+### Stubstatus Metrics
+
+The Nginx stubstatus stream collects data from the Nginx ngx_http_stub_status module. It scrapes the server status data from the web page generated by ngx_http_stub_status.
+
+This is a default stream. If the host datasource is unconfigured, this stream is enabled by default.
+
+An example event for nginx looks as following:
+
+```$json
+{
```
ruflin: Where did you take this even from? Existing from Beat? This will probably look slightly different when shipped through the agent.

dev/packages/beats/nginx-0.0.1/docs/README.md
```diff
@@ -1,15 +1,127 @@
...
+| nginx.access.body_sent.bytes | Alias for field "http.response.body.bytes" | alias |
```
ruflin: These alias fields are not part of the even anymore. These were only used for migration from 6.x to 7.x and are not needed anymore.

dev/import-beats-resources/nginx/docs/README.md
mtojek: Yes, I will refresh the samples once I run the agent.

dev/import-beats-resources/nginx/docs/README.md
mtojek: I would keep it here at the moment, later on we can think about refactoring this part.

dev/packages/beats/mysql-0.0.1/dataset/status/manifest.yml
mtojek: So this is mostly manual work here. Once I fix it, it would be harder to automatically import beats. Should all options have these properties?

dev/packages/beats/mysql-0.0.1/dataset/galera_status/manifest.yml
mtojek: Responded above

ruflin:
@mtojek An other thing that just came to my mind is that as part of the migration we should also start thinking about optimising the dashboards. For example the query for `event.dataset` should become a query for `stream.dataset` which is a constant keyword and much more efficient. The gent should always add these fields (@michalpristas will know more if this already happens).

I don't think we need to do all these optimisations directly during the migration but could be done later on. But we should start a checklist with these already now so we don't forget.

dev/packages/beats/nginx-0.0.1/docs/README.md
mtojek: Based on the offline discussion - I will skip all the alias fields with the `migration: true` property.

mtojek:
> @mtojek An other thing that just came to my mind is that as part of the migration we should also start thinking about optimising the dashboards. For example the query for `event.dataset` should become a query for `stream.dataset` which is a constant keyword and much more efficient. The gent should always add these fields (@michalpristas will know more if this already happens).

Ok, but this is a new requirement that I wasn't aware before. Probably will have to adjust the dashboard generating process. I will create an issue for this.

dev/packages/beats/nginx-0.0.1/docs/README.md
mtojek: Fixed.

mtojek:
> One thing missing here is the updating the fields.yml files. The problem becomes obvious with the nginx access logs. Historically there was just one massive fields.yml for Filebeat where all the fields were shipped together and we had a tree of definitions: Global, ECS, Module, Fileset. Now that we have a template per Dataset, all the fields used must be defined on the Dataset level. This has the advantage that the template becomes much more compact but brings the challenge, there is no easy way to tell which fields from Global, ECS etc. are used in the dataset. We can't just take all of ECS as this is too many fields and not all are used.
> 
> Having it documented separately also solves an other issue: It allows us to document how an ECS fields is exactly used in a dataset. What does `source.ip` exactly mean in the context of nginx.access logs. So far, we only had a generic doc but now we can fill in the details.

https://github.com/elastic/package-registry/issues/316

dev/import-beats-resources/nginx/docs/README.md
mtojek: https://github.com/elastic/package-registry/issues/317

dev/packages/beats/nginx-0.0.1/docs/README.md
mtojek: Issue created for this.

dev/packages/beats/mysql-0.0.1/dataset/status/manifest.yml
mtojek: https://github.com/elastic/package-registry/issues/318

ruflin: I think most fields will need it, at least part of the options. And yes, this will probably rely on manual work as it is not something we have today.

PR #314, opened 2020-03-31
Fix System Host overview dashboard

ruflin:
On the Beats side, we took this dashboard and replace the hostname field with the local hostname. This had the nice onboarding experience that a user had the data directly filtered on his own host. This does not work anymore so this filter has to be removed. Instead I expect us when linking for example from an agent to a dashboard or similar, that we can apply some filters in advance.


ruflin:
@mtojek @exekias This is one of these issues where I wonder if it is really worth writing some code or just have a checklist or something similar for manual adjustment.

PR #321, opened 2020-03-31
Endpoint schema updates

jonathan-buttner:
This PR adds some missing fields to the mappings and adds the endpoint dashboard.


dev/packages/example/endpoint-1.0.0/kibana/index-pattern/events-endpoint-*.json
```diff
@@ -0,0 +1,15 @@
+{
```
ruflin, line 1: @jonathan-buttner I would assume this index pattern is not needed because it is generated by EPM? Or is there something more specific about this index pattern then `events-*`?

jonathan-buttner: Oh good point, I'll switch it to use `events-*`

Issue #316, opened 2020-03-31
MySQL integration: add ECS fields to docs


mtojek:
Reported in https://github.com/elastic/package-registry/pull/313

>One thing missing here is the updating the fields.yml files. The problem becomes obvious with the nginx access logs. Historically there was just one massive fields.yml for Filebeat where all the fields were shipped together and we had a tree of definitions: Global, ECS, Module, Fileset. Now that we have a template per Dataset, all the fields used must be defined on the Dataset level. This has the advantage that the template becomes much more compact but brings the challenge, there is no easy way to tell which fields from Global, ECS etc. are used in the dataset. We can't just take all of ECS as this is too many fields and not all are used.
Having it documented separately also solves an other issue: It allows us to document how an ECS fields is exactly used in a dataset. What does source.ip exactly mean in the context of nginx.access logs. So far, we only had a generic doc but now we can fill in the details.

mtojek:
Resolving. Merged.

PR #322, opened 2020-03-31
Add ecs fields

mtojek:
Changes:
* write ECS fields
* update beats repo
* convert "mm" to "px" in SVG size

Issue: https://github.com/elastic/package-registry/issues/316


dev/import-beats/fields.go
```diff
@@ -72,3 +73,17 @@ func loadDatasetFields(modulePath, moduleName, datasetName string) ([]byte, erro
 	}
 	return buffer.Bytes(), nil
 }
+
+func loadEcsFields(ecsDir string) ([]fieldsTableRecord, error) {
+	ecsFieldsPath := filepath.Join(ecsDir, "generated/beats/fields.ecs.yml")
```
ruflin, line 78: Good for now but at some point we should start to depend on a specific release of ECS and not master. Best same version Beats is on.

dev/import-beats/images_svg.go
```diff
@@ -39,6 +39,7 @@ func SvgDecodeConfig(r io.Reader) (image.Config, error) {
 		width, err = strconv.ParseFloat(svgFile.Width, 32)
 		if err != nil {
 			svgFile.Width = strings.ReplaceAll(svgFile.Width, "pt", "")
+			svgFile.Width = strings.ReplaceAll(svgFile.Width, "mm", "")
```
ruflin, line 42: Out of curiosity, don't know much about the content of svg: Why does pt and mm have to be replaced?

dev/import-beats/fields.go
mtojek: At the moment, we can switch branches manually on the ECS side.

dev/import-beats/images_svg.go
mtojek: Icons in the Kibana and the EUI repositories have dimensions defined in various formats. Honestly I would enforce to use pixels, but most likely I don't have an influence on that :) hence, convertion.

ruflin: @hbharding Any comments on the above? ^

hbharding: I'm not 100% sure. EUI uses unitless values for width, height, and viewbox, so what you've done looks correct. We can use CSS on our side to make the icon bigger or smaller using pixel values.

@cchaos or @snide, I'm wondering if you have any input on this. Our package registry for ingest manager lets authors define an SVG icon for their package. Are there certain things we should we should enforce in our code like "your svg must be square", "your svg format must be minimized", etc.? Or should we just trust the author knows what they are doing and perhaps we post some guidelines/docs somewhere for authors?

snide: We only use pixels. You can use an `enum` scale that pre-defines that pixel value, but it's always in pixel. As Henry alludes to, CSS can be used to provide custom widths (again in pixels) if you need to go beyond that.

For the third party logos, if you see any icons that AREN'T using pixels in their SVG itself, please make issues and we'll fix them. They came a lot of times from outside sources and people may not have stripped weird sizing formats from them.

mtojek: @snide With pleasure: https://github.com/elastic/kibana/issues/62262

PR #323, opened 2020-04-01
Manually prepare mysql package (alpha)

mtojek:
_I'm opening this as draft to present all manual changes that were applied._

Changes:
* copied content of mysql-0.0.1
* refactored vars according to: https://github.com/elastic/package-registry/issues/318
* extended generator to use "alpha" packages

Action items:
- [x] Wait for fixing https://github.com/elastic/kibana/issues/62330
- [x] Sync with master
- [x] Wait for fixing https://github.com/elastic/kibana/issues/63251
- [x] Wait for fixing https://github.com/elastic/kibana/issues/63726
- [x] Copy to docs real events in agent mode

Closes: https://github.com/elastic/package-registry/issues/317
Closes: https://github.com/elastic/package-registry/issues/318
Closes: https://github.com/elastic/package-registry/issues/319
Closes: https://github.com/elastic/package-registry/issues/320

Screenshot:
screencapture-localhost-5601-ekv-app-ingestManager-2020-04-01-15_48_02

PR #327, opened 2020-04-02
Remove migrated not-ECS migrated fields

mtojek:
This PR addresses comment reported in https://github.com/elastic/package-registry/pull/323/files/b4ffdfb6aafe5a70f2ff8ee6ed60c9788dd7f1cb#r402212487

All fields files have been parsed and re-marshalled.

Difference can be also spotted in MySQL README file.


dev/packages/beats/aerospike-0.0.1/dataset/namespace/fields/fields.yml
```diff
@@ -1,158 +1,158 @@
...
+  - name: client
+    type: group
+    description: |
+      Client stats.
+    fields:
+    - name: delete
       type: group
-      description: >
-        Client stats.
+      description: |
```
ruflin, line 14: Not related directly to this PR but something we should discuss for the future: We have these group descriptions for historical reasons but not 100% sure if these still will make sense in the future. Especially as these are for docs only and were always a challenge to visualise.

dev/packages/beats/mysql-0.0.1/dataset/error/fields/fields.yml
```diff
@@ -1,17 +1,13 @@
...
-      migration: true
-    - name: level
-      type: alias
-      path: log.level
-      migration: true
-    - name: message
-      type: alias
-      path: message
-      migration: true
+  - name: level
```
ruflin, line 6: @mtojek I think I don't fully understand the change. Why is thread_id removed but not level here?

mtojek: As the PR title suggests I removed all fields marked with "migration" which are not ECS fields. In the other PR, you asked for removing it: https://github.com/elastic/package-registry/pull/323#discussion_r402241871

If I should remove all fields with `migration: true`, don't we need ecs.yml in every fields folder? As we don't have it there, I left ECS fileds in place.

dev/packages/beats/aerospike-0.0.1/dataset/namespace/fields/fields.yml
mtojek: I didn't know about this. Any opinion on this @exekias ?

dev/packages/beats/mysql-0.0.1/dataset/error/fields/fields.yml
ruflin: We need to remove all migration fields, also the ones that point to ECS fields.

We will need a subset of the ecs.yml field in most datasets, but not all of ecs fields. We only need the fields from ECS that are actually used by the dataset. For example if `log.level` is used, it needs to be in, but if we don't use source.ip, it shouldn't be.

dev/packages/beats/aerospike-0.0.1/dataset/namespace/fields/fields.yml
exekias: I'm definitely ok with removing something we don't really show in the docs (or somewhere else)

mtojek: I've updated the PR, removed all descriptions as you suggested.

dev/packages/beats/mysql-0.0.1/dataset/error/fields/fields.yml
mtojek: I've updated the PR.

Used ECS fields are stored in dedicated `ecs.yml`. Also, all used fields are in corresponding README files.

dev/packages/beats/auditd-0.0.1/dataset/log/fields/package-fields.yml
```diff
@@ -1,112 +1,62 @@
...
-        - name: group.id
-          type: keyword
-          description: >
-            Unique identifier for the group on the system/platform.
-        - name: group.name
-          type: keyword
-          description: >
-                Name of the group.
+  - name: terminal
+    type: keyword
```
ruflin, line 5: The description is now removed here and this is not a group. Not sure if this is on purpose?

mtojek: > Did you by chance remove for descriptions then you wanted? Only the group ones should be removed.

Misunderstanding. I corrected the source code. Only group fields have stripped description now.

> This only works for modules where we actually migrated fields to ECS and did not use ECS from day one. Newer modules have used ECS from day one so you can't find the ECS fields based on a migration flag.

I adjusted the PR:
1. If the field is an alias and has `migration: true`, then skip it.
2. If the field is an alias and the path is same as ECS, then skip it.

dev/packages/beats/apache-0.0.1/dataset/access/fields/package-fields.yml
```diff
@@ -1,130 +1,14 @@
...
-            - name: city_name
-              type: alias
-              path: source.geo.city_name
-              migration: true
-            - name: region_iso_code
-              type: alias
-              path: source.geo.region_iso_code
-              migration: true
-    - name: error
+    - name: geoip
```
ruflin, line 9: This is interesting. We have a group with no entries. I wonder what EPM does with it 🤔 @skh @neptunian

PR #328, opened 2020-04-02
remove lifecycle.rollover_alias from base index templates

neptunian:
this can be removed because we are going to be using datastreams which does not require this to be configured.


ruflin:
@neptunian Thanks, directly merged and will roll it out.

PR #335, opened 2020-04-09
Add stream template to package json

ruflin:
The index.json file for a package so far did contain all information about streams but was missing the content of the stream.yml file. This PR adds it to the index.json file to simplify the building of the UI.

With this, the stream.yml becomes a requirement for a dataset. Because of this all missing stream.yml are added / renamed.


ruflin:
Merged this for now. I might follow up with some more config options around stream.yml as at the moment only 1 is possible, but multiple inputs could use the same dataset.

Issue #179, opened 2019-12-10
Document usage of handlebars


ruflin:
Handlebard templates are currently used for the agent configs. This should be documented.

PR #338, opened 2020-04-12
Add Netflow integration 0.0.2

alakahakai:
This PR adds the Netflow integration as a replacement of the Filebeat netflow module. 

Here are some screenshots of the ingest manager UI.

image
image
image


dev/packages/example/cisco-1.0.0/dataset/asa/filebeat/config/input.yml
```diff
@@ -0,0 +1,15 @@
+{{#if input == "syslog"}}
```
ruflin, line 1: This will become very interesting as it actually uses 2 different inputs which we support in our manifest but did not have an example for it yet. I think this will have to go into 2 different files.

I wonder why you keep the filebeat config inside?

dev/packages/example/cisco-1.0.0/dataset/asa/filebeat/module.yml
```diff
@@ -0,0 +1,3 @@
+- module: cisco
```
ruflin, line 1: I don't think moving forward we need all the filebeat files?

dev/packages/example/cisco-1.0.0/dataset/asa/agent/input/input.yml
```diff
@@ -0,0 +1,14 @@
+{{#if input == "syslog"}}
```
ruflin, line 1: My comment further down in the filebeat config should have gone here.

Also if I remember correctly, the directly should be called stream/stream.yml

dev/packages/example/cisco-1.0.0/dataset/ftd/agent/input/input.yml
```diff
@@ -0,0 +1,14 @@
+{{#if input == "syslog"}}
```
ruflin, line 1: Same applies here, seems like we support 2 inputs.

dev/packages/example/cisco-1.0.0/manifest.yml
```diff
@@ -0,0 +1,32 @@
...
+- src: /img/cisco.svg
+  title: cisco
+  size: 216x216
+  type: image/svg+xml
+datasources:
+- name: cisco
+  title: Cisco logs
+  description: Collect logs from Cisco instances
+  inputs:
+  - type: logs
```
ruflin, line 30: In the configs you have also the syslog input?

dev/packages/example/netflow-1.0.0/dataset/log/agent/input/input.yml
```diff
@@ -0,0 +1,21 @@
+input: netflow
```
ruflin, line 1: And we have also netflow input here!

mtojek:
Did you use the import-beats script?

dev/packages/example/cisco-1.0.0/dataset/asa/filebeat/module.yml
alakahakai: I was keeping them here because the coredns package had them. I wasn't sure where they were needed  and I could not test end-to-end because fleet was not working in ingest manager.

alakahakai:
> Did you use the import-beats script?

Yes, but I had to make changes according to the coredns package, which @ruflin asked me to follow. I will give it a try end-to-end once the fleet function is working in ingest manager. It was crashing Kibana when I tried it last time.

mtojek:
> > Did you use the import-beats script?
> 
> Yes, but I had to make changes according to the coredns package, which @ruflin asked me to follow. I will give it a try end-to-end once the fleet function is working in ingest manager. It was crashing Kibana when I tried it last time.

When you notice that it crashed or generated dashboards in a wrong way, please open an issue and bring samples. It will help improve the generating process.

dev/packages/example/cisco-1.0.0/dataset/asa/agent/input/input.yml
andrewkroh, line 1: What's the templating language used in these files? 

@ruflin I think it would be nice to use the appropriate file extension for the template type (e.g. .erb, .j2, .mustache) to make it clear to editors (both the people and the tools).

ruflin: It is handlebars (reminds me off https://github.com/elastic/package-registry/issues/179). And the above is actually not valid handlebars :-( Having support for 2 inputs is coming here: https://github.com/elastic/package-registry/pull/346

I agree we should probably adjust the file extension to make it better for the tooling. If we do, I wonder if we should the same for ingest pipeline files which reference an other pipeline. The challenge I see here is that 99% of the pipelines don't use it but a very few do. Does this still mean all should have to have a handlebars suffix or do we just check for both files?

andrewkroh: I think it would be easier if files that use templating had an extension that indicates it. Then when writing the code that uses the modules you know when you need to preprocess the file, and then you don't need to guess or unnecessarily preprocess every file like we do today in Beats.

alakahakai:
Made this for netflow only. Will open another one for cisco.

ruflin:
@alakahakai Let me know when it's ready for review.

dev/packages/example/cisco-1.0.0/dataset/asa/agent/input/input.yml
ruflin: I wonder if supporting both types and not process would complicate the code? What is your concern about processing files that don't need processing?

andrewkroh: My concern is that you need to make sure you haven't used content that requires escaping in the templating language. So if you don't need those features you just call the file `config.yml` instead of `config.yml.handlebars`. Hopefully that isn't too much of issue in most files. Maybe we just process all files through the template engine for now, but put the template extension on them now.

I'd just be happy if my editor could easily recognize that these are template type X and recommend the appropriate plugin. 😄 

Screen Shot 2020-04-21 at 9 43 50 PM

ruflin: Should we go with `config.yml.hbs` or `config.hbs`? I skiped `handlebars` as it is very long and `.hbs` seems to work at least in my IDE too. I kind of like the two suffix so the end goal format is also in there.

andrewkroh: config.yml.hbs SGTM

mtojek: @ruflin if this is an update to the format of package, please open an issue in the repository for adjusting all packages.

mtojek: I introduced the `.yml.hbs` naming in this PR: https://github.com/elastic/package-registry/pull/380

ruflin: @mtojek thanks for taking care of it. I assume this make the issue obsolete.

dev/packages/alpha/netflow-0.0.2/dataset/log/agent/stream/stream.yml
```diff
@@ -0,0 +1,21 @@
+input: netflow
+protocols: [v1, v5, v6, v7, v8, v9, ipfix]
+host: '{{this.netflow_host}}:{{this.netflow_port}}'
+max_message_size: '{{this.max_message_size}}'
+expiration_timeout: '{{this.expiration_timeout}}'
+queue_size: {{this.queue_size}}
+{{#if this.timeout}}
```
ruflin, line 7: I'm stumbling over the `this` here. @mtojek do you know if this is necessary?

dev/packages/alpha/netflow-0.0.2/dataset/log/fields/ecs.yml
```diff
@@ -0,0 +1,3141 @@
+- name: '@timestamp'
```
ruflin, line 1: Does this only contain the fields that are used by this dataset or is this all of ECS? The goal should be to only have the fields in here that are actually used.

dev/packages/alpha/netflow-0.0.2/docs/README.md
```diff
@@ -0,0 +1 @@
+# Netflow module
```
ruflin, line 1: Would be nice to have a bit more docs here as this is shown in the UI.

dev/packages/alpha/netflow-0.0.2/kibana/dashboard/34e26884-161a-4448-9556-43b5bf2f62a2.json
```diff
@@ -0,0 +1,294 @@
...
+                            "negate": false,
+                            "params": {
+                                "query": "netflow"
+                            },
+                            "type": "phrase",
+                            "value": "netflow"
+                        },
+                        "query": {
+                            "match": {
+                                "input.type": {
```
ruflin, line 26: I wonder if this query will still be needed as I assume all the data in the index will be related to netflow. Should this be replaced by a query for type + dataset?

dev/packages/alpha/netflow-0.0.2/dataset/log/agent/stream/stream.yml
mtojek: This is already fixed, even though both forms (with/without `this`) are correct.

The previous implementation of processing stream configuration left `this` for consistency. If you upgrade to the new implementation, you will see a better formed file: https://github.com/elastic/package-registry/blob/master/dev/packages/beats/netflow-0.0.1/dataset/log/agent/stream/netflow.yml.hbs

dev/packages/alpha/netflow-0.0.2/dataset/log/fields/ecs.yml
andrewkroh: So using the netflow module as an example the fields consist of:

- some base filebeat fields that may not be in ECS (like input.type)
- all filebeat netflow input fields (some 400+ fields)
- a subset of ECS namespaces (like `agent`, `source`, `destination`, `network`, `event`)

Very few of these fields are something that the module actually creates. This module basically brings together Filebeat with an Ingest Node pipeline to add geo location and autonomous system numbers. So keeping this module updated will be a lot of work. And since it depends on Filebeat's fields, this template will vary based on the Filebeat version.

So I wonder if it would be possible to list some dependent templates in addition to custom fields created by a module to create a composite. Like saying that for this module it depends on

- filebeat $automatic_version | include_namespaces(agent | netflow | input.type | source | destination)
- ECS 1.4 | include_namespaces(source | destination)

Screen Shot 2020-04-28 at 10 35 55 AM

ruflin: @andrewkroh Nice diagram!

v2 Elasticsearch templates support components and one of the reasons we pushed for this is exactly the above problem. So long term, I'm pretty sure we will support this. Short term, I think we should get this into the development process. The current plan is to use the integrations repository for the development of the packages. There you could even use sysmlinks or whatever reference you need to in the end build the final package with all the files you need for the above. This keeps the final package simple for now.

I was not aware that netflow has 400+ fields (sounds like a lot) so in this case, the template will be pretty big in any case. Good news it is only the netflow template and does not affect anything else. Also it shouldn't be a problem if we have too many fields right now and later optimise as during update, the template will be overwritten.

andrewkroh: Yeah, I think some symlinks could accomplish this. Does the current module more allow for more than one fields file?

ruflin: you can have as many as you want as long as they end in `.yml` and are in the correct directory.

mtojek:
Here are some spottings on my side:

* NetFlow log logs: maybe NetFlow standard logs?
* netflow_host, netflow_port: the prefix netflow is redundant here
* all configuration variables: UI labels should be human readable and not contain "_" (underscore)
* not sure if we need to show to user all variables - maybe just host and port?
* all configuration variables expect host and port: it would be feasible to leave a one sentence description
* all configuration variables should contain information about their types (text, number, etc.)

PR #341, opened 2020-04-14
Add support for updating vendor

mtojek:
This PR extends the `mage`  command with support for refreshing the `vendor` directory.


go.sum
```diff
@@ -6,10 +6,6 @@ github.com/elastic/go-ucfg v0.7.0 h1:1+C/sZdJKww8hKl7XtLPTjs4cFslhQF2fazKTF+ZE+4
 github.com/elastic/go-ucfg v0.7.0/go.mod h1:iaiY0NBIYeasNgycLyTvhJftQlQEUO2hpF+FX0JKxzo=
 github.com/gorilla/mux v1.7.2 h1:zoNxOV7WjqXptQOVngLmcSQgXmgk4NMz1HibBchjl/I=
 github.com/gorilla/mux v1.7.2/go.mod h1:1lud6UwP+6orDFRuTfBEV8e9/aOM/c4fVVCaMa2zaAs=
-github.com/jstemmer/go-junit-report v0.9.1 h1:6QPYqodiu3GuPL+7mfx+NwDdp2eTkp9IfEUpgAwUN0o=
```
ruflin, line 9: I think we need these for Jenkins. @andrewkroh might know more here?

magefile.go
```diff
@@ -210,3 +215,23 @@ func Clean() error {
 
 	return os.Remove("package-registry")
 }
+
+func Vendor() error {
+	fmt.Println(">> mod - updating vendor directory")
+
+	err := sh.RunV("go", "mod", "tidy")
```
ruflin, line 222: Perhaps we can remove tidy in this PR so it doesn't remove the go-junit part?

ruflin:
@mtojek Hm, I perhaps just found a problem with this. Will this create a diff every time a dependency is updated? Meaning a PR could pass CI today but fails tomorrow? This should not be the case.

go.sum
mtojek: I updated the code to add the `go-junit-report` dependency after `go mod tidy`.

magefile.go
mtojek: I think I solved this, see above.

mtojek:
> @mtojek Hm, I perhaps just found a problem with this. Will this create a diff every time a dependency is updated? Meaning a PR could pass CI today but fails tomorrow? This should not be the case.

The dependency graph can be updated with the `go get` command. The command is not executed by `mage check`, so the PR won't be updated automatically while bulding in CI. Also, there is a vendor directory to prevent the original sources from disappearing/changing.

Please correct me if I'm wrong.

Issue #336, opened 2020-04-10
prometheus/collector: conflicting name


mtojek:
I've just spotted that there is a naming conflict in Beats:
- `x-pack/metricbeat/prometheus/collector`
- `metricbeat/prometheus/collector`

These modules datasets are related to different features:
https://github.com/elastic/beats/tree/master/metricbeat/module/prometheus/collector
https://github.com/elastic/beats/tree/master/x-pack/metricbeat/module/prometheus/collector

ruflin:
@exekias Can you elaborate on how the two are related? Do we need both? Is one a (future) input?

exekias:
They are pretty similar (and share part of the implementation), the difference is that the x-pack version makes use of `histogram` field in ES, which is only available in basic.

As the agent will be basic we only care about the x-pack implementation there. I would suggest that the script goes over OSS first and overrides any collision with what's found in x-pack maybe?

ruflin:
++, or we have a blacklist that we don't want to convert as this seems to be only one we need magic. Whatever is easier to implement.

PR #344, opened 2020-04-14
Resolve dataset naming conflict

mtojek:
This PR adjusts the dataset visiting code to resolve naming conflicts.

In case of a dataset conflict, same name, same type (logs, metrics), the latter dataset will replace existing one:
`prometheus/collector` will be replaced with `x-pack/prometheus/collector`.

Also, cleaned some old files.

Closes: https://github.com/elastic/package-registry/issues/336


dev/import-beats/packages.go
```diff
@@ -54,10 +54,10 @@ func (pc *packageContent) addDatasets(ds []datasetContent) {
 				if v.beatType != dc.beatType {
 					pc.datasets[i].name = fmt.Sprintf("%s-%s", pc.datasets[i].name, pc.datasets[i].beatType)
 					dc.name = fmt.Sprintf("%s-%s", dc.name, dc.beatType)
+					pc.datasets = append(pc.datasets, dc)
 				} else {
-					dc.name = fmt.Sprintf("%s-%s", dc.name, dc.beatType)
+					pc.datasets[i] = dc
```
ruflin, line 59: It would be nice to log if we overwrite a dataset so at least looking at the log output we could tell how many are overwritten.

mtojek: Fixed

PR #351, opened 2020-04-16
Prepare basic testing environment for integrations

mtojek:
This PR prepares a basic testing environment that will boot up containers:
* Elasticsearch
* Kibana with enabled Ingest Manager
* Integrations registry

Let's consider this as grounds for the Integrations Developer Environment. I wonder if it's possible to use elastic-agent image and enroll it, so that the final setup would be ES, Kibana, Package Registry and an enrolled agent.

### DIFF
--- a/testing/environments/README.md
+++ b/testing/environments/README.md
@@ -1,0 +1,11 @@
+Firstly, refresh docker images:
+
+```bash
+$ docker-compose -f snapshot.yml pull
+```
+
+Run docker containers:
+
+```bash
+$ docker-compose -f snapshot.yml -f local.yml up --force-recreate kibana integrations-registry
+```
--- a/testing/environments/README.md
+++ b/testing/environments/README.md
@@ -7,5 +7,5 @@ $ docker-compose -f snapshot.yml pull
 Run docker containers:
 
 ```bash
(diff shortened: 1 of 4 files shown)


mtojek:
> Tried this out locally and works as expected. This is a great start.
> 
> One thing I was thinking if we should perhaps run the registry locally instead of a container and connect to it (if possible). This would make it possible to change the package, run mage build and run the registry to test the new package. Perhaps this is also possible with a container?
> 
> The above would also have the side effect that we need an easy way in Kibana to "reinstall" a package again and not use anything cached. Delete and install might be the best option here to refresh it without having to restart ES and KB as this from my perspective takes too long. (@neptunian for your awareness)

I hear your voice. Accessing the registry locally would mean communication across namespaces or bridging. At the moment I used to rebuild the registry image and restart, but I think we can improve it.

I also had an idea to include in the cluster a container with the agent, but as far as I know, there isn't any? Should I prepare it? We can also add some orchestration code to enroll the agent, so the final result would be a running cluster with enrolled agent that you can use for whatever you want. WDYT @ruflin ?

ruflin:
I think there is already a container for the agent and @michalpristas works also on a simple setup that does the above. You two should sync up / team up.

PR #352, opened 2020-04-16
import-beats: adjust generated files

mtojek:
Changes in `import-beats`:
* fields: add default type "keyword" when missing
* change input title from "Collecting..." to "Collect"
* regenerate files

PR #354, opened 2020-04-20
import-beats: Decode fields layerListJSON, mapStateJSON

mtojek:
This PR decodes two Kibana fields that have been already encoded in Beats: layerListJSON, mapStateJSON.

It's a followup of the PR comment: https://github.com/elastic/package-registry/pull/352#discussion_r411161119

Actually I'm not sure if we should store them decode, could somebody from the Kibana team confirm that's correct?

### DIFF
--- a/dev/import-beats/kibana.go
+++ b/dev/import-beats/kibana.go
@@ -19,15 +19,7 @@ import (
 )
 
 var (
-	fieldsToEncode = []string{
-		"attributes.kibanaSavedObjectMeta.searchSourceJSON",
-		"attributes.optionsJSON",
-		"attributes.panelsJSON",
-		"attributes.uiStateJSON",
-		"attributes.visState",
-	}
-
-	fieldsToDecode = []string{
+	encodedFields = []string{
 		"attributes.kibanaSavedObjectMeta.searchSourceJSON",
 		"attributes.layerListJSON",
 		"attributes.mapStateJSON",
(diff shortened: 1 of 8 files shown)


dev/import-beats/kibana.go
```diff
@@ -19,12 +19,22 @@ import (
...
+	}
+
+	fieldsToDecode = []string{
+		"attributes.kibanaSavedObjectMeta.searchSourceJSON",
+		"attributes.layerListJSON",
+		"attributes.mapStateJSON",
 		"attributes.optionsJSON",
 		"attributes.panelsJSON",
-		"attributes.kibanaSavedObjectMeta.searchSourceJSON",
+		"attributes.uiStateJSON",
```
ruflin, line 36: Why is encoded and decode not the same list?

Could you import this list from here https://github.com/elastic/package-registry/blob/master/dev/generator/main.go#L253 so we have a single point of truth?

mtojek: It's not the same list, because some fields are already encoded in dashboards stored in Beats (see mapStateJSON in https://github.com/elastic/beats/blob/master/x-pack/filebeat/module/o365/_meta/kibana/7/dashboard/Filebeat-O365-Audit.json#L842).

Another solution would be detecting if the field is "string" and then try to decode it. WDYT?

mtojek:
> At the moment, Kibana only supports encoded objects. If we change this list, we also need to change it here: https://github.com/elastic/package-registry/blob/master/dev/generator/main.go#L253
> 
> Storing it decode is for versioning purpose to be able to review diffs. This is something specific to packages.

I responded in the other comment.

dev/import-beats/kibana.go
ruflin: Oh, I see. So we basically missed it already in Beats, I wonder if we should add it there @exekias .

A check for string would be nice as it would mean we only have to keep one list and things to break if the dashboards in Beats itself are adjusted.

mtojek: Fixed!

ruflin:
Your code changes LGTM but you still have to update https://github.com/elastic/package-registry/blob/master/dev/generator/main.go#L253 with the new saved object list as otherwise the packages will be broken. I suggest we merge the two into 1, one using the other.

mtojek:
> Your code changes LGTM but you still have to update https://github.com/elastic/package-registry/blob/master/dev/generator/main.go#L253 with the new saved object list as otherwise the packages will be broken. I suggest we merge the two into 1, one using the other.

Fixed.

ruflin:
For linking, found the issue related to encoding objects in Kibana: https://github.com/elastic/kibana/issues/43673

PR #355, opened 2020-04-20
import-beats: Support auth in Kibana

mtojek:
This PR adds support for Auth in Kibana, as it's enabled by default in the testing environment.

### DIFF
--- a/dev/import-beats/kibana.go
+++ b/dev/import-beats/kibana.go
@@ -33,7 +33,11 @@ type kibanaContent struct {
 }
 
 type kibanaMigrator struct {
-	hostPort   string
+	hostPort    string
+	authEnabled bool
+	username    string
+	password    string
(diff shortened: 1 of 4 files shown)


dev/import-beats/kibana.go
```diff
@@ -61,6 +68,9 @@ func (km *kibanaMigrator) migrateDashboardFile(dashboardFile []byte, moduleName
 		return nil, errors.Wrapf(err, "creating POST request failed")
 	}
 	request.Header.Add("kbn-xsrf", "8.0.0")
+	if km.authEnabled {
```
ruflin, line 71: Do you need authEnabled or could just check here if a username is set? -1 param like this.

mtojek:
> Will this also need some update to the docs?

Updated.

PR #360, opened 2020-04-21
import-beats: Fix: do not encode twice

mtojek:
This PR fixes bug in the `generator`, which performed encoding twice on Kibana objects. Some objects (e.g. in the `example` package) are already stored in the encoded form.

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -266,7 +266,6 @@ var (
 // The reason is that for versioning it is much nicer to have the full
 // json so only on packaging this is changed.
 func encodedSavedObject(data []byte) (string, error) {
-
 	savedObject := MapStr{}
 	json.Unmarshal(data, &savedObject)
 
@@ -281,7 +280,7 @@ func encodedSavedObject(data []byte) (string, error) {
 		// In this case skip the encoding.
 		_, isString := out.(string)
 		if isString {
-			continue
+			return "", fmt.Errorf("expect non-string field type (fieldName: %s)", v)
 		}
 
 		// Marshal the value to encode it properly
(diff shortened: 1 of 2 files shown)


dev/generator/main.go
```diff
@@ -277,6 +277,13 @@ func encodedSavedObject(data []byte) (string, error) {
 			continue
 		}
 
+		// It may happen that some objects existing in example directory might be already encoded.
+		// In this case skip the encoding.
+		_, isString := out.(string)
+		if isString {
+			continue
```
ruflin, line 284: As I would consider this an invalid package it would be nice if we could also add a log output here so we can fix the packages. Or more radical, return an error?

mtojek: Fixed.

dev/packages/example/endpoint-1.0.0/kibana/map/a3a3bd10-706b-11ea-9bc8-6b38f4d29a16.json
```diff
@@ -28,8 +28,152 @@
...
-        "mapStateJSON": "{\"zoom\":0.71,\"center\":{\"lon\":-72.02031,\"lat\":-18.76202},\"timeFilters\":{\"from\":\"now-15d\",\"to\":\"now\"},\"refreshConfig\":{\"isPaused\":false,\"interval\":0},\"query\":{\"query\":\"\",\"language\":\"kuery\"},\"filters\":[]}",
+        "layerListJSON": [
```
ruflin, line 31: @jonathan-buttner FYI

jonathan-buttner: Oh nice, thanks!

PR #363, opened 2020-04-21
Adding removable flag

jonathan-buttner:
Adding removable flag to the endpoint package


ruflin:
Directly merged so I can deploy it and it is available.

PR #364, opened 2020-04-21
Support YAML config options

mtojek:
Changes:
- support YAML fields
- determine if field is required
- refresh beats

### DIFF
--- a/dev/packages/beats/apache-0.0.1/manifest.yml
+++ b/dev/packages/beats/apache-0.0.1/manifest.yml
@@ -9,7 +9,7 @@ categories:
 - logs
 - metrics
 release: ga
-removable: false
+removable: true
 requirement:
   kibana:
     versions: '>=7.3.0 <8.0.0'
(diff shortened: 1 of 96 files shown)


mtojek:
Redis key `key.pattern` option added in https://github.com/elastic/beats/pull/17875

dev/packages/beats/apache-0.0.1/manifest.yml
```diff
@@ -9,6 +9,7 @@ categories:
 - logs
 - metrics
 release: ga
+removable: false
```
ruflin, line 12: This should be true as the default is true. I suspect an error in the package code?

mtojek: Fixed.

dev/packages/beats/system-0.0.1/manifest.yml
```diff
@@ -9,6 +9,7 @@ categories:
 - logs
 - metrics
 release: ga
+removable: true
```
ruflin, line 12: Nit: This is where it is false. But I think we can do this as a manual adjustment.

mtojek: Fixed.

If this is only for the system package, I guess that this option could be inverted - single property for non-removable/locked/etc.

PR #366, opened 2020-04-22
Fix ingeset pipeline path for system datasets

ruflin:
The auth and syslog system dataset contained a wrong path to the ingest pipeline which meant it is not loaded. This fixes it.


util/dataset.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/elastic/go-ucfg/yaml"
 )
 
+const (
+	DirIngestPipeline = "ingest-pipeline"
```
ruflin, line 20: @mtojek I'm tempted to say we should rename this to `ingest_pipeline` but don't have a strong argument for either. Any opinion?

ruflin:
@neptunian This fixes the logs for the system package.

dev/import-beats/packages.go
```diff
@@ -324,7 +324,7 @@ func (r *packageRepository) save(outputDir string) error {
 
 			// dataset/elasticsearch
 			if len(dataset.elasticsearch.ingestPipelines) > 0 {
-				ingestPipelinesPath := filepath.Join(datasetPath, "elasticsearch", "ingest-pipeline")
+				ingestPipelinesPath := filepath.Join(datasetPath, "elasticsearch", util.DIR_INGEST_PIPELINE)
```
mtojek, line 327: is this correct path (caps)?

util/dataset.go
mtojek: I'm fine with this path.

dev/import-beats/packages.go
ruflin: :-( No. Seems like my IDE did not catch this during refactoring :-(

util/dataset.go
ruflin: Which one, `-` or `_`?

mtojek: I mean, I'm with both, so also with the underscore :)

ruflin: I was hoping you would have a preference to tip it in one direction :-)

PR #367, opened 2020-04-22
Add independent Dockerfile

ruflin:
The package storage and the package registry are soon going to be decoupled. To be able to still build a docker container with both inside a Dockerfile is needed that can pull in the dependencies from an external sources instead of local files.

To get this started, a separate Dockerfile is introduced that does not load any files from the repository directly but pulls it from Github. This is to start play around with this idea.

The end goal is being able to pass in the version of the packages that should be used to built the container and the version of the registry.

Where this Dockerfile should land in the end is not clear yet but the package-registry seems to be a good place at the moment.

Other changes:

* Updated the Golang version of the container running the registry

### DIFF
--- a/dev/release/Dockerfile
+++ b/dev/release/Dockerfile
@@ -22,13 +22,27 @@ EXPOSE 8080
 ENV GO111MODULE=on
 RUN go mod vendor
 RUN go get -u github.com/magefile/mage
+# Prepare all the packages to be built
 RUN mage build
 
-# This directory contains the packages at the moment but is only used during the build process
-# If we keep it, it means all packages exist twice.
-RUN rm -rf dev
+# Build binary
+RUN go build .
 
-ENTRYPOINT ["go", "run", "."]
+# Move all files need to run to its own directory
+# This will become useful for staged builds later on
+RUN mkdir /registry
+RUN mv package-registry /registry/
+RUN mv config.yml /registry/
+RUN mv public /registry/
+
+# Clean up files not needed
+RUN rm -rf /go/src/github.com/elastic/package-registry
+RUN rm -rf /go/package-storage
+
+# Change to new working directory
+WORKDIR /registry
+
+ENTRYPOINT ["./package-registry"]
 # Make sure it's accessible from outside the container
 CMD ["--address=0.0.0.0:8080"]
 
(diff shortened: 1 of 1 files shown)


dev/release/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# This Dockerfile allows to build the package-registry and packages together.
```
mtojek, line 1: It's cool that you touched this.

I see a couple of issue with this file. Can we split this Dockerfile into multiple stages (https://docs.docker.com/develop/develop-images/multistage-build/)? There is no need to store source code in the image.

ruflin: I was not even aware multi stage builds exists. I need to look more into this.

ruflin: I had a look at the multi stage builds and this will be exactly what we need. I assume this will also make it obsolete to clean up images from dependencies which were installed through apt-get or similar.

As I'm still iterating on this for now I would like to keep it all together and do the multi stage in a second step if that is ok for you. In any case, it will need quite a bit of optimisations.

dev/release/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ARG GO_VERSION=1.14.2
+FROM golang:${GO_VERSION}
+
+RUN \
+    apt-get update \
+      && apt-get install -y --no-install-recommends \
+         zip rsync \
+      && rm -rf /var/lib/apt/lists/*
+
+RUN git clone https://github.com/elastic/package-storage.git
```
ruflin, line 12: @mtojek I created the package-storage repository and now pull it in. I took some of the existing packages, copied them there with an older version so they are shipped but will not have any effect.

I realised this has an interesting side effect: Assuming we define that all packages in the storage layer must be under `/packages`, this would work with 1 or n repositories. More important, using Github for this is only one option, it could be any storage layer.

mtojek: Are you sure that the order Docker commands is the optimal one? Maybe it's better to build a repository code first. Also, `git clone` would pick up always the latest version of packages, not sure if it isn't better to depend on particular revisions.

ruflin: I doubt the order is optimal. At the moment it is all about getting something working we can iterate on. I hope in the end the container we build, will have a name that lets us identify which package registry version and which version of the packages was built so we could also reproduce it outside a docker container.

mtojek: Got it, thanks!

ruflin:
@mtojek 100% agree. I filed https://github.com/elastic/package-registry/issues/371 to make sure we track it.

PR #370, opened 2020-04-22
Add Cisco integration 0.0.2

alakahakai:
Cisco integration 0.0.2

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,10 +184,11 @@ what's been already fixed, as the script has overridden part of it).
 
     The goal of this action item is to verify if produced artifacts are correct.
 
-    The fields files (`package-fields.yml`, `fields.yml` and `ecs.yml`) in the package were created from original
-    `fields.yml` files and the ECS schema. It may happen that original sources have a typo, bad description or misses
-    a field definition. The sum of fields in all present files should contain only fields that are really used, e.g.
-    not all existing ECS fields.
+    The fields files (package-fields.yml, fields.yml and ecs.yml) in the package were created from original fields.yml
+    files (that may contain ECS schema fields) and fields.epr.yml (defining some other fields used in the ingest
+    pipeline). It may happen that original sources have a typo, bad description or misses a field definition.
+    The sum of fields in all present files should contain only fields that are really used, e.g. not all existing ECS
+    fields.
 
     It may happen that the ingest pipeline uses fields abstracted from ECS, but not mentioned in `fields.yml`.
     Integrations should contain these fields and also have them documented.


dev/import-beats-resources/cisco/docs/README.md
```diff
@@ -0,0 +1,29 @@
+# Cisco Integration
+
+This integration is for Cisco network device's logs. It includes the following
+filesets for receiving logs over syslog or read from a file:
```
mtojek, line 4: When talking about integrations, we shouldn't use words like "filesets" or "metricsets" as they don't exist anymore. It's safer to replace these terms with "datasets".

dev/packages/alpha/cisco-0.0.2/dataset/asa/agent/stream/stream.yml
```diff
@@ -0,0 +1,15 @@
+{{#if input == "syslog"}}
```
mtojek, line 1: This line is not correct in the handlebars syntax and needs to split in multiple stream files.

Unfortunately this is a known issue, but it requires some advanced templating processing - we didn't introduce it.

dev/packages/alpha/cisco-0.0.2/dataset/asa/elasticsearch/ingest-pipeline/default.yml
```diff
@@ -0,0 +1,1281 @@
+---
```
mtojek, line 1: If this file contains any fields not mentioned in the fields files (e.g. event.code, event.provider), please create in Beats a file called `fields.epr.yml` and describe them there (there is a paragraph about this in CONTRIBUTING).

dev/packages/alpha/cisco-0.0.2/dataset/asa/elasticsearch/ingest-pipeline/default.yml
```diff
@@ -0,0 +1,1281 @@
...
+description: "Pipeline for Cisco {< .internal_PREFIX >} logs"
```
mtojek, line 2: @ruflin not sure what to do with such entries

dev/packages/alpha/cisco-0.0.2/dataset/asa/elasticsearch/ingest-pipeline/default.yml
```diff
@@ -0,0 +1,1281 @@
...
+
+  #
+  # Rename some 7.x fields
+  #
+  - rename:
+      field: log.original
+      target_field: event.original
+      ignore_missing: true
+  - rename:
+      field: cisco.{< .internal_prefix >}.list_id
```
mtojek, line 1274: @ruflin not sure what to do with such entries.

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
```diff
@@ -0,0 +1,52 @@
+title: Cisco asa logs
```
mtojek, line 1: I assume that "ASA" should be written in caps. If so, then please adjust all titles and descriptions.

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
```diff
@@ -0,0 +1,52 @@
...
+        required: true
+        show_user: true
+        default: asa
+      - name: internal_PREFIX
+        type: text
+        title: Internal PREFIX
+        multi: false
+        required: true
+        show_user: true
+        default: ASA
```
mtojek, line 50: This is funny :) We have two variables with different spelling (_prefix vs _PREFIX). Is it a wrong configuration definition in Beats and can be somehow compacted?

Also, for a normal user this variable may sound a bit odd. You can add a "description" property and describe what is it all about.

dev/packages/alpha/cisco-0.0.2/manifest.yml
```diff
@@ -0,0 +1,49 @@
...
+    size: 216x216
+    type: image/svg+xml
+datasources:
+  - name: cisco
+    title: Cisco logs
+    description: Collect logs from Cisco instances
+    inputs:
+      - type: logs
+        title: Collect logs from Cisco instances
+        description: Collecting Cisco asa, ftd and ios logs
```
mtojek, line 34: I assume that ASA, FTP and IOS can be written in caps.

mtojek:
@alakahakai Thank you for the contribution. I left a couple of comments that can apply to few place in your contribution, but I didn't mark every single place (e.g. ``{{ if input == ... }}``).

You can also extend the description of the issue with commands you use to import modules or test them manually.

Fingers crossed for serving this package through the registry and I'm happy to see some screenshots in the PR description with the UI.

mtojek:
Also, I converted your PR to draft to prevent an accidental merging. I used to convert them once I finish all necessary works (I mean, "in my opinion" the package is complete, it works in my environment and I see flowing data on dashboards).

dev/packages/alpha/cisco-0.0.2/dataset/asa/agent/stream/stream.yml
ruflin: As https://github.com/elastic/package-registry/pull/346 was merged, we can start to use it in the packages but it will not be supported yet in the UI. I suggest we pick one of the two inputs for now.

@alakahakai Which one do you think is more important? syslog or file?

dev/packages/alpha/cisco-0.0.2/dataset/asa/elasticsearch/ingest-pipeline/default.yml
ruflin: Me neither. How does this work in Beats @adriansr , I see it was introduced in https://github.com/elastic/beats/pull/13286 ? Is this to reuse some ingest pipeline?

For now we should just type it out.

dev/packages/alpha/cisco-0.0.2/dataset/ftd/agent/stream/stream.yml
```diff
@@ -0,0 +1,14 @@
+{{#if input == "syslog"}}
```
ruflin, line 1: Same as in the other dataset, lets split it up an pick one.

dev/packages/alpha/cisco-0.0.2/dataset/ftd/elasticsearch/ingest-pipeline/default.yml
```diff
@@ -0,0 +1,1281 @@
+---
+description: "Pipeline for Cisco {< .internal_PREFIX >} logs"
```
ruflin, line 2: Lets hardcode for now.

dev/packages/alpha/cisco-0.0.2/dataset/ftd/manifest.yml
```diff
@@ -0,0 +1,52 @@
...
+        show_user: true
+        default: 7
+      - name: internal_prefix
+        type: text
+        title: Internal Prefix
+        multi: false
+        required: true
+        show_user: true
+        default: ftd
+      - name: internal_PREFIX
```
ruflin, line 44: This reminds me of the other variable used. I think I need more background on how this should be used?

dev/packages/alpha/cisco-0.0.2/dataset/ios/agent/stream/stream.yml
```diff
@@ -0,0 +1,19 @@
+{{#if input == "syslog"}}
```
ruflin, line 1: Split up as above.

dev/packages/alpha/cisco-0.0.2/docs/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
ruflin, line 1: jep :-D

dev/packages/alpha/cisco-0.0.2/manifest.yml
```diff
@@ -0,0 +1,49 @@
...
+  - src: /img/cisco.svg
+    title: cisco
+    size: 216x216
+    type: image/svg+xml
+datasources:
+  - name: cisco
+    title: Cisco logs
+    description: Collect logs from Cisco instances
+    inputs:
+      - type: logs
```
ruflin, line 32: As I see here just one input and not multiple, I assume logs (file) is the one we should start with? Long term if we support multiple inputs, we need to have here more then one.

dev/packages/alpha/cisco-0.0.2/dataset/asa/elasticsearch/ingest-pipeline/default.yml
adriansr: So this pipeline is shared between cisco/asa and cisco/ftd, because the ftd has to support all the same messages as the asa plus an extra few ones. The `.internal_prefix` being `asa` or `ftd` used to define the final location for fields in a document (`cisco.asa.xxx` or `cisco.ftd.xxx`). `.internal_PREFIX` it's just the upper-case equivalent, `ASA` and `FTD`.

In retrospect this could've been implemented by separating the common part into a different pipeline instead of one big templated pipeline.

ruflin: I doubt we will support reusing pipelines in packages in the near term. Instead I think this kind of logic / magic should happen in the future where the packages are developed (I know, today it is the same place). This will keep the package format and logic in EPM simple.

So for now, lets just hard code it / copy it.

dev/packages/alpha/cisco-0.0.2/dataset/asa/agent/stream/stream.yml
alakahakai: syslog is more important, since file input is used for testing only. For handlebars, can we add the support for Ember Truth Helper addon?

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
alakahakai: @adriansr Can you help to explain the reason of the two variables? thanks

mtojek:
Could you please post some screenshots in the description?

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
```diff
@@ -0,0 +1,59 @@
...
+        type: text
+        title: Tags
+        multi: true
+        required: true
+        show_user: true
+        default:
+          - cisco-asa
+      - name: syslog_port
+        type: integer
+        title: Syslog Port
```
mtojek, line 20: Is there a configuration option for `Syslog Host` or it's only port as it's localhost? Maybe a description would clarify this.

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
```diff
@@ -0,0 +1,59 @@
...
+    title: Cisco ASA logs
+    description: Collect Cisco ASA logs
+    template_path: stream.yml
+    vars:
+      - name: tags
+        type: text
+        title: Tags
+        multi: true
+        required: true
+        show_user: true
```
mtojek, line 15: Is it required to show this field to user?

dev/packages/alpha/cisco-0.0.2/dataset/asa/manifest.yml
```diff
@@ -0,0 +1,59 @@
...
+        default:
+          - cisco-asa
+      - name: log_level
+        type: integer
+        title: Log Level
+        multi: false
+        required: true
+        show_user: true
```
mtojek, line 58: Same here: is it required to present it to user? I think screenshots will help us understand which fields are required.

dev/packages/alpha/cisco-0.0.2/dataset/ios/manifest.yml
```diff
@@ -0,0 +1,53 @@
...
+        default:
+          - cisco-ios
+      - name: syslog_port
+        type: integer
+        title: Syslog Port
+        multi: false
+        required: true
+        show_user: true
+        default: 9002
+  - input: file
```
mtojek, line 33: Did you check if this option works? I wonder if this should be "file" or "log".

PR #374, opened 2020-04-23
testing: Improve kibana healthcheck

mtojek:
This PR improves the Kibana healthcheck to wait for the UI to be ready.

### DIFF
--- a/testing/environments/snapshot.yml
+++ b/testing/environments/snapshot.yml
@@ -22,7 +22,7 @@ services:
   kibana:
     image: docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:5601/login"]
+      test: "curl -f http://localhost:5601/login | grep kbn-injected-metadata 2>&1 > /dev/null"
       retries: 600
       interval: 1s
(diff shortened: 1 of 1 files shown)


testing/environments/snapshot.yml
```diff
@@ -22,7 +22,7 @@ services:
   kibana:
     image: docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:5601/login"]
+      test: "curl -f http://localhost:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null"
```
ruflin, line 25: Maybe you can take some inspiration here? https://github.com/elastic/beats/blob/master/testing/environments/snapshot.yml#L32

mtojek: I changed the healthcheck test because this cluster has auth enabled, so it's not straightforward to reach out to the API. When Kibana is ready, it exposes the login panel,  so no need to fetch the status from an encrypted endpoint.

ruflin: I would assume you could use `elastic:changeme` hardcoded to access it but if the login change works, SGTM.

PR #378, opened 2020-04-23
Use index templates v2 in package 'base'

skh:
* Move the pre-built packages in `base` over to index templates v2
* introduces new directory `component-template`

**Do not merge before https://github.com/elastic/kibana/pull/64305 has been merged, as the current EPM implementation can't handle pre-built v2 packages**.


dev/packages/example/base-1.0.0/elasticsearch/component-template/events-mappings.json
```diff
@@ -0,0 +1,148 @@
+{
```
ruflin, line 1: More general question we had around `ingest-pipeline` directory. Should it be `component_template` or `component-template`? What is our rule? Any preference?

skh: I have no opinion here, I followed the example of `ingest-pipeline`.

(Kibana saved objects use hyphenation and not snake case, maybe that's how the initial format was derived even for assets that are not Kibana assets.)

skh:
> My assumption is that in packages and EPM we ONLY support v2 templates. Is this correct?

Yes. For a smooth developer experience this week the code will be able to install pre-built v1 templates that are in the `base` package before this PR is merged, but other than  that everything will be v2.

dev/packages/example/base-1.0.0/elasticsearch/component-template/events-mappings.json
ruflin: lets keep the `-` for now. The part where I stumbled over `_` is in the ES API.

ruflin:
@skh Conflict in the CHANGELOG. And because of my other PR, probably best rebase ...

dev/packages/example/base-1.0.0/elasticsearch/component-template/events-mappings.json
ruflin, line 1: After a rebase, this should become `base/1.0.0/..` as the path.

skh:
Closed, see https://github.com/elastic/package-registry/pull/392

PR #380, opened 2020-04-24
import-beats: support multiple streams

mtojek:
_Finally managed to add the missing piece. Got really frustrated..._

Changes:

* convert Golang templates into Handlebars (doesn't support all possible tags, but vast majority of the important one)
* replace old `strings.Replace` for template convertion with parsing tree nodes
* consider multiple streams while compacting
* add ".hbs" extension to stream agent templates
* rename existing stream.yml files to stream.yml.hbs


dev/generator/main.go
```diff
@@ -155,7 +155,7 @@ func buildPackage(packagesBasePath string, p util.Package) error {
 			return err
 		}
 
-		err = ioutil.WriteFile(filepath.Join(dirPath, "stream.yml"), []byte(streamFields), 0644)
+		err = ioutil.WriteFile(filepath.Join(dirPath, "stream.yml.hbs"), []byte(streamFields), 0644)
```
ruflin, line 158: @andrewkroh You will like this.
@nchaulet @jen-huang Will this break anything at the moment in the config builder? I hope not.

dev/import-beats/streams_config_parser.go
```diff
@@ -0,0 +1,374 @@
...
+
+	"github.com/elastic/package-registry/util"
+)
+
+type streamConfigParsed struct {
+	tree *parse.Tree
+}
+
+func parseStreamConfig(content []byte) (*streamConfigParsed, error) {
+	mapOfParsed, err := parse.Parse("hello", string(content), "", "", map[string]interface{}{
```
ruflin, line 23: "hello"?

dev/packages/beats/activemq-0.0.1/dataset/audit/agent/stream/log.yml.hbs
```diff
@@ -0,0 +1,6 @@
+type: log
+paths:
+{{#each paths as |path i|}}
```
ruflin, line 3: Is the second part `as |path i|` needed?

dev/packages/beats/cisco-0.0.1/dataset/ios/agent/stream/pipeline.js
```diff
@@ -0,0 +1,222 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+var ciscoIOS = (function() {
```
ruflin, line 5: No need to answer this in this PR but I wonder how this will work. I assume at the moment Filebeat will still ship with this. I assume at the moment this should not be in the package.

dev/packages/beats/coredns-0.0.1/dataset/log/agent/stream/log.yml.hbs
```diff
@@ -0,0 +1,7 @@
+type: log
```
ruflin, line 1: I think this should be `input: log` as based on this the mapping to the input is done? Or did this become obsolete because of the path?

testdata/public/package/multiple-false/0.0.1/index.json
```diff
@@ -19,8 +19,8 @@
     "/package/multiple-false/0.0.1/manifest.yml",
     "/package/multiple-false/0.0.1/docs/README.md",
     "/package/multiple-false/0.0.1/dataset/foo/manifest.yml",
-    "/package/multiple-false/0.0.1/dataset/foo/fields/stream.yml",
-    "/package/multiple-false/0.0.1/dataset/foo/agent/stream/stream.yml",
+    "/package/multiple-false/0.0.1/dataset/foo/fields/stream.yml.hbs",
```
ruflin, line 22: Any chance we could take the `.hbs` part into a separate PR. We might need here some discussions with Kibana team around this and also makes it easier to reference to it later on if questions come up.

+1 on this change BTW.

dev/import-beats/streams_config_parser.go
mtojek: There is only single template, so need to define a specific template name hence this "hello world". I can rename it to something meaningful

dev/packages/beats/activemq-0.0.1/dataset/audit/agent/stream/log.yml.hbs
mtojek: This is tricky actually, because the "i" is always present when unrolling the "range". Wouldn't like to detect this in the code. I will let the developer adjust it later on. Is it fine?

dev/packages/beats/coredns-0.0.1/dataset/log/agent/stream/log.yml.hbs
mtojek: I think you're right... I will adjust the PR.

testdata/public/package/multiple-false/0.0.1/index.json
mtojek: Hmm... honestly I would keep it here as I already applied in multiple places (simpler option at the moment, preserving we will rename to this). Also it's a parsed template, so definetely a .hbs file. WDYT?

dev/packages/beats/activemq-0.0.1/dataset/audit/agent/stream/log.yml.hbs
ruflin: SGTM as long as this works on the Kibana side (I assume it does).

testdata/public/package/multiple-false/0.0.1/index.json
ruflin: Ok, in this case we should add a line to the changelog. And perhaps we can do a Github issue instead to describe the modifivation of the change instead and reference this PR.

dev/import-beats/streams_config_parser.go
ruflin: ah, got it. Something more descriptive would be nice.

dev/packages/beats/coredns-0.0.1/dataset/log/agent/stream/log.yml.hbs
mtojek: Fixed!

dev/import-beats/streams_config_parser.go
mtojek: Fixed.

testdata/public/package/multiple-false/0.0.1/index.json
mtojek: Changelog entry added.

PR #382, opened 2020-04-24
Add experimental releases for packages and datasets

ruflin:
So far the available release types were beta and ga. In this PR experimental is added to the list and set as default.

Having experimental around for packages provides some interesting options. Experimental packages are not shown in the search endpoint by default. To query for the experimental package the flag `experimental=true` has to be added. This allows to publish some packages without directly making it available to everyone.

The idea is that we introduce in EPM a setting "Experimental" that users can enable to also see and install experimental packages.

On the dataset level also experimental datasets exist. My expectation here is that these would not be shown by the Config Builder as long as Experimental:true is not enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add `removable` flag to package manifest. Default is true. [#359](https://github.com/elastic/integrations-registry/pull/359)
 * Add stream template to package json. [#335](https://github.com/elastic/integrations-registry/pull/335)
 * Add support for multiple inputs per dataset. [#346](https://github.com/elastic/integrations-registry/pull/346)
+* Add experimental releases for packages and datasets. [#](https://github.com/elastic/integrations-registry/pull/)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add `removable` flag to package manifest. Default is true. [#359](https://github.com/elastic/integrations-registry/pull/359)
 * Add stream template to package json. [#335](https://github.com/elastic/integrations-registry/pull/335)
 * Add support for multiple inputs per dataset. [#346](https://github.com/elastic/integrations-registry/pull/346)
-* Add experimental releases for packages and datasets. [#](https://github.com/elastic/integrations-registry/pull/)
+* Add experimental releases for packages and datasets. [#382](https://github.com/elastic/integrations-registry/pull/382)
 
 ### Deprecated
 
--- a/categories.go
+++ b/categories.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strconv"
 	"time"
 
 	"github.com/elastic/package-registry/util"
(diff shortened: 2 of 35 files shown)


ruflin:
@mtojek If we go forward with this, my thinking would be that soon we switch the generator to have experimental as default, but first we need the setting in Kibana
@neptunian @nchaulet Would be such a setting for EPM possible?
@jen-huang WDYT about the dataset part that is not shown if experimental?

nchaulet:
For EPM I think we can pass the flag if the user provide it in the query string for example.

> On the dataset level also experimental datasets exist

Do you mean we can have experimental dataset inside a package that is not experimental? 


Also regarding name what do you think of using `alpha` instead of `experimental` I think it sounds more a versioning concern.

mtojek:
> @mtojek If we go forward with this, my thinking would be that soon we switch the generator to have experimental as default, but first we need the setting in Kibana

`import-beats` bases on the release tag in Beats. We can ignore the Beats tag at all and set everywhere "experimental" for every package.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add `removable` flag to package manifest. Default is true. [#359](https://github.com/elastic/integrations-registry/pull/359)
 * Add stream template to package json. [#335](https://github.com/elastic/integrations-registry/pull/335)
 * Add support for multiple inputs per dataset. [#346](https://github.com/elastic/integrations-registry/pull/346)
+* Add experimental releases for packages and datasets. [#](https://github.com/elastic/integrations-registry/pull/)
```
mtojek, line 29: nit: missing PR reference

categories.go
```diff
@@ -28,6 +29,19 @@ func categoriesHandler(packagesBasePath string, cacheTime time.Duration) func(w
 			notFound(w, err)
 			return
 		}
+
+		query := r.URL.Query()
+		var experimental bool
+		// Read query filter params which can affect the output
+		if len(query) > 0 {
```
mtojek, line 36: I think you can extract this block to a separate method.

categories.go
```diff
@@ -37,6 +51,11 @@ func categoriesHandler(packagesBasePath string, cacheTime time.Duration) func(w
 				continue
 			}
 
+			// Skip experimental packages if flag is not specified
+			if p.Release == util.ReleaseExperimental && !experimental {
```
mtojek, line 55: switch the order of comparisons :)

search.go
```diff
@@ -65,6 +66,13 @@ func searchHandler(packagesBasePath string, cacheTime time.Duration) func(w http
 					internal, _ = strconv.ParseBool(v)
 				}
 			}
+
+			if v := query.Get("experimental"); v != "" {
+				if v != "" {
+					// In case of error, keep it false
+					experimental, _ = strconv.ParseBool(v)
```
mtojek, line 73: I'm not fan of swallowing errors... this should reflect in HTTP 400 Bad Request.

ruflin: You are right, if wrong params are submitted, we should just error and not assume defaults like we do here. As this applies to also the other api params, ok if I follow up with it?

categories.go
ruflin: you mean to reuse it for categories and search?

search.go
mtojek: Sounds good!

categories.go
mtojek: Yes and try to adjust the production code to be as concise as possible (SRP, KISS).

ruflin: The query part between search and categories is not identical, the part that could be reused is the check around the flag. I can extract that part if you prefer. 

Side note: I don't consider copying code as a fundamentally bad thing ;-)

mtojek: Don't need to spend time right now, we can refactor it later on.

ruflin: yes, we have to.

ruflin:
I filed https://github.com/elastic/kibana/issues/64869 as a follow up for Kibana.

jfsiii:
I missed this initially, but this reads like there would be 3 or N flags (experimental, beta, ga). Is that correct?

Can multiple values be supplied? e.g. ?experimental=true&ga=true&beta=true 

Since these all correspond to a single release field, I'd expect a single param, like release=experimental. Multiple could be ? release=experimental,ga,beta or ?release=experimental&release=ga&release=beta

ruflin:
Currently it is 3, I could see a 4th one coming which is `deprecated`. 

Lets assume we go with the `release` param. If no query is provided, it would return `beta,ga` by default. Now what happens if `release=experimental` is used, will it only return experimental ones or also beta,ga?

I'm now wondering if both flags can be come useful. `release` as you suggest it is a filter. `experimental=true` as a boolean is a shortcut for `release=experimental,beta,ga`.

PR #385, opened 2020-04-27
import-beats: Use stream.dataset field

mtojek:
Changes:

* use `stream.dataset` in the logic responsible for replacing `event.module` files
* replace all remaining places in which `event.dataset` is used


mtojek:
> Could we add to the migration guide also a note that this must be done / checked? Using `stream.dataset` instead of `event.dataset` also makes queries a lot more efficient as this is a constant_keyword.

There is a check in the script that will fail the migration if detects `event.dataset` field.

ruflin:
I probably mean more "development guide" on my end. I'm not worried about the migration script part but devs building packages after it and using it just out of habit.

PR #388, opened 2020-04-27
import-beats: Do not print table with no fields in README

mtojek:
Only for consistency - if there is a dataset with no fields defined, we shouldn't print empty tables.

This PR adjusts the `import-beats` code for better user experience.


mtojek:
> Did you find a dataset where this is the case. Because a dataset without fields sounds like an incomplete dataset :-)

Yes, indeed! It was AWS cloudwatch package, which is in-progress.

PR #392, opened 2020-04-28
Use index templates v2 in package 'base'

skh:
**Replaces https://github.com/elastic/package-registry/pull/378**

* Move the pre-built packages in `base` over to index templates v2
* introduces new directory `component-template`

**Do not merge before https://github.com/elastic/kibana/pull/64305 has been merged, as the current EPM implementation can't handle pre-built v2 packages**.


ruflin:
I tested this with https://github.com/elastic/kibana/pull/64305 and works as expected.

PR #395, opened 2020-04-28
Update golang dependencies

ruflin:
Update Golang dependencies and update CI Dockerfile to Go 1.14.2


mtojek:
Hm.. I think you forgot about the vendor dir?

ruflin:
I did indeed forget to push the vendor files. I'm glad we have this check in place. Updated.

Issue #375, opened 2020-04-23
Define naming of docker images for packages + registry


ruflin:
The Elastic Package Registry (epr.elastic.co) run a docker container that contains the packages + the registry service itself. Today the packages are in the same repository as the registry but this is changing. Packages will be moving to https://github.com/elastic/package-storage Also there is work ongoing to have a [Dockerfile](https://github.com/elastic/package-registry/blob/master/dev/release/Dockerfile) available that is independent of both repository.

This means the Docker image will contain data from two different sources. For debugging and support purpose I think it is important that we can identify what exactly went into this Docker image to potentially reproduce issues.

My proposal related to naming each Docker image is as following:

`{date}-{package-storage-hash}-{package-registry-hash}`

`hash` in the above can be a commit hash but also a branch or a tag. The `date` part is for human and sorting purpose to identify when exactly an image was built. Two examples of the above would look as following:

* `202004231234-c890763-c1d2c2c`
* `202004231234-c890763-v0.3.0`

The package hash will change much more often than the registry one but it also the more important one for debugging purpose. That is why I think it should come first. 

To build this image, both the `package-storage` hash and `package-registry` hash are passed to the Dockerfile and then used to check out the correct version (https://github.com/elastic/package-registry/blob/master/dev/release/Dockerfile#L19).

Questions
* Is the above possible to build / automate? Does it make sense?
* What is the date/time format we should use?

ruflin:
@kuisathaverat Would be great to get your feedback / input on the above.

kuisathaverat:
>* Is the above possible to build / automate? Does it make sense?

It depends, 😄 What are the uses cases for that tag? I think on the following
* Someone wants to get the latest version
* Someone wants to get a specific version of the package-registry
* Someone wants to get a specific version of the package-registry and specific version of the package-storage

There are some questions about what we will release, or how we will release:
* Would we release several versions of the same package-registry on the same day? probably yes, but would change the package-storage?
* Would we release old versions of the package-registry with new versions of the package-storage?
* What is more important the package-registry version or the package-storage version?

we can have several tags for the same version depending on the use, a date based tag is good for human usage, but again it depends on how we will release the versions it would be the latest version or something else. A tag `{package-storage-hash}-{package-registry-hash}` or {package-registry-hash}-{package-storag-hash} is really good for developers we should release one of those too.

>* What is the date/time format we should use?

[...]

PR #405, opened 2020-04-29
Update release Dockerfile to contain labels

ruflin:
For the relases we discussed in https://github.com/elastic/package-registry/issues/375 that we want to have the revisions for debugging purpose. This adds both and adjust the Dockerfile to rely on git clone instead of go get.

Additional change is to adjust the not logging requests for /healthcheck as otherwise this gets spammed in the logs.

### DIFF
--- a/dev/release/Dockerfile
+++ b/dev/release/Dockerfile
@@ -3,20 +3,35 @@
 ARG GO_VERSION=1.14.2
 FROM golang:${GO_VERSION}
 
+# Get dependencies
 RUN \
     apt-get update \
       && apt-get install -y --no-install-recommends \
          zip rsync \
       && rm -rf /var/lib/apt/lists/*
 
-RUN git clone https://github.com/elastic/package-storage.git
-RUN go get github.com/elastic/package-registry
-WORKDIR "/go/src/github.com/elastic/package-registry"
-RUN rsync -a /go/package-storage/packages/ dev/packages/example
+WORKDIR /home
 
-# Define which version of the packages should be checked out.
-ARG PACKAGES_VERSION=master
-RUN git checkout ${PACKAGES_VERSION}
+# Check out package storage
+RUN git clone https://github.com/elastic/package-storage.git
+WORKDIR /home/package-storage
+ARG PACKAGE_STORAGE_REVISION=master
+RUN git checkout ${PACKAGE_STORAGE_REVISION}
+LABEL package-storage-revision=${PACKAGE_STORAGE_REVISION}
+
+WORKDIR /home
+
+# Checkout out package registry
+RUN git clone https://github.com/elastic/package-registry.git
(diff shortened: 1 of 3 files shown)


dev/release/Dockerfile
```diff
@@ -3,20 +3,35 @@
...
-WORKDIR "/go/src/github.com/elastic/package-registry"
-RUN rsync -a /go/package-storage/packages/ dev/packages/example
+WORKDIR /home/package-storage
+ARG PACKAGE_STORAGE_REVISION=master
+RUN git checkout ${PACKAGE_STORAGE_REVISION}
+LABEL package-storage-revision=${PACKAGE_STORAGE_REVISION}
+
+WORKDIR /home
+
+# Checkout out package registry
```
mtojek, line 24: nit: Checkout out

dev/release/Dockerfile
```diff
@@ -3,20 +3,35 @@
...
+RUN git clone https://github.com/elastic/package-registry.git
+WORKDIR "/home/package-registry"
+ARG PACKAGE_REGISTRY_REVISION=master
+RUN git checkout ${PACKAGE_REGISTRY_REVISION}
+LABEL package-registry-revision=${PACKAGE_REGISTRY_REVISION}
+
+# Get package storage packages into registry
```
mtojek, line 31: nit: `package storage packages` sounds a bit weird

dev/release/Dockerfile
```diff
@@ -3,20 +3,35 @@
...
+# Note: In the future, packages will only be in the storage and not the registry
+RUN rsync -a /home/package-storage/packages/ /home/package-registry/dev/packages/example
```
mtojek, line 33: verbosity?

main.go
```diff
@@ -134,5 +134,9 @@ func loggingMiddleware(next http.Handler) http.Handler {
 
 // logRequest converts a request object into a proper logging event
 func logRequest(r *http.Request) {
+	// Do not log requests to the health endpoint
+	if r.RequestURI == "/health" {
```
mtojek, line 138: Hm... I think you shouldn't strip it. It's valuable to see that the app receives requests.

mtojek:
Did you merge this already?

ruflin:
@mtojek Yes, sorry, merged already as I wanted to test something. I will follow up on your comments.

mtojek:
> @mtojek Yes, sorry, merged already as I wanted to test something. I will follow up on your comments.

No worries, I'm absolutely fine with followups.

main.go
ruflin: My problem here was that I have configured the Docker healthcheck to poll it every second. So this ends up with quite a few entries.

dev/release/Dockerfile
ruflin: you mean too much or too little?

PR #406, opened 2020-04-29
Adjust MySQL package

mtojek:
Changes:
* release package as beta
* Kibana: remove references to Metricbeat, Filebeat, module

### DIFF
--- a/dev/packages/alpha/mysql/0.0.2/dataset/error/manifest.yml
+++ b/dev/packages/alpha/mysql/0.0.2/dataset/error/manifest.yml
@@ -1,5 +1,4 @@
 title: MySQL error logs
-release: ga
 type: logs
 streams:
 - input: logs
--- a/dev/packages/alpha/mysql/0.0.2/dataset/error/manifest.yml
+++ b/dev/packages/alpha/mysql/0.0.2/dataset/error/manifest.yml
@@ -1,5 +1,6 @@
 title: MySQL error logs
 type: logs
+release: beta
 streams:
 - input: logs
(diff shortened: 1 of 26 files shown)


dev/packages/alpha/mysql/0.0.2/dataset/error/manifest.yml
```diff
@@ -1,5 +1,4 @@
 title: MySQL error logs
-release: ga
```
ruflin, line 2: I think our default is experimental. So if you remove it completely, it will be experimental by default.

ruflin: Hm, thought about this again. We don't have this feature yet on the Kibana side so it will show them.

mtojek: Ah, right. Fixed.

mtojek: Nevertheless I will leave them marked as "beta".

dev/packages/alpha/mysql/0.0.2/manifest.yml
```diff
@@ -8,7 +8,7 @@ type: integration
 categories:
 - logs
 - metrics
-release: ga
+release: beta
```
ruflin, line 11: I nice feature of having these as beta and the automatically imported packages as experimental is that currently the experimental ones don't show up in Kibana.

PR #411, opened 2020-04-29
Removing release Dockerfile

ruflin:
This file fits better into the package-storage repository: https://github.com/elastic/package-storage/blob/master/dev/Dockerfile


ruflin:
@mtojek Cleanup of the file will happen in the package-storage repository. Did not forget about it ;-)

PR #413, opened 2020-04-29
Update packages to remove input definitions

ruflin:
The stream template should not need the input definitions as they are already availabe in the manifest.

### DIFF
--- a/dev/packages/alpha/mysql/0.0.2/dataset/error/agent/stream/stream.yml.hbs
+++ b/dev/packages/alpha/mysql/0.0.2/dataset/error/agent/stream/stream.yml.hbs
@@ -1,4 +1,3 @@
-input: log
 paths:
 {{#each paths}}
   - {{this}}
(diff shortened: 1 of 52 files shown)


ruflin:
@mtojek Any chance you could test this with your nginx setup to see if things still work as expected?

mtojek:
> @mtojek Any chance you could test this with your nginx setup to see if things still work as expected?

I can do it.

mtojek:
@ruflin 

I've modified the `stream.yml` file in the `nginx.access` file to remove the `input: log` from packages and it seems to be working (I hope no layer cached the previous configuration :D )

action_store:
```
      - dataset: nginx.access
        enabled: true
        exclude_files:
        - .gz$
        id: logs-nginx.access
        paths:
        - /var/log/nginx/access.log*
        processors:
        - add_locale: null
```

1. Does it mean that the value in `stream.yml` file is ignored or redundant?

2. Could you please adjust all other files in `alpha`?

ruflin:
The value in the stream.yml is ignored and has not meaning. I think I initially introduced it (with a typo 🤦 ) to match templates to inputs, but I think the template_path is a much better option.

@mtojek Could you take over the updating / cleaning up of all the files?
@alakahakai FYI, the above also effects your PR's.

ruflin:
@mtojek Never mind, I managed to update all of them. Find / Replace to the rescue :-)

PR #420, opened 2020-05-04
Ensure @timestamp is part of each dataset

ruflin:
The @timestamp field is required for all templates. Because of this it is enforced during the build of a package and added in combination with the stream fields.

The `stream.yml` field was renamed to `base-fields.yml` to also remove confusion around the agent stream.yml files.


mtojek:
Linter failed in the CI.

ruflin:
@mtojek sorry about that, I must get better at running the test locally. Generated files now fixed.

Issue #404, opened 2020-04-29
Validate ingest pipelines in YAML format


skh:
EPM installs ingest pipelines in YAML format as they come with the package. Elasticsearch only accepts ingest pipelines that start with `---` on a separate line. Therefore, pipelines that can be installed with `libbeat` can't be installed with EPM. This leads to issues like https://github.com/elastic/kibana/issues/63726

This task is to add validation for YAML-format pipelines so all problems can be detected when packages are built.

mtojek:
@skh @ruflin 

I assume that the validation should be present on both sides. Does Kibana fail with HTTP 400 with a descriptive message for invalid YAML files?

skh:
> Does Kibana fail with HTTP 400 with a descriptive message for invalid YAML files?

Not at the moment, but I've added https://github.com/elastic/kibana/issues/64739.

mtojek:
I started looking into this issue and identified a couple of placeholders that need a default value.

Re `{< IngestPipeline "pipeline-plaintext" >}`:

@skh What would be the pipeline name used by the Ingest Manager? In beats it was: `filebeat-5.2.0-nginx-access-pipeline`

Re...
 ```
- rename:
      field: cisco.{< .internal_prefix >}.list_id
      target_field: cisco.{< .internal_prefix >}.rule_name
      ignore_missing: true
```

... and...
```
 - drop:
      if: "ctx.event.severity > {< .log_level >}"
```
To let validation pass, I can temporarily use a "fix-me" value.

Re:
```
    {<if .use_namespace >}
    {
      "script": {
        "lang": "painless",
        "source": "def dict = ['result': new HashMap()]; for (entry in ctx['json'].entrySet()) { dict['result'][entry.getKey()] = entry.getValue(); } ctx['osquery'] = dict; ctx.remove('json');"
      }
    },
    {< end >}
```

I can temporarily remove the condition.

ruflin:
I think the only templating we should support in the pipelines is referencing an other pipeline. Everything else should not be supported anymore. I assume `{< .log_level >}` is something that Beats replaces today? Or how does this work?

skh:
We currently don't rename index pipelines during package installation, so when we find `{< IngestPipeline "pipeline-plaintext" >}` we assume that there is a `pipeline-json.{json,yml}` pipeline in the same directory and insert a reference to it. We use the filename of the pipeline in the package as the pipeline name, minus the extension.

[...]

PR #423, opened 2020-05-04
import-beats: Validate ingest pipelines

mtojek:
This PR extends the `import-beats` script with support for validation of ingest pipelines.


dev/packages/beats/cisco/0.0.1/dataset/asa/elasticsearch/ingest-pipeline/default.yml
```diff
@@ -1,5 +1,5 @@
 ---
-description: "Pipeline for Cisco {< .internal_PREFIX >} logs"
+description: "Pipeline for Cisco FIX_ME logs"
```
ruflin, line 2: NICE

dev/packages/beats/osquery/0.0.1/dataset/result/elasticsearch/ingest-pipeline/default.json
```diff
@@ -14,14 +14,14 @@
...
-    {<if .use_namespace >}
+    
     {
       "script": {
         "lang": "painless",
         "source": "def dict = ['result': new HashMap()]; for (entry in ctx['json'].entrySet()) { dict['result'][entry.getKey()] = entry.getValue(); } ctx['osquery'] = dict; ctx.remove('json');"
       }
     },
-    {< end >}
+
```
ruflin, line 24: It is valid json with these empty lines?

mtojek: It passes the self-validation :)

ruflin:
@mtojek I filed https://github.com/elastic/package-registry/issues/404 as an indirect follow up on this issue as we should also validate it during packaging.

mtojek:
> @mtojek I filed #404 as an indirect follow up on this issue as we should also validate it during packaging.

Will try to add it in a sec.

ruflin:
@mtojek separate PR please ;-)

PR #425, opened 2020-05-04
Refactor package-registry

mtojek:
Changes:
- remove redundant os.Exit after log.Fatal
- check errors on generator commands
- use Go http logic for serving content: prevent any path traversal bugs, detect content-type automatically
- use http.FileServer for package content (support for If-Modified-Since, range queries)

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 
 	"github.com/magefile/mage/sh"
+	"github.com/pkg/errors"
 
 	"github.com/elastic/package-registry/util"
 )
@@ -59,17 +60,14 @@ func main() {
 
 	if sourceDir == "" || publicDir == "" {
 		log.Fatal("sourceDir and publicDir must be set")
-		os.Exit(1)
 	}
 
 	if err := Build(sourceDir, publicDir); err != nil {
 		log.Fatal(err)
-		os.Exit(1)
 	}
 }
 
 func Build(sourceDir, publicDir string) error {
-
 	err := BuildPackages(sourceDir, filepath.Join(publicDir, packageDirName))
 	if err != nil {
 		return err
@@ -80,8 +78,11 @@ func Build(sourceDir, publicDir string) error {
 // CopyPackage copies the files of a package to the public directory
 func CopyPackage(src, dst string) error {
 	log.Println(">> Copy package: " + src)
-	os.MkdirAll(dst, 0755)
-	err := sh.RunV("rsync", "-a", src, dst)
+	err := os.MkdirAll(dst, 0755)
+	if err != nil {
+		return err
+	}
+	err = sh.RunV("rsync", "-a", src, dst)
 	if err != nil {
 		return err
 	}
@@ -91,7 +92,6 @@ func CopyPackage(src, dst string) error {
 
 // BuildPackage rebuilds the zip files inside packages
 func BuildPackages(sourceDir, packagesPath string) error {
-
 	list, err := filepath.Glob(sourceDir + "/*/*")
 	if err != nil {
 		return err
@@ -135,7 +135,6 @@ func BuildPackages(sourceDir, packagesPath string) error {
 }
 
 func buildPackage(packagesBasePath string, p util.Package) error {
-
 	// Change path to simplify tar command
 	currentPath, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


handler.go
```diff
@@ -5,107 +5,80 @@
...
+		defer f.Close()
 
-func sendHeader(w http.ResponseWriter, r *http.Request) {
-	extension := filepath.Ext(r.RequestURI)
+		stat, err := f.Stat()
+		if err != nil {
+			return "", errors.New("404 Page Not Found Error")
+		}
 
-	switch extension {
```
ruflin, line 80: How are these headers managed with `http.FileSystem`?

dev/generator/main.go
```diff
@@ -221,25 +220,26 @@ func buildPackage(packagesBasePath string, p util.Package) error {
...
+		dst := filepath.Join(tarGzDirPath, tarGzName)
 
 		// As the package directories are now {packagename}/{version} when just running tar, the dir inside
 		// the package had the wrong name. Using `-s` or `--transform` for some reason worked on the command line
 		// but not when run through Golang. So the hack for now is to just copy over all files with the correct name
 		// and then run tar on it.
 		// This could become even useful in the future as things like images or videos should potentially not be part of
 		// a tar.gz to keep it small.
-		err := CopyPackage(packagesBasePath+"/"+p.Name+"/"+p.Version+"/", copiedPackagePath)
+		src := packagesBasePath + "/" + p.Name + "/" + p.Version + "/"
```
ruflin, line 231:
```suggestion
		src := filepath.Join(packagesBasePath, p.Name, p.Version) + "/"
```

I think we have to leave in the last `/`?

handler.go
mtojek: https://golang.org/src/net/http/sniff.go?s=646:688#L11

ruflin: Did you by chance check if we still ship the same ones as before or if there is a diff?

mtojek: I randomly checked SVG and TGZ, and it looks correct.

ruflin: Could you check all of them just to make sure if we have changes, we know about it? Unfortunately we don't have any tests for this yet.

What does it do for the ones where we set `charset=UTF-8`?

mtojek: Let me add some unit tests first and then we can come back to the discussion.

mtojek: @ruflin 
I've just added a unit test that verifies content-types based on the automatic detection. I assume it bases on some RFC. There are slight differences, but to my mind it's not fine to leave it as is, unless we need the information about content-type for further processing.

main_test.go
```diff
@@ -99,3 +100,39 @@ func runEndpoint(t *testing.T, endpoint, path, file string, handler func(w http.
...
+		expectedContentType string
+	}{
+		{"/index.json", "application/json"},
+		{"/activemq-0.0.1.tar.gz", "application/x-gzip"},
+		{"/favicon.ico", "image/x-icon"},
+		{"/metricbeat-mysql.png", "image/png"},
+		{"/kibana-coredns.jpg", "image/jpeg"},
+		{"/README.md", "text/plain; charset=utf-8"},
+		{"/logo_mysql.svg", "image/svg+xml"},
+		{"/manifest.yml", "text/plain; charset=utf-8"},
```
ruflin, line 116: @jfsiii Will this have any effect on the Kibana side?

handler.go
ruflin: I'm not too worried about the content type changes as my assumption is that the ones that are not the same, should have been adjusted anyways. My only concern would be, that we break something in Kibana (I assume not).

mtojek:
Content-types are different for different operating systems (Darwin vs Linux). I had to look deeper for the root cause, but temporarily we can run tests only for the Linux platform.

mtojek:
Found it (/etc/mime.types):

```go
	// If Content-Type isn't set, use the file's extension to find it, but
	// if the Content-Type is unset explicitly, do not sniff the type.
	ctypes, haveType := w.Header()["Content-Type"]
	var ctype string
	if !haveType {
		ctype = mime.TypeByExtension(filepath.Ext(name))
		if ctype == "" {
			// read a chunk to decide between utf-8 text and binary
			var buf [sniffLen]byte
			n, _ := io.ReadFull(content, buf[:])
			ctype = DetectContentType(buf[:n])
			_, err := content.Seek(0, io.SeekStart) // rewind to output whole file
			if err != nil {
				Error(w, "seeker can't seek", StatusInternalServerError)
				return
			}
		}
		w.Header().Set("Content-Type", ctype)
	} else if len(ctypes) > 0 {
		ctype = ctypes[0]
	}
```

mime_types.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"log"
+	"mime"
+)
+
+func init() {
+	mustAddMimeExtenstionType(".gz", "application/gzip")
```
ruflin, line 9: LGTM. Would be nice to have a note in here explaining why we define it here with all the things you were digging up.

PR #426, opened 2020-05-04
Validate ingest pipelines while packaging

mtojek:
Changes:
* validate ingest pipelines while packaging

Issue: https://github.com/elastic/package-registry/issues/404

### DIFF
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -166,9 +166,8 @@ func (d *DataSet) Validate() error {
 			err = validateIngestPipelineFile(jsonPipelinePath)
 			if err != nil {
 				return errors.Wrapf(err, "validating ingest pipeline JSON file failed (path: %s)", jsonPipelinePath)
-			} else {
-				validFound = true
 			}
+			validFound = true
 		}
 
 		yamlPipelinePath := filepath.Join(pipelineDir, d.IngestPipeline+".yml")
@@ -180,9 +179,8 @@ func (d *DataSet) Validate() error {
 			err = validateIngestPipelineFile(yamlPipelinePath)
 			if err != nil {
 				return errors.Wrapf(err, "validating ingest pipeline YAML file failed (path: %s)", jsonPipelinePath)
-			} else {
-				validFound = true
 			}
+			validFound = true
 		}
 
 		if !validFound {
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -161,7 +161,8 @@ func (d *DataSet) Validate() error {
 		_, errJSON := os.Stat(jsonPipelinePath)
 		if errJSON != nil && !os.IsNotExist(errJSON) {
 			return errors.Wrapf(errJSON, "stat ingest pipeline JSON file failed (path: %s)", jsonPipelinePath)
-		} else if !os.IsNotExist(errJSON) {
+		}
+		if !os.IsNotExist(errJSON) {
 			err = validateIngestPipelineFile(jsonPipelinePath)
 			if err != nil {
 				return errors.Wrapf(err, "validating ingest pipeline JSON file failed (path: %s)", jsonPipelinePath)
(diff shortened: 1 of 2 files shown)


util/dataset.go
```diff
@@ -147,17 +150,55 @@ func (d *DataSet) Validate() error {
...
+		jsonPipelinePath := filepath.Join(pipelineDir, d.IngestPipeline+".json")
+		_, errJSON := os.Stat(jsonPipelinePath)
 
-		if os.IsNotExist(errYAML) && os.IsNotExist(errJSON) {
-			return fmt.Errorf("Defined ingest_pipeline does not exist: %s", pipelineDir+d.IngestPipeline)
+		yamlPipelinePath := filepath.Join(pipelineDir, d.IngestPipeline+".yml")
+		_, errYAML := os.Stat(yamlPipelinePath)
+
+		var validFound bool
+		if !os.IsNotExist(errJSON) {
```
ruflin, line 165: Nit: Why not put the checks for each error directly after where the error happened?

util/dataset.go
```diff
@@ -147,17 +150,55 @@ func (d *DataSet) Validate() error {
...
 	}
 	return nil
 }
+
+func validateIngestPipelineFile(pipelinePath string) error {
+	f, err := ioutil.ReadFile(pipelinePath)
+	if err != nil {
+		return errors.Wrapf(err, "reading ingest pipeline file failed")
+	}
+	ext := pipelinePath[strings.LastIndex(pipelinePath, ".")+1:]
```
ruflin, line 192: `filepath.Ext`?

util/dataset.go
```diff
@@ -147,17 +150,55 @@ func (d *DataSet) Validate() error {
...
+			return fmt.Errorf("defined ingest_pipeline does not exist: %s", pipelineDir+d.IngestPipeline)
 		}
 	}
 	return nil
 }
+
+func validateIngestPipelineFile(pipelinePath string) error {
+	f, err := ioutil.ReadFile(pipelinePath)
+	if err != nil {
+		return errors.Wrapf(err, "reading ingest pipeline file failed")
```
ruflin, line 190: It would be nice to have the path reported as part of the error which makes it much easier to find which file did not parse.

util/dataset.go
mtojek: Fixed

util/dataset.go
mtojek: Thanks, fixed

util/dataset.go
mtojek: Fixed

util/dataset.go
```diff
@@ -147,17 +150,61 @@ func (d *DataSet) Validate() error {
...
 	if d.IngestPipeline != "" {
-		_, errJSON := os.Stat(filepath.Join(pipelineDir, d.IngestPipeline+".json"))
-		_, errYAML := os.Stat(filepath.Join(pipelineDir, d.IngestPipeline+".yml"))
+		var validFound bool
+
+		jsonPipelinePath := filepath.Join(pipelineDir, d.IngestPipeline+".json")
+		_, errJSON := os.Stat(jsonPipelinePath)
+		if errJSON != nil && !os.IsNotExist(errJSON) {
+			return errors.Wrapf(errJSON, "stat ingest pipeline JSON file failed (path: %s)", jsonPipelinePath)
+		} else if !os.IsNotExist(errJSON) {
```
ruflin, line 164: I don't think you need an else here as if it matches before, it will return.

util/dataset.go
```diff
@@ -147,17 +150,61 @@ func (d *DataSet) Validate() error {
...
+			err = validateIngestPipelineFile(jsonPipelinePath)
+			if err != nil {
+				return errors.Wrapf(err, "validating ingest pipeline JSON file failed (path: %s)", jsonPipelinePath)
+			} else {
```
ruflin, line 168: no else needed as it returns on 167

util/dataset.go
mtojek: Fixed

util/dataset.go
mtojek: Fixed

mtojek:
> Changelog still missing ;-)

My bad, forgot about it. Sorry!

util/dataset.go
ruflin: is it? :-D Perhaps I'm commenting too early ;-)

mtojek: I was looking on the other if-else struct. Now, it's truly fixed :D

PR #428, opened 2020-05-05
Add kafka integration 0.0.2

jsoriano:
Add kafka packages.

One particularity of the Metricbeat module for Kafka is that it monitors Kafka brokers, producers and consumers. Kafka brokers are monitored using the kafka protocol, and Jolokia for JMX metrics. Producers and consumers and monitored using Jolokia.

Related to these particularities, I have divided the kafka metricsets into three groups after [these thoughts](https://github.com/elastic/package-registry/pull/428#discussion_r424038774):
* `kafka`, to monitor the Kafka brokers, it uses the Filebeat module, and the `consumergroup`, `broker` and `partition` metricsets of the Metricbeat module.
* `kafka_java_consumer`, to monitor Kafka Java consumers, it uses the `consumer` metricset of the Kafka module.
* `kafka_java_producer`, to monitor Kafka Java producers, it uses the `producer` metricset of the Kafka module.

This PR includes only a package for `kafka`, we will have to discuss if we also want `kafka_java_consumer` and `kafka_java_producer` as packages or as data sources, or following any other strategy.

Some screenshots:

If we include `kafka`, `kafka_java_consumer`, and `kafka_java_producer` as three different packages, they appear like this (**this PR only includes the `kafka` one**):
Captura de pantalla de 2020-05-15 16-17-08

Integration for Kafka brokers. The configuration for the `broker` metricset has a setting for the jolokia endpoint:
Captura de pantalla de 2020-05-15 16-17-49

Kafka logs can be configured as files inside the Kafka home directory, following the config of the current Filebeat module:
Captura de pantalla de 2020-05-15 16-19-47


.gitignore
```diff
@@ -1,6 +1,7 @@
-.DS_Store
 package-registry
-
-.idea
 build
 public/*
+
+.DS_Store
```
ruflin, line 5: Can you move this change to a separate PR?

dev/packages/alpha/kafka/0.0.2/dataset/log/agent/stream/log.yml.hbs
```diff
@@ -0,0 +1,11 @@
+paths:
+{{#each paths as |path i|}}
+ - {{kafka_home}}{{path}}
```
jsoriano, line 3: `{{kafka_home}}` here is always being replaced as an empty string, this variable is defined here: https://github.com/elastic/package-registry/pull/428/files#diff-8ddc28ec4e204b300774321f2349a083R7-R12
@mtojek @ruflin, any idea where can be the problem?

dev/packages/alpha/kafka/0.0.2/manifest.yml
```diff
@@ -0,0 +1,77 @@
...
+        default:
+          - 'localhost:9092'
+      - name: jolokia_hosts
+        type: text
+        title: Address of Jolokia agents installed in Kafka (required for some metricsets)
+        multi: true
+        required: false
+        show_user: true
+        default:
+          - 'localhost:8778'
```
jsoriano, line 60: This module has some metricsets that need to connect to Kafka, and some others that need to connect to Jolokia. I am defining two "hosts" variables, one for each kind of metricset.

dev/packages/alpha/kafka/0.0.2/dataset/log/agent/stream/log.yml.hbs
ruflin: I just played around with handlebars and to get this working it must be:
```
- {{../kafka_home}}{{path}}
```

jsoriano: Interesting, thanks! I will give it a try.

dev/packages/alpha/kafka/0.0.2/manifest.yml
ruflin: Are there any metricsets that need to connect to both? If not, I think we should specify two different inputs instead.

jsoriano: Oh, right, makes sense, I didn't think on this possibility, I will give a try, thanks!

jsoriano: How can I define two different inputs for the same module? I have tried with `input: kafka_jolokia/metrics`, but then it configures metricbeat with `module: kafka_jolokia`, that doesn't exist. And with `input: kafka/jolokia_metrics` it doesn't create any configuration.

mtojek: I think you should leave the main manifest.yml as is (2 inputs: logs and kafka/metrics) and go deeper to datasets. `agent/stream/stream.yml.hbs` is the configuration of metricbeat that is delivered to the agent. If you want to take a look at integration with 2 stream configurations, you can take a look at "o365".

ruflin: I think the problem here is that on the Metricbeat side we have 1 module using 2 different inputs. We will face the same issue in Kubernetes. Until we have support for inputs in Metricbeat I think we must add some logic to the agent to do the correct routing.

@ph @michalpristas In summary, the Metricbeat Kafka module uses two different inputs: One is Kafka, the other one is Jolokia but both of the code is inside the Kafka module. The config file here makes it more obvious: https://github.com/elastic/beats/blob/master/metricbeat/module/kafka/_meta/config.yml (see the different hosts configs). We have the same challenge with Kubernetes: https://github.com/elastic/beats/blob/master/metricbeat/module/kubernetes/_meta/config.yml

@jsoriano To unblock you what I suggest for now is to just have all input configs inside the dataset instead of the package manifest. It means the user has to change the host info in multiple places but at least we have it working until Agent supports multiple inputs per module. An other workaround could be to support named inputs. So the same input can be specified multiple times on the package level but a name is given, so the dataset will match the correct one 🤔

jsoriano: > To unblock you what I suggest for now is to just have all input configs inside the dataset instead of the package manifest. It means the user has to change the host info in multiple places but at least we have it working until Agent supports multiple inputs per module.

Well, my previous option of having two hosts with a single metrics input in the package manifest was also working, I could revert to this one. The good thing of this option is that hosts have to be configured in a single place, the bad thing is that it is not obvious what stream uses each host, but it could be added as part of the description.

ruflin: I forgot that we can two different names for the config even though it is called the same in the end. Nice feature :-) +1 on what you had initially for now.

When you think of monitoring Kafka, you think of it of monitoring two different "inputs / parts" or it just one thing you want to configure in one input?

jsoriano: > When you think of monitoring Kafka, you think of it of monitoring two different "inputs / parts" or it just one thing you want to configure in one input?

Yeah, I am thinking that maybe we could separate this into multiple integrations, we have now these metricsets:
* `partition` collects metrics from the kafka broker, using kafka protocol
* `consumergroup` collects metrics from the kafka broker, using kafka protocol
* `broker` collects metrics from the kafka broker, using jolokia
* `consumer` collects metrics from kafka consumers (implemented in java), using jolokia
* `producer` collects metrics from kafka consumers (implemented in java), using jolokia

I think that most of the cases they are going to be used separatedly:
* `partition`, `consumergroup` and optionally `broker` to monitor brokers
* `consumer` to monitor consumers
* `producer` to monitor producers

Would it make sense to separate this into three integrations? I think it could be more clear. We would need in any case a separated setting for the jolokia host of `broker`, but it could be defined only in this dataset.

jsoriano: I have pushed the change for the approach with three integrations (not very tested yet).

ruflin: Sorry for the late answer. When you say 3 integrations, you mean 3 inputs? Could you share a screenshot on how it looks now?

jsoriano: Sorry, I think I mean three packages. Screenshots are updated.

jsoriano: Umm, is it possible to have three different datasources in the same package? Would it work the same as the three packages I am adding now?

jsoriano: We had a discussion offline about this, we have decided to go on by now only with the metricsets intended to monitor brokers, so I will remove the consumer and producer metricsets from this PR.

dev/import-beats-resources/kafka/docs/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+This integration collects logs and metrics from [https://kafka.apache.org](Kafka) servers.
+
+## Compatibility
+
+The `log` dataset module is tested with logs from Kafka 0.9, 1.1.0 and 2.0.0.
+
+The `broker`, `consumergroup` and `partition` metricsets are tested with Kafka 0.10.2.1, 1.1.0, 2.1.1, and 2.2.2.
+
+<!-- TODO: Add a link to Jolokia "input" in Metricbeat -->
```
ruflin, line 11: This shows up in the UI, but can also be fixed later.

mtojek:
> @mtojek If this is merged, could you take over the "transition" to package-registry/integrations and cleanup here?

Yes, once it's merged.

dev/import-beats-resources/kafka/docs/README.md
jsoriano: I have fixed this and a couple of typos.

Issue #414, opened 2020-04-29
Separate "import-beats" script from package-registry code


mtojek:
Ideally the https://github.com/elastic/package-registry repository should contain only the registry code.

The first step to come closer to the final result is moving away the `import-beats` script.

Action items:
- [x]  Copy `dev/import-beats` script to the `integrations` repo
- [x] update `import-beats` docs: README and CONTRIBUTING
- [x]  Copy `testing/environment` script to the integrations and make it working (package registry in different repo)
- [x] Add CI to integrations to still validate packages (copy from package-storage) - https://github.com/elastic/package-registry/issues/414
- [x] Adjust dev build / container build to potentially also include the packages from the integrations repo - https://github.com/elastic/package-registry/pull/435
- [x] Remove copied artifacts from package-registry (remove once everything works as intended)

mtojek:
Alright, I'm going with this one (wouldn't like to start before the long weekend).

mtojek:
@kuisathaverat shall I ask you for enabling CI support on the Jenkins side?

PR #433, opened 2020-05-05
Remove dev beats

mtojek:
Issue: https://github.com/elastic/package-registry/issues/414

Changes:
* remove dev beats totally as they've been migrated to the new place
* leave a deprecation notice


mtojek:
@ruflin are we going to remove the import-beats logic now or later?

PR #434, opened 2020-05-05
Clean import-beats leftovers

mtojek:
Issue: https://github.com/elastic/package-registry/issues/414

This PR cleans the leftovers of the import-beats script.


magefile.go
```diff
@@ -68,18 +68,6 @@ func Build() error {
 	return sh.Run("go", "build", ".")
 }
 
-func ImportBeats() error {
```
ruflin, line 71: Line 35 also needs some cleanup.

mtojek: Fixed.

magefile.go
```diff
@@ -32,7 +32,7 @@ var (
 
 	publicDir    = "./public"
 	buildDir     = "./build"
-	packagePaths = []string{"./dev/packages/alpha/", "./dev/packages/example/", "./dev/packages/beats/"}
```
ruflin, line 35: Why did you remove the example ones too? I would keep them in as default but I opened https://github.com/elastic/package-registry/pull/435 to make it more flexible.

mtojek: Fixed.

PR #435, opened 2020-05-05
Introduce PACKAGE_PATHS env variable select paths to build

ruflin:
The `PACKAGE_PATHS` environment variables allows to overwrite the package paths which are set by default. This becomes useful for example in a Docker container if only one of the directories should be built.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -38,6 +38,11 @@ var (
 
 func Build() error {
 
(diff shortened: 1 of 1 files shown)


ruflin:
@mtojek I think it will not fully solve the adjust deb build part but be a step in the direction to solve it.

PR #437, opened 2020-05-05
Add AWS Package

kaiyan-sheng:
This PR is to add AWS package by running `PACKAGES=aws mage ImportBeats` in elastic/integrations repo and make manual changes afterwards.

Following guide here: https://github.com/elastic/integrations/blob/master/CONTRIBUTING.md

## screenshots
Updated:
updated-aws-package (1)


kaiyan-sheng:
Question: where can I change to remove this `Namespace` from advanced? `Namespace` should be a config only under cloudwatch metric dataset under `metrics`.
Screen Shot 2020-05-06 at 3 12 50 PM

ruflin:
@kaiyan-sheng You can't as these are two different namespaces. This is the namespace for the index. We should add some docs around it. I will file an issue in Kibana.

ruflin:
Created https://github.com/elastic/kibana/issues/65615#issuecomment-625088117

kaiyan-sheng:
@ruflin Good to know! Thank you for creating the issue!

kaiyan-sheng:
Dataset for aws logs are looking good:
Screen Shot 2020-05-07 at 10 04 54 AM

kaiyan-sheng:
Dataset for aws metrics works too now in Kibana without any yml format input parameter:
Screen Shot 2020-05-07 at 10 44 10 AM

kaiyan-sheng:
Dataset for aws metrics works with yml format input parameters now (metrics and tags_filter):
Screen Shot 2020-05-07 at 12 35 51 PM

mtojek:
Great, that you managed to overcome all bugs!

Here are some of my observations:
1. Spelling: use AWS, EC2, CloudTrail, CloudWatch, S3 etc.
2. AWS ec2 logs (s3). I think if there is only one input, there is no need to define it in braces (I know the script works this way, sorry about that).
3. Not sure about defining „regions” as a different setting for every dataset? WDYT @ruflin
4. Can we add to checkbox descriptions links to original AWS pages? WDYT @ruflin 
5. I think we need to select a subset of datasets enabled by default. I suggest 5 biggest/most popular AWS services.
6. aws cloudwatch metrics - Metrics field: it contains values that will not work an ordinary user. Can we provide a YAML that will work for all/majority of users out of the box?
7. Visibility Timeout, Api Timeout - I think we can hide them (`show_user: false`)
8. I’m not sure what for are the fields Queue URL and Endpoint? It would be great to add 1-2 sentences descriptions with sample values. 


Observations (nothing that requires correction): /cc @ruflin
1. Tags Filter: it looks to me as a new widget instead of YAML.
2. In this case, logs and metrics share same settings. Does it sound like we need global integration-level variables?
3. User needs to provide either credential profile or access key + secret key + session token or credential file. Current state of UI allows for defining all of them and I don’t what happens there :)

kaiyan-sheng:
> 3. Not sure about defining „regions” as a different setting for every dataset? WDYT @ruflin

Use case would be user wants to collect metrics from both EC2 (in us-west-1) and SQS (us-east-2) (as an example).

> 5. I think we need to select a subset of datasets enabled by default. I suggest 5 biggest/most 
popular AWS services.

Will do, also I would disable beta version filesets/metricsets as well.

> 6. aws cloudwatch metrics - Metrics field: it contains values that will not work an ordinary user. Can we provide a YAML that will work for all/majority of users out of the box?

Yes, the current example will not work for ordinary user. We can remove the filters in the default `metrics` example like this:
```
   default: |
      - namespace: AWS/EC2
        tags.resource_type_filter: "ec2"
        name:
          - CPUUtilization
          - DiskWriteOps
        statistic:
          - Average
          - Maximum
```
But user won't know there are filters (`dimensions` and `tags`) available from just looking at the default value. Should we link to some documentation here?

> Observations (nothing that requires correction): /cc @ruflin
> 3. User needs to provide either credential profile or access key + secret key + session token or credential file. Current state of UI allows for defining all of them and I don’t what happens there :)

I'm thinking the same thing here 😄 I think we can use `show_user: false` to `access key + secret key + session token + role arn`. So user will first see using `credential profile name` as suggested.

mtojek:
> Use case would be user wants to collect metrics from both EC2 (in us-west-1) and SQS (us-east-2) (as an example).

On the other hand, user can deploy different configurations to different agents (in different regions) :) Any preference here, @ruflin ?

> But user won't know there are filters (dimensions and tags) available from just looking at the default value. Should we link to some documentation here?

You can mention their existence in the field description.  It should be ok for now ;)

> I'm thinking the same thing here 😄 I think we can use show_user: false to access key + secret key + session token + role arn. So user will first see using credential profile name as suggested.

Good idea and will work out of the box! Please go for it.

ruflin:
Catching up on all the comments:

* region per dataset or global: Both works. If we have it global, it means the user needs to configure AWS twice if he wants to have different regions. No preference on my end.
* Checkbox descriptions: Not sure I follow @mtojek 
* tags filter: how is this different from multi text?
* logs and metrics share setting: interesting, it basically means it is on the data source level. I wonder if an other concept called "credentials" could help here in the future. Or description as marcin noted.
* Default values: We could also provide some example values which are commented out so users know they exist?

mtojek:
> Checkbox descriptions: Not sure I follow @mtojek

e.g.

AWS DynamoDB metrics

Collect AWS DynamoDB metrics. Visit the link to browse the list of dimensions and metrics: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/metrics-dimensions.html#dynamodb-metrics-dimensions

ruflin:
Got it, could you file an issue for this in the Kibana repo?

mtojek:
> Got it, could you file an issue for this in the Kibana repo?

We discussed it offline, no need to report it to Kibana :)

kaiyan-sheng:
> * tags filter: how is this different from multi text?

@ruflin `tags.filter` requires a key value pair so I went directly with yaml type. But when I set type to yaml, `multi=true` will not work. Is it better to use multi text here instead?

> * Default values: We could also provide some example values which are commented out so users know they exist?

Done, I commented the filter values out. Thanks!

ruflin:
@kaiyan-sheng Not sure if multi text will work here, I think not (@jen-huang ?). I would go with the yaml version for now.

jen-huang:
Yes `multi` is for array of strings so I think yaml is better here for key:value pairs.

kaiyan-sheng:
👍 Thank you for all the help!!

ruflin:
@mtojek could you take a lead on moving this over to the integrations repo and add it to the storage. All future PR's should go against integrations repo.

mtojek:
> @mtojek could you take a lead on moving this over to the integrations repo and add it to the storage. All future PR's should go against integrations repo.

sure, will do!

PR #442, opened 2020-05-07
ci: Migrate job to JJBB

kuisathaverat:
It migrates the current job to JJBB, this means the Job definition is on the repo. It also moves the Jenkinsfile inside the `.ci` folder to keep CI stuff in the `.ci` folder only.


kuisathaverat:
After merging it, Branches/PRs made before the merge need to move the Jenkinsfile inside `.ci` folder.

.ci/package-storage.yml
```diff
@@ -0,0 +1,43 @@
+---
```
ruflin, line 1: Should this file be called `package-registry`?

kuisathaverat: C&P 🤦

PR #449, opened 2020-05-11
Update custom log package

ruflin:


dev/packages/example/log/0.9.0/dataset/log/agent/stream/stream.yml.hbs
```diff
@@ -5,3 +5,5 @@ paths:
 {{/each}}
 
 dataset: {{dataset}}
+
+{{custom}}
```
ruflin, line 9: @nchaulet I tried to have a fully custom part in this config here. When I try to fill it in with `foo: bar` in the UI I get the following error:

```
can not read an implicit mapping pair; a colon is missed at line 7, column 13: "##custom##" ^
```

If would assume it replaced `{{custom}}` by `foo: bar` so it should be valid?

nchaulet: This a good scenario that we did not cover, I did the compilation of yaml to be able to support values like `custom: {{custom}}`
https://github.com/elastic/kibana/blob/master/x-pack/plugins/ingest_manager/server/services/epm/agent/agent.test.ts

Thinking of the solution we have:

1. We can probably add a new concept in the registry variables, to flag a variable that should be spread to the agent yaml.
2. we can try to detect in the template that custom is a root level yaml value and do another special treatment, I have a POC here https://github.com/elastic/kibana/pull/66120

ruflin: ++ on option 2 as I would prefer to have as few "types" as possible on the package level.

Issue #447, opened 2020-05-11
Move alpha packages to package-storage and integrations


ruflin:
Currently some of the packages are still developed inside the package-registry. But development of the packages should happen in https://github.com/elastic/integrations or an other repo and then for each version, a PR against https://github.com/elastic/package-storage should be opened. To get to this point, a few things must happen:

* Move over all existing alpha packages to package-storage. I propose we use version 0.1.0 for these.
* Remove alpha packages from registry
* Update Docker images in registry and storage to use the new location for the build
  * Potentially update magefile in registry to allow to easily boot up a local env or should it be in storage or integrations?
* Update docs

ruflin:
@jonathan-buttner I wonder where we should move the development of the endpoint package? I would assume `integrations` repo would work for now. Like this you can benefit from all the tooling there.

mtojek:
Thanks for the issue.

> Move over all existing alpha packages to package-storage. I propose we use version 0.1.0 for these.

I assume we should wait with this one, once every package is done?

ruflin:
I would move already the ones that are completed to make sure no further changes happen in the package-registry and wait for the ones that still haven an open PR

mtojek:
The thing that concerns me is that there is no validation mechanism in the package-registry, so it might be hard to address/fix/reproduce bugs without copying it first to the `integrations`.

ruflin:
Doesn't have the package-registry the same validation mechanisms at the moment that integrations has?

Independent from that: No objections from my end to move over open PR's, I just worried it might slow down the contributors.

mtojek:
> Doesn't have the package-registry the same validation mechanisms at the moment that integrations has?

The tools like https://github.com/elastic/integrations/issues/14 will be available in integrations.

ruflin:
@mtojek I think I miss something here. I expect that the existing open PR's with packages will be merged in the next days and all future development happens in integrations. Did you mean above that no validation is happeing in the package-storage?

mtojek:
I see a situation in which developer needs to fix "redis" integration and debug the ingest pipeline. In "integrations" you have tools to do this. As the "package-storage" is a simple store, you have no tools, so it's easier for you to copy them to "integrations" first.

Of course we can make the integration-tools public and free to use everywhere, but not sure if we have too many storage places now.

jonathan-buttner:
[...]

PR #450, opened 2020-05-12
Docker: use packages from package-storage

mtojek:
Issue: https://github.com/elastic/package-registry/issues/447

This PR modifies package-registry to use packages from package-storage. It also publishes `example` packages (rsync).


magefile.go
```diff
@@ -32,7 +32,7 @@ var (
 
 	publicDir    = "./public"
 	buildDir     = "./build"
-	packagePaths = []string{"./dev/packages/alpha/", "./dev/packages/example/"}
+	packagePaths = []string{"./dev/packages/storage/", "./dev/packages/example/"}
```
ruflin, line 35: We could add this directory to gitignore and then have the magefile to also get the storage dependencies in. The nice part would be that then also `go run .` works with all of them.

mtojek: I will adjust the PR

mtojek: Done

PR #453, opened 2020-05-13
Add support for datastreams on base templates

ruflin:
The v2 templates are updated to use datastreams. This means now moving forward all data is ingested using datastreams. This requires at least Elasticesarch 7.9.

The data_stream config is put into the template itself and not one of the components as this part should never change.

In this PR, the base package is readded to the repository as currently no other place exists for development and testing.

### DIFF
--- a/dev/packages/example/base/0.2.0/docs/README.md
+++ b/dev/packages/example/base/0.2.0/docs/README.md
@@ -1,0 +1,3 @@
+# Base package
+
+This is installed in the background by EPM to setup the Elastic Stack.
--- a/dev/packages/example/base/0.2.0/elasticsearch/component-template/events-mappings.json
+++ b/dev/packages/example/base/0.2.0/elasticsearch/component-template/events-mappings.json
@@ -1,0 +1,148 @@
+{
+    "template": {
+        "mappings": {
+            "_meta": {
+            },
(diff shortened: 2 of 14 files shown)


skh:
What is the strategy to keep this change out of packages that end up being served to 7.8 users? Is this documented somewhere which packages go to the experimental / alpha registry?

ruflin:
The package-storage from where this is built in the very near future as a branch `experimental`: https://github.com/elastic/package-storage/tree/experimental This is what will be served by `epr-experimental.elastic.co`. As soon as all PR's are merged with the packages that should go into experimental I will do it once manually but do not expect it touch it again in the future.

skh:
As commented on https://github.com/elastic/kibana/pull/66367#pullrequestreview-410833867 I would prefer if we could wait a few days until the addition of datastreams to ES is a bit more complete.

ruflin:
@skh Now that the Kibana one is merged, should we get this in too?

simitt:
Are these templates and policies auto-generated, and if so, could you point me to the generator logic?

ruflin:
The logs, metrics templates are manually created. But most other templates which are part of the datasets, are generated based on the fields.yml in Kibana. Is that what you are looking for?

PR #454, opened 2020-05-13
Fix checkout issues when using 'experimental' branch

mtojek:
This PR fixes problems with checkout "experimental" branch of the package-storage.

Also, used "build" directory and changed `os.Remove` to `os.RemoveAll` when deleting `package-registry` (doesn't fail when the file is missing).

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -33,7 +33,7 @@ var (
 	publicDir      = "./public"
 	buildDir       = "./build"
 	storageDir     = "./dev/packages/storage"
-	storageRepoDir = "./.package-storage"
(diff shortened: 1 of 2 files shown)


magefile.go
```diff
@@ -33,7 +33,7 @@ var (
 	publicDir      = "./public"
 	buildDir       = "./build"
 	storageDir     = "./dev/packages/storage"
-	storageRepoDir = "./.package-storage"
+	storageRepoDir = filepath.Join(buildDir, ".package-storage")
```
ruflin, line 36: Lets not make it a `.`. Otherwise browsing it is harder.

mtojek: Fixed.

PR #460, opened 2020-05-14
Remove package version from path

mtojek:
Changes:
* add support for package sources with/without version in the path
* I left the "multiversion" package as is to verify legacy form (with version)

In the next I will flatten package paths by stripping the version


ruflin:
I think the problem here is, that parts of the generator should be part of the registry service.

My assumption in the future: The registry can be pointed at `package-storage` an will do all the building still needed on its own. No need for `mage build` initially.

What this means for the `integrations` repo is, that before the registry can be run on `package/mysql` without a version, there a build step must happen to build `bulid/package/mysql/0.2.0` and on this the registry is run.

mtojek:
This PR modifies generator to support two variants temporarily, until I will adjust all packages (integrations, package storage). I don't want to push incompatible, blocking changes.

If you take a look at the implementation, I changed the way the generator reads package version - from package path to `manifest.yml`.

mtojek:
> What this means for the `integrations` repo is, that before the registry can be run on `package/mysql` without a version, there a build step must happen to build `bulid/package/mysql/0.2.0` and on this the registry is run.

As I said, I've only changed internals, the built `public` directory contains exactly the same output as before.

> My assumption in the future: The registry can be pointed at package-storage and will do all the building still needed on its own. No need for mage build initially.

It smells like a deep refactoring, so the registry becomes a true static file server. We might plan how to rollout the change flawlessly for contributors. Honestly I would get rid of the `mage build` step at all and enforce the validation (I see an analogy to the HTTP server and serving static HTML files).

ruflin:
You mention above, you would also change the package-storage? But there we need all versions available and not only the most recent one.

mtojek:
We agreed offline to put this PR on hold due to other incoming changes to the registry repo (removing the `mage build` step).

mtojek:
I'm closing this, but will come up an alternative PR.

Issue #469, opened 2020-05-15
Generate index.json on-the-fly


mtojek:
Serve the `index.json` file for the package-registry:

```json
{
  "service.name": "package-registry",
  "version": "0.4.0"
}
```

PR #468, opened 2020-05-15
Serve favicon as embedded resource

mtojek:
Issue: https://github.com/elastic/package-registry/issues/463

PR #470, opened 2020-05-15
Generate index.json

mtojek:
Issue: https://github.com/elastic/package-registry/issues/469

_Let's first merge this PR: https://github.com/elastic/package-registry/pull/468_


index.go
```diff
@@ -0,0 +1,32 @@
...
+
+type indexData struct {
+	ServiceName string `json:"service.name"`
+	Version     string `json:"version"`
+}
+
+func indexHandler(cacheTime time.Duration) (func(w http.ResponseWriter, r *http.Request), error) {
+	data := indexData{
+		ServiceName: "package-registry",
+		Version:     "0.4.0",
```
ruflin, line 21: Nit: Could have this as a constant in the main.go file?

index.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+type indexData struct {
+	ServiceName string `json:"service.name"`
+	Version     string `json:"version"`
+}
+
+func indexHandler(cacheTime time.Duration) (func(w http.ResponseWriter, r *http.Request), error) {
+	data := indexData{
+		ServiceName: "package-registry",
```
ruflin, line 20: Nit: Could we have this as a constant?

mtojek:
> Changelog?

Yup, already added.

index.go
mtojek: Fixed

index.go
mtojek: Fixed

main.go
```diff
@@ -121,13 +124,20 @@ func getRouter(config Config, packagesBasePath string) (*mux.Router, error) {
...
 
 	router := mux.NewRouter().StrictSlash(true)
+	router.HandleFunc("/", indexHandlerFunc)
+	router.HandleFunc("/index.json", indexHandlerFunc)
 	router.HandleFunc("/search", searchHandler(packagesBasePath, config.CacheTimeSearch))
 	router.HandleFunc("/categories", categoriesHandler(packagesBasePath, config.CacheTimeCategories))
 	router.HandleFunc("/health", healthHandler)
 	router.HandleFunc("/favicon.ico", faviconHandleFunc)
-	router.PathPrefix("/").HandlerFunc(catchAll(http.Dir(config.PublicDir), config.CacheTimeCatchAll))
+	router.PathPrefix("/epr").HandlerFunc(catchAll(http.Dir(config.PublicDir), config.CacheTimeCatchAll))
```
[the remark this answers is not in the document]
mtojek: From the user perspective - no, they're already exposed. 

Speaking about the implementation, this has been already adjusted (see the master branch). Archives exposed via the `/epr` endpoint are built dynamically, not served from the file system.

mtojek:
> I would prefer to keep it in index.json so other tools can read it from the filesystem. This approach requires an HTTP request if I understand correctly.

This PR is only for the main `index.json` file. I've just started working on the package `index.json` files (PR soon).

jfsiii:
> This PR is only for the main index.json file.

Yes, that was my understanding.

I'll try to keep an eye out, but can you add me on that PR when you submit it? Thanks

PR #472, opened 2020-05-18
Stream archived content

mtojek:
This PR replaces serving archived artifacts from static resources, but supports creating them on the fly (on demand).

I added some unit tests for common cases.


artifacts.go
```diff
@@ -0,0 +1,137 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package main
```
ruflin, line 5: Could we move this code under `util` so we can reuse it? Lets assume we have one day a tool `elastic-package` that allows us to validate a package and also build the `.tar.gz` file to upload manually to Kibana. Would be nice to share this code there. It will probably required to separate the handler code from the building the tar.gz code but I would assume this makes it easier to test if not already the case.

mtojek: I'm ok with moving it to a different place except utils, commons, shared, etc. ;)

artifacts.go
```diff
@@ -0,0 +1,137 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/blang/semver"
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+)
+
+const artifactsRouterPath = "/epr/{packageName}/{packageName:[a-z]+}-{packageVersion}.tar.gz"
```
ruflin, line 22: I expected that we have some packages with an `_` but couldn't find one: https://github.com/elastic/integrations/tree/master/dev/packages/beats

testdata/artifacts/package/example/0.0.2/docs/README.md
```diff
@@ -0,0 +1,22 @@
+[role="xpack"]
```
ruflin, line 1: Couldn't you reuse here one of the existing test package? https://github.com/elastic/package-registry/tree/master/testdata/package

artifacts.go
ruflin: No strong opinion around the name. I assume this means we should also move the other `utils` code?

testdata/artifacts/package/example/0.0.2/docs/README.md
mtojek: I think this is possible now as I changed the way I compare package content (compare listings instead of created archive which was different depending on the platform).

artifacts.go
mtojek: Right, `php_fpm`

ruflin: Sounds like we need to support it then.

We should make sure that our validation code for package names (if we have one) validates for the same pattern. Perhaps we can even reuse it.

docs/api/example-0.0.2.tar.gz
```diff
@@ -0,0 +1,27 @@
+0 docs/
```
ruflin, line 1: I like the idea of this file but wondering about the `.tar.gz` part. Any good alternative? Something like `.tar.gz.sum` (also not too good).

testdata/artifacts/package/example/0.0.2/docs/README.md
ruflin: Lets use an existing one if possible. If we can't lets at least use the same path to have all test packages in one place and then I would remove quite a few of the assets to just have 1-2 files per type.

ruflin:
Don't forget the changelog ;-)

artifacts.go
mtojek: The entire project can be adjusted to the standard layout: https://github.com/golang-standards/project-layout

testdata/artifacts/package/example/0.0.2/docs/README.md
mtojek: Fixed

artifacts.go
mtojek: Moved to "archiver"

artifacts.go
mtojek: > Sounds like we need to support it then.

Done

> We should make sure that our validation code for package names (if we have one) validates for the same pattern. Perhaps we can even reuse it.

I don't see any. Do we have it?

artifacts.go
ruflin: SGTM to cleanup to follow the standard layout but lets do it in small steps.

artifacts.go
ruflin: Looks like we don't :-( https://github.com/elastic/package-registry/blob/master/util/package.go#L264

mtojek:
> Don't forget the changelog ;-)

Yes, already added.

docs/api/example-0.0.2.tar.gz
mtojek: Renamed to `example-0.0.2.tar.gz-preview.txt`

archiver/archive.go
```diff
@@ -0,0 +1,112 @@
...
+			log.Printf("Error occurred while closing tar writer: %v", err)
+		}
+
+		err = gzipWriter.Close()
+		if err != nil {
+			log.Printf("Error occurred while closing gzip writer: %v", err)
+		}
+	}()
+
+	err := filepath.Walk(packagePath, func(path string, info os.FileInfo, err error) error {
```
ruflin, line 34: Should we have some logic to skip hidden files that might be created by the system like `.DS_Store`?

archiver/archive.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func writeFileContentToArchive(path string, writer io.Writer) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return errors.Wrapf(err, "opening file failed (path: %s)", path)
+	}
+	defer func() {
+		err := f.Close()
+		if err != nil {
```
ruflin, line 102: What about using a named return to still return an error in case this fails instead of writing it to the log file?

artifacts.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const artifactsRouterPath = "/epr/{packageName}/{packageName:[a-z_]+}-{packageVersion}.tar.gz"
+
+var errArtifactNotFound = errors.New("artifact not found")
+
+func artifactsHandler(packagesBasePath string, cacheTime time.Duration) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		packageName := vars["packageName"]
```
ruflin, line 28: Do we need to validate that these entries exist?

artifacts.go
```diff
@@ -0,0 +1,59 @@
...
+		packageVersion := vars["packageVersion"]
+
+		_, err := semver.Parse(packageVersion)
+		if err != nil {
+			badRequest(w, "invalid package version")
```
ruflin, line 33: Would be nice to print out the package version that was invalid here.

archiver/archive.go
mtojek: I wouldn't pollute the code with references to files that are ".gitignored" anyway. The docker image is created and pushed in the CI environment, I don't think we can be afraid of presence of such files there.

archiver/archive.go
mtojek: This code is in defer, so it's not on the direct execution path. Also, if we've already started streaming and something really bad happens, we won't inform the use what happened, we can just cut the response stream.

artifacts.go
mtojek: Package name is restricted with regex:

`const artifactsRouterPath = "/epr/{packageName}/{packageName:[a-z_]+}-{packageVersion}.tar.gz"`

Package version is checked with `semver.Parse`:

```go
		_, err := semver.Parse(packageVersion)
		if err != nil {
			badRequest(w, "invalid package version")
			return
		}
```

archiver/archive.go
ruflin: This code is not only run inside docker but also locally.

artifacts.go
ruflin: I was more referring to check if the key exists: `vars["packageName"]` but as we have it in the router path I guess this should never happen as otherwise the routing would be wrong.

archiver/archive.go
ruflin: Not sure I get `This code is in defer, so it's not on the direct execution path`. You can still assign the error with a named return. So the function that calls "writeFileContentToArchive" can decide if it wants to drop the error or not.

artifacts.go
mtojek: The version is in URL and I don't expect that the user visits this endpoint anyway. I would leave it as is.

mtojek:
> I tried to run this with my local Kibana setup and got the following error:

I reproduced it, thanks! 

```
kibana_1            | {"type":"log","@timestamp":"2020-05-20T08:46:33+00:00","tags":["error","plugins","ingestManager"],"pid":6,"message":"Cannot find asset system-0.1.0/dataset/auth/fields/base-fields.yml"}
kibana_1            | {"type":"log","@timestamp":"2020-05-20T08:46:33+00:00","tags":["error","plugins","ingestManager"],"pid":6,"message":"Error: Cannot find asset system-0.1.0/dataset/auth/fields/base-fields.yml\n    at Object.getAsset (/usr/share/kibana/x-pack/plugins/ingest_manager/server/services/epm/registry/index.js:171:35)\n    at map (/usr/share/kibana/x-pack/plugins/ingest_manager/server/services/epm/packages/assets.js:73:29)\n    at Array.map (<anonymous>)\n    at getAssetsData (/usr/share/kibana/x-pack/plugins/ingest_manager/server/services/epm/packages/assets.js:71:26)\n    at process._tickCallback (internal/process/next_tick.js:68:7)"}
``` 

Investigating now...

mtojek:
Fixed. It was the issue with missing root in the package.

artifacts.go
mtojek: Correct

archiver/archive.go
mtojek: Yes, but on the other hand you can overwrite errors that caused the problem: https://play.golang.org/p/R4ZO8tvjVgr

mtojek: I used multierror for this. Fixed.

artifacts.go
mtojek: Fixed

archiver/archive.go
```diff
@@ -0,0 +1,143 @@
...
+
+// ArchivePackage method builds and streams an archive with package content.
+func ArchivePackage(w io.Writer, properties PackageProperties) (err error) {
+	gzipWriter := gzip.NewWriter(w)
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer func() {
+		var me multierror.Errors
+
+		if err != nil {
+			me = append(me, err)
```
ruflin, line 35: Nit: I stumbled over `me` on what it is. Perhaps use `multiErr` instead?

archiver/archive.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			me = append(me, errors.Wrapf(err, "closing tar writer failed"))
+		}
+
+		err = gzipWriter.Close()
+		if err != nil {
+			me = append(me, errors.Wrapf(err, "closing gzip writer failed"))
+		}
+
+		if me != nil {
```
ruflin, line 48: Nice to see the full list of errors bubbling up. Will make debugging easier.

archiver/archive.go
```diff
@@ -0,0 +1,143 @@
...
+			err = me.Err()
+		}
+	}()
+
+	rootDir := fmt.Sprintf("%s-%s", properties.Name, properties.Version)
+
+	err = filepath.Walk(properties.Path, func(path string, info os.FileInfo, err error) error {
```
ruflin, line 55: Out of curiosity: Will this follow symlinks?

archiver/archive.go
```diff
@@ -0,0 +1,143 @@
...
+
+func buildArchiveHeader(info os.FileInfo, relativePath string) (*tar.Header, error) {
+	header, err := tar.FileInfoHeader(info, "")
+	if err != nil {
+		return nil, errors.Wrapf(err, "reading file info header failed (info: %s)", info.Name())
+	}
+
+	header.Name = relativePath
+	if info.IsDir() {
+		header.Name = header.Name + "/"
```
ruflin, line 111: Is this `/` dependent on the platfrom or always `/`?

artifacts.go
```diff
@@ -0,0 +1,72 @@
...
+		w.Header().Set("Content-Type", "application/gzip")
+		cacheHeaders(w, cacheTime)
+
+		err = archiver.ArchivePackage(w, archiver.PackageProperties{
+			Name:    packageName,
+			Version: packageVersion,
+			Path:    packagePath,
+		})
+		if err != nil {
+			log.Printf("archiving package path '%s' failed: %v", packagePath, err)
```
ruflin, line 68: Is the error here already part of `w` so we don't need to have a badRequest here?

dev/generator/main.go
```diff
@@ -211,40 +211,6 @@ func buildPackage(packagesBasePath string, p util.Package) error {
 			return err
 		}
 	}
-
-	if tarGz {
```
ruflin, line 215: 👍

archiver/archive.go
mtojek: I've added an additional condition to check the `/`.

archiver/archive.go
mtojek: I will quote the docs here:

>// Walk does not follow symbolic links.

archiver/archive.go
mtojek: Fixed.

artifacts.go
mtojek: If the handler has already started streaming response (an archive) and an internal error occurs, you can't do anything in this situation. Headers are already sent.

You can simply cut-off the stream and wait for your upstream dependency to retry the activity.

PR #473, opened 2020-05-18
Update experimental build package reference

ruflin:
The experimental branch is used to build epr-experimental.elastic.co. It should not contain the example packages and also depend on the experimental branch from the package-stoarge.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -33,7 +33,7 @@ var (
 	publicDir      = "./public"
 	buildDir       = "./build"
 	storageRepoDir = filepath.Join(buildDir, "package-storage")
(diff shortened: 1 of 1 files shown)


ruflin:
@kuisathaverat I'm thinking we should push this to `package-registry:experimental`. What other parts would I have to change?

ruflin:
Tested this with Kibana 7.8 and seems to work.

kuisathaverat:
@ruflin we do not have to change anything on the Jenkinsfile it pushes based on the branch

ruflin:
Great, merged. Will test in a bit if it worked!

Issue #464, opened 2020-05-15
Generate package/index.json on the fly


mtojek:
Currently, `index.json` files are generated during the `mage build` step. The goal of this task is to build this resource dynamically.

PR #479, opened 2020-05-25
Generate package index.json

mtojek:
Issue: https://github.com/elastic/package-registry/issues/464

@jfsiii API hasn't changed in this PR, but I wish to get rid of `index.json` from inside of a bundle (`.tar.gz`).


docs/api/package.json
```diff
@@ -34,82 +34,14 @@
       "type": "image/png"
     }
   ],
-  "assets": [
```
ruflin, line 37: What about this missing content here?

main_test.go
```diff
@@ -63,7 +63,6 @@ func TestEndpoints(t *testing.T) {
 		{"/search?internal=bar", "/search", "search-package-internal-error.json", searchHandler(packagesBasePath, testCacheTime)},
 		{"/search?experimental=true", "/search", "search-package-experimental.json", searchHandler(packagesBasePath, testCacheTime)},
 		{"/search?experimental=foo", "/search", "search-package-experimental-error.json", searchHandler(packagesBasePath, testCacheTime)},
-		{"/package/example/1.0.0", "", "package.json", catchAll(http.Dir(publicPath), testCacheTime)},
```
ruflin, line 66: Could we still have a test for this? The API endpoint should still work.

docs/api/package.json
mtojek: Fixed. I didn't know I have to load assets additionally.

main_test.go
mtojek: This test is used in the `TestPackageIndex` method.

main_test.go
```diff
@@ -87,9 +86,34 @@ func TestArtifacts(t *testing.T) {
...
+	tests := []struct {
+		endpoint string
+		path     string
+		file     string
+		handler  func(w http.ResponseWriter, r *http.Request)
+	}{
+		{"/package/example/1.0.0/index.json", packageIndexRouterPath1, "package.json", packageIndexHandler},
+		{"/package/missing/1.0.0/index.json", packageIndexRouterPath1, "index-package-not-found.txt", packageIndexHandler},
+		{"/package/example/999.0.0/index.json", packageIndexRouterPath1, "index-package-revision-not-found.txt", packageIndexHandler},
+		{"/package/example/a.b.c/index.json", packageIndexRouterPath1, "index-package-invalid-version.txt", packageIndexHandler},
```
ruflin, line 116: Could we add a test that doesn't have `index.json` inside, for example `/package/example/1.0.0`?

archiver/archive.go
```diff
@@ -53,6 +55,11 @@ func ArchivePackage(w io.Writer, properties PackageProperties) (err error) {
 
 	rootDir := fmt.Sprintf("%s-%s", properties.Name, properties.Version)
 
+	err = writePackageIndexToArchive(properties.Path, rootDir, tarWriter)
```
ruflin, line 58: Could you add a note here that we have this in because Kibana still relies on it but it should potentially be removed in the future?

main_test.go
mtojek: Added

archiver/archive.go
mtojek: I left a note

PR #481, opened 2020-05-26
Improve base templates

ruflin:
This PR is to iterated on the base templates to come to a conclusion what the content should be.

At first, trying out how our validation will behave if I use comments in the template (which are valid in ES).

### DIFF
--- a/dev/packages/example/base/0.2.0/elasticsearch/component-template/logs-settings.json
+++ b/dev/packages/example/base/0.2.0/elasticsearch/component-template/logs-settings.json
@@ -5,15 +5,16 @@
                 "lifecycle": {
                     "name": "logs-default"
                 },
+                // Best compression optimises for storage
                 "codec": "best_compression",
                 "refresh_interval": "5s",
+                // Do we need number_of_shards default in this template as it is the same value as in ES?
                 "number_of_shards": "1",
                 "query": {
                     "default_field": [
                         "message"
                     ]
-                },
-                "number_of_routing_shards": "30"
+                }
             }
         },
         "aliases": {}
--- a/dev/packages/example/base/0.2.0/elasticsearch/component-template/metrics-settings.yml
+++ b/dev/packages/example/base/0.2.0/elasticsearch/component-template/metrics-settings.yml
@@ -1,0 +1,8 @@
+template:
+  settings:
+    index:
+      codec: "best_compression"
+      refresh_interval: "5s"
+      query:
+        default_field:
+          - "message"


dev/packages/example/base/0.2.0/elasticsearch/component-template/logs-settings.json
```diff
@@ -5,15 +5,16 @@
                 "lifecycle": {
                     "name": "logs-default"
                 },
+                // Best compression optimises for storage
```
ruflin, line 8: @jpountz This works with ES but most validators and Github are not very happy with this. What does ES use to validate these?

ruflin:
I realised perhaps a better way to write our index template and component templates is in yaml as it also allows to add comments and works around the issue of having comments in json. In addition, we already use yaml for quite a few ingest pipelines and often my personal view is, it makes them more readable. I think the same applies here.

@skh Would it be possible to support also yaml index templates and component templates in EPM?

dev/packages/example/base/0.2.0/elasticsearch/component-template/logs-settings.json
jpountz: Do you mean validating JSON files that include comments? We're relying on the library we use for JSON parsing, Jackson. https://github.com/FasterXML/jackson

ruflin: Yes, exactly. Most JSON validators I have seen so complain if they see comments. But potentially my workaround is to just use yaml instead (see other change in this PR).

skh:
> Would it be possible to support also yaml index templates and component templates in EPM?

As long as ES accepts them in yaml (I haven't tested that), we handle it the same way as we do with ingest pipelines and decide based on the file ending what format to send, see https://github.com/elastic/kibana/blob/master/x-pack/plugins/ingest_manager/server/services/epm/elasticsearch/ingest_pipeline/install.ts#L141-L143 .

So, yes, but it needs a code change in Ingest Manager / EPM.

ruflin:
@skh Yes, ES accepts yaml, tested it on my end.
```
curl -X PUT "elastic:changeme@localhost:9200/_component_template/foo?pretty" -H 'Content-Type: application/yaml' -d'
template:
  settings:
    index:
      codec: "best_compression"
      refresh_interval: "5s"
      query:
        default_field:
          - "message"
'
```
I will file an issue in KB for it.

ruflin:
Issue in Kibana repo: https://github.com/elastic/kibana/issues/67318

dev/packages/example/base/0.2.0/elasticsearch/component-template/logs-settings.json
jpountz: I'll bring this question to our Fix-it meeting to discuss what should be the way forward.

ruflin: @jpountz Not sure if ES needs to fix anything here as in ES things work just as expected (as long as I don't try to validate it somewhere :-) ).

jpountz: Maybe not indeed, but I think it's important to at least raise awareness about this problem, ie. the fact that we allow comments in JSON doesn't mean that it's easy for consumers of the Elasticsearch APIs to leverage this feature.

ruflin:
Closing this PR as changes to the base templates now happen in https://github.com/elastic/elasticsearch/pull/57629 As soon as this is merged, we should get the changes into the integrations repo.

PR #483, opened 2020-05-27
Remove requirement for streams definition in dataset manifest

ruflin:
Previously streams definition in the dataset manifest was required. This led to the problem that for example the endpoint package specified an input there even though it is not used: https://github.com/elastic/package-storage/blob/master/packages/endpoint/0.1.0/dataset/events/manifest.yml This removes the requirement for the streams definition and adds a package for testing it.


CHANGELOG.md
```diff
@@ -51,9 +51,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add basic access logs. [#400](https://github.com/elastic/integrations-registry/pull/400)
 * Validate ingest pipeline during packaging phrase. [#426](https://github.com/elastic/package-registry/pull/426)
 * Use http.FileServer to serve package content and define HTTP headers [#425](https://github.com/elastic/package-registry/pull/425)
+* Remove requirement for streams definition in dataset manifest. [#](https://github.com/elastic/package-registry/pull/)
 
 
-## [Unreleased](https://github.com/elastic/package-registry/compare/v0.2.0...v0.3.0)
```
ruflin, line 56: Found this issue while updating the changelog.

docs/api/example-0.0.2.tar.gz-preview.txt
```diff
@@ -17,6 +17,7 @@
 91348 example-0.0.2/kibana/index-pattern/filebeat.json
 0 example-0.0.2/kibana/infrastructure-ui-source/
 797 example-0.0.2/kibana/infrastructure-ui-source/default.json
+0 example-0.0.2/kibana/search/
```
ruflin, line 20: @mtojek Kind of suprised about this change. Lets see what CI thinks.

ruflin: Turns out I had locally still an empty directory probably which is not checked into git. Cleaning up the testdata directory and checking it out again solved this issue (I hope).

ruflin:
@jonathan-buttner Could you try to to update the endpoint package on the above?

PR #484, opened 2020-05-28
Do not compress archived data

mtojek:
This PR disables compressing on-the-fly for archived data. Even though the heaviest method on the request path is `LoadAssets()`, it will speed up processing.


mtojek:
There was a general idea (mostly for Kibana) for performance improvement focused on shortening total time of package installation and Kibana's boot up. As the size of resources is not huge, we can disable the compression and simply serve the content.

ruflin:
So the problem is that when it is compress, it needs more CPU resources on the Kibana side to decompress?

mtojek:
When it compresses it can stream the archive slower...

mtojek:
Decided to put this one on hold until it's a real problem.

PR #485, opened 2020-05-29
Detect package root directory

mtojek:
In the integrations repository package content won't be placed in the directory with version. The generator needs to support it this way.

### DIFF
--- a/dev/generator/main.go
+++ b/dev/generator/main.go
@@ -90,40 +90,57 @@ func CopyPackage(src, dst string) error {
 	return nil
 }
 
-// BuildPackage rebuilds the zip files inside packages
 func BuildPackages(sourceDir, packagesPath string) error {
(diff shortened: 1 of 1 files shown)


ruflin:
@mtojek I'm wondering if we will still need the generator in the package-registry in the future or if this becomes an integrations tool only. All the packages in the package-registry repo in the future should only be for testing and already be in the format, the registry can read from.

In summary, what I'm getting at: Would it make more sense to just copy over the generator code to integrations repo and start to phase it out here step by step instead of adding more logic here?

mtojek:
@ruflin At the most we also need the generator in the `package-storage`. Do you have any suggestion what to do that repository?

mtojek:
Also, the crucial part at the moment is a focus on flattening the project structure and developing scripts that can support the process of updating package-registry. If that's not a problem, I prefer to deliver that one first.

PR #488, opened 2020-06-02
Delete package index.json

mtojek:
Delete package `index.json` from archive and do not serve it on endpoint (not used by Kibana).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+* Delete package index.json from archives. Don't serve index.json as resource. [#488](https://github.com/elastic/package-registry/pull/488)
+
 ### Known Issue
 
 ## [Unreleased](https://github.com/elastic/package-registry/compare/v0.3.0...v0.4.0)
--- a/archiver/archive.go
+++ b/archiver/archive.go
@@ -7,7 +7,6 @@ package archiver
 import (
 	"archive/tar"
 	"compress/gzip"
-	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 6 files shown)


mtojek:
I'm confused now. The `index.json` is still accessible on `/` and Kibana doesn't use the one in the archive. Could you please provide detailed overview on why this should stay as is?

ruflin:
I misread the code and I thought it also removes the one from the service and `/` is not possible anymore. My bad.

About removing it from the package: Did you try building a config and save it in Kibana? I'm curious if it still works.

jfsiii:
I ran the registry locally with

```
docker pull docker.elastic.co/package-registry/package-registry:PR-488
docker run -p 8080:8080  $(docker image ls docker.elastic.co/package-registry/package-registry:PR-488 -q)
```

and ES with (needs gold+ to set the registry)

```
yarn es snapshot --license=trial  -E xpack.security.authc.api_key.enabled=true
```

and Kibana with

```
 yarn start --no-base-path --xpack.ingestManager.enabled=true --xpack.ingestManager.epm.registryUrl=http://localhost:8080
```

then added a Data Source for nginx and everything looked good
Screen Shot 2020-06-02 at 7 34 27 AM

mtojek:
I confirm that it works too, but accidentally found a side issue in Kibana: https://github.com/elastic/kibana/issues/67938

nchaulet:
I am curious here if deleting index.json is not the wrong solution, should we not use it in the packageIndexHandler instead?

mtojek:
We're adjusting the solution to not require `mage build` at all, hence the `index.json` is created dynamically. Keep in mind that in the future we'll have 100000.. of revisions, so we might face scaling problems storing all of these.

ruflin:
@nchaulet Mid and long term users will be able to build packages in Kibana and upload them manually to Kibana. For these packages, the index.json is missing. This change makes sure, they are not used "accidentially" by the Ingest Manager when not accessing the registry directly.

PR #489, opened 2020-06-02
Fix typo in README

ruflin:
This is also to check if CI is going red because of an invalid package or not.


ruflin:
CI passed after fixing the package-storage.

PR #493, opened 2020-06-03
Basic integration test with Kibana

ruflin:
This is a simple integration test to validate changes made against the registry are still working with the current version of Kibana.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -94,9 +94,13 @@ pipeline {
         deleteDir()
         unstash 'source'
         dir("${BASE_DIR}"){
-          insideGo(){
-            sh(label: 'Runs the (integration) tests',script: 'mage -debug testIntegration ')
-          }
+          sh(label: 'Runs the (integration) tests',script: '''
+            eval "$(gvm 1.14.2)"
+            go get -u github.com/magefile/mage
+            go get -u github.com/elastic/go-licenser
+            go get -u -v golang.org/x/tools/cmd/goimports
+            mage -debug testIntegration
+          ''')
         }
       }
     }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -87,6 +87,20 @@ pipeline {
       }
     }
     /**
+     Execute integration tests.
+     */
+    stage('TestIntegration') {
+      steps {
+        deleteDir()
+        unstash 'source'
+        dir("${BASE_DIR}"){
+          insideGo(){
+            sh(label: 'Runs the (integration) tests',script: 'mage -debug testIntegration ')
+          }
+        }
+      }
+    }
+    /**
      Publish Docker images.
      */
     stage('Publish Docker image'){
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -90,6 +90,10 @@ pipeline {
      Execute integration tests.
      */
     stage('TestIntegration') {
+      environment {
+        HOME = "${env.WORKSPACE}"
+        PATH = "${env.HOME}/go/bin:${env.PATH}"
+      }
       steps {
         deleteDir()
         unstash 'source'
--- a/magefile.go
+++ b/magefile.go
@@ -142,6 +142,10 @@ func Test() error {
 	return sh.RunV("go", "test", "./...", "-v", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")
 }
 
+func TestIntegration() error {
+	return sh.RunV("go", "test", "./...", "-v", "-tags=integration")
+}
+
 // Format adds license headers, formats .go files with goimports, and formats
 // .py files with autopep8.
 func Format() {
--- a/magefile.go
+++ b/magefile.go
@@ -143,7 +143,7 @@ func Test() error {
 }
 
 func TestIntegration() error {
-	return sh.RunV("go", "test", "./...", "-v", "-tags=integration")
+	return sh.RunV("go", "test", "./...", "-v", "-tags=integration", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")
 }
 
 // Format adds license headers, formats .go files with goimports, and formats
--- a/main_integration_test.go
+++ b/main_integration_test.go
@@ -2,6 +2,8 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// +build integration
+
 package main
 
 import (
@@ -21,7 +23,7 @@ import (
 // and the setup command works as expected.
 func TestSetup(t *testing.T) {
 
-	err := os.Chdir("environments")
+	err := os.Chdir("testing/environments")
 	if err != nil {
 		t.Error(err)
 	}
--- a/main_integration_test.go
+++ b/main_integration_test.go
@@ -23,7 +23,12 @@ import (
 // and the setup command works as expected.
 func TestSetup(t *testing.T) {
 
-	err := os.Chdir("testing/environments")
+	currentDir, err := os.Getwd()
+	if err != nil {
+		t.Error(err)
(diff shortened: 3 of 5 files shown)


testing/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestSetup(t *testing.T) {
+
+	err := os.Chdir("environments")
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Make sure services are shut down again at the end of the test
+	defer func() {
+		err = sh.Run("docker-compose", "-f", "snapshot.yml", "-f", "local.yml", "down", "-v")
```
ruflin, line 27: This spins up the Stack + Registry but pulls in master. Instead it should build the local registry and run it.

ruflin:
@kuisathaverat As you see in this PR, I want to run a locally built registry against the stack setup. I guess the easiest case will be if all 3 are containers and the registry container is built as part of the test. Any recommendations on other ways to do it?

kuisathaverat:
Because your Dockerfile build the binaries you can delegate on docker-compose to build and start the Docker container, it is something that we make on the Opbeans

https://github.com/elastic/opbeans-java/blob/master/docker-compose.yml#L3-L6

something like this should work

```
# Defines if ports should be exported.
# This is useful for testing locally with a full elastic stack setup.
# All services can be reached through localhost like localhost:5601 for Kibana
# This is not used for CI as otherwise ports conflicts could happen.
version: '2.3'
services:
  kibana:
    ports:
      - "127.0.0.1:5601:5601"
    depends_on:
      elasticsearch:
        condition: service_healthy

  elasticsearch:
    ports:
      - "127.0.0.1:9200:9200"

  integrations-registry:
    build: ../..
    image: docker.elastic.co/package-registry/package-registry:master
    ports:
      - "127.0.0.1:8080:8080"
```

testing/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			t.Error(err)
+		}
+
+	}()
+	// Spin up services
+	go func() {
+		err = sh.Run("docker-compose", "-f", "snapshot.yml", "-f", "local.yml", "up", "--force-recreate")
```
mtojek, line 35: How about `docker-compose pull`?

ruflin: Doesn't the pull happening automatically if images are missing?

mtojek: "if images are missing" 

so if you rerun the test locally after a week, you won't get fresh images.

ruflin: Adjusted

ruflin:
@kuisathaverat What you recommended works, exactly what I needed. Next problem: I think `docker-compose` is not available. What is the best way to get it on the test server? Will this even work as I think we run our tests already inside a docker container 🤔

ruflin:
@EricDavisX @ph As soon as I get this working, I think this adds one more building block to our testing. It makes sure that it tests each change to the package registry with the current master snapshot of Kibana and runs the ingest management setup. Later on we might pick multiple Kibana versions to test against ...

ph:
@ruflin Lets add it to the testing issue when its stable. I would like to update as we improve out testing.

ph:
^ that should trigger a ping.

kuisathaverat:
>@kuisathaverat What you recommended works, exactly what I needed. Next problem: I think docker-compose is not available. What is the best way to get it on the test server? Will this even work as I think we run our tests already inside a docker container

yes, it is inside a Docker container so we have to change it to use a Go version installed on the worker. I will make a PR with the change.

.ci/Jenkinsfile
```diff
@@ -86,6 +86,20 @@ pipeline {
...
+     Execute integration tests.
+     */
+    stage('TestIntegration') {
+      steps {
+        deleteDir()
+        unstash 'source'
+        dir("${BASE_DIR}"){
+          insideGo(){
+            sh(label: 'Runs the (integration) tests',script: 'mage -debug testIntegration ')
+          }
```
kuisathaverat, line 99: for the moment use this script to run the test instead of indideGo function

```suggestion
          sh(label: 'Runs the (integration) tests',script: '''
            eval "$(gvm 1.14.2)"
            go get -u github.com/magefile/mage
            go get -u github.com/elastic/go-licenser
            go get -u -v golang.org/x/tools/cmd/goimports
            mage -debug testIntegration
          ''')
```

ruflin:
@kuisathaverat It seems mage is installed but then it cannot use it? https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fpackage-registry/detail/PR-493/12/pipeline

.ci/Jenkinsfile
```diff
@@ -86,6 +86,24 @@ pipeline {
         }
       }
     }
+    /**
+     Execute integration tests.
+     */
+    stage('TestIntegration') {
```
kuisathaverat, line 92: There are a couple of missing env vars

```suggestion
    stage('TestIntegration') {
            environment {
              HOME = "${env.WORKSPACE}"
              PATH = "${env.HOME}/go/bin:${env.PATH}"
            }
```

.ci/Jenkinsfile
```diff
@@ -86,6 +86,24 @@ pipeline {
...
+        dir("${BASE_DIR}"){
+          sh(label: 'Runs the (integration) tests',script: '''
+            eval "$(gvm 1.14.2)"
+            go get -u github.com/magefile/mage
+            go get -u github.com/elastic/go-licenser
+            go get -u -v golang.org/x/tools/cmd/goimports
+            mage -debug testIntegration
+          ''')
+        }
+      }
```
kuisathaverat, line 109: This grabs the JUnit results

```suggestion
      }
      post {
        junit(allowEmptyResults: true, keepLongStdio: true, testResults: "junit-report.xml")
      }
```

kuisathaverat: wait, TestIntegration does not generate JUnit results like Test do

ruflin: I wonder if it has to. I packed it into a golang test file but it could also be a binary that is executed through `go run .`.

ruflin:
Tests are green and working! Thanks for the help @kuisathaverat I suggest we get this in soonish and then iterate on it.

magefile.go
```diff
@@ -142,6 +142,10 @@ func Test() error {
 	return sh.RunV("go", "test", "./...", "-v", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")
 }
 
+func TestIntegration() error {
+	return sh.RunV("go", "test", "./...", "-v", "-tags=integration")
```
kuisathaverat, line 146: This should generate a JUnit report for the integration tests

```suggestion
	return sh.RunV("go", "test", "./...", "-v", "-tags=integration", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")
```

ruflin: added it to see what happens! :-)

ruflin:
@mtojek WDYT about getting this in?

main_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+		err = sh.Run("docker-compose", "-f", "snapshot.yml", "-f", "local.yml", "up", "--force-recreate", "--remove-orphans", "--build")
+		if err != nil {
+			t.Error(err)
+		}
+	}()
+
+	// Check for 5 minutes if service is available
+	for i := 0; i < 5*60; i++ {
+		output, _ := sh.Output("docker-compose", "-f", "snapshot.yml", "-f", "local.yml", "ps")
```
jonathan-buttner, line 59: Will rerunning the `docker-compose` command cause any issues while previous commands are starting up?

ruflin: If nothing is running, it returns an error. This is the reason I skip the error check. But as this is "only" the `ps` command it should not have any side effects. But I might miss something here.

PR #496, opened 2020-06-04
Remove magefile / generator code which is not needed anymore

ruflin:
The writeJsonFile code was not used anymore. And on the generator side the Load* methods are also not needed anymore, as no JSON files are generated.


ruflin:
jenkins, test this

PR #498, opened 2020-06-08
fix: correct path for JUnit test results

kuisathaverat:
The path for the JUnit test results is wrong and because of that, the test results are not reported to Jenkins.

remove the github.com/jstemmer/go-junit-report dependency from the project and the Docker image for the go-agent are no longer used.


kuisathaverat:
They are not there too 🤔 
```
[2020-06-08T11:47:09.050Z] Recording test results
[2020-06-08T11:47:09.204Z] None of the test reports contained any result
```

kuisathaverat:
This command generate the report file 
`go test ./... -v 2>&1|go-junit-report > junit-report.xml`

but this no
`mage test`

It runs the same command but from Go
`return sh.RunV("go", "test", "./...", "-v", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")`

@ruflin Do you have any idea why?

ruflin:
My assumption would be that `mage` swallows some things? @andrewkroh might know more?

kuisathaverat:
Mage capture the stdout and stderr so I think we have to make it in another way.

```
// RunV is like Run, but always sends the command's stdout to os.Stdout.
func RunV(cmd string, args ...string) error {
	_, err := Exec(nil, os.Stdout, os.Stderr, cmd, args...)
	return err
}
```

ruflin:
@kuisathaverat So using RunV instead Run should fix it?

kuisathaverat:
no, Run makes the same, I am going to ask @mdelapenya because I know we have XUnit output in the e2e test.

mdelapenya:
We use `go-junit-report` as a binary to capture the `go test` unit test output. Maybe you can use the library as a library, generating the XML file directly from `mage`

kuisathaverat:
an improvement could be to show the output progressively on the command line, not at the end of the test, but I am a Go newbie.

mdelapenya:
One question: is the formatter of the `go-junit-report` library used elsewhere? It's main purpose is to generate xUnit XML files for the CI, not for local development, so I'd suggest not depending on that in the source code, but using it directly on CI time.

kuisathaverat:
this implies to have an additional script to post-process the Go test results and call go-junit-report

magefile.go
```diff
@@ -128,7 +132,39 @@ func Test() error {
...
 		return err
 	}
-	return sh.RunV("go", "test", "./...", "-v", "2>&1", "|", "go-junit-report", ">", "junit-report.xml")
+
+	output, errTest := sh.OutputWith(nil, "go", "test", "./...", "-v")
+	if errTest != nil {
+		return errTest
+	}
+
+	fmt.Println(output)
```
ruflin, line 141: debug left over or you need this one?

.ci/Jenkinsfile
```diff
@@ -95,16 +94,19 @@ pipeline {
...
-            eval "$(gvm 1.14.2)"
-            go get -u github.com/magefile/mage
-            go get -u github.com/elastic/go-licenser
-            go get -u -v golang.org/x/tools/cmd/goimports
-            mage -debug testIntegration
-          ''')
+        cleanup()
+        withMageEnv(){
+          dir("${BASE_DIR}"){
+            sh(label: 'Runs the (integration) tests',script: 'mage -debug testIntegration|tee tests-report.txt')
```
v1v, line 100: Redirect stderr?

```suggestion
            sh(label: 'Runs the (integration) tests',script: 'mage -debug testIntegration 2>&1 | tee tests-report.txt')
```

.ci/Jenkinsfile
```diff
@@ -70,19 +69,19 @@ pipeline {
...
     stage('Test') {
       steps {
-        deleteDir()
-        unstash 'source'
+        cleanup()
         dir("${BASE_DIR}"){
-          insideGo(){
-            sh(label: 'Runs the (unit) tests',script: 'mage -debug test ')
+          withMageEnv(){
+            sh(label: 'Runs the (unit) tests',script: 'mage -debug test|tee tests-report.txt')
```
v1v, line 75: ditto

```suggestion
            sh(label: 'Runs the (unit) tests',script: 'mage -debug test 2>&1 | tee tests-report.txt')
```

kuisathaverat: I have tested and there is nothing in the stderr even do some tests fail

.ci/Jenkinsfile
kuisathaverat: I have tested and there is nothing in the stderr even do some tests fail

andrewkroh:
Beats uses that junit report tool as a library from mage. It tee's the go test output to the console and to a file so you can see the output live. Then after the tests finish it uses the library to generate report. https://github.com/elastic/beats/blob/05c906553e7158a15cb571dfdc23596f45c07750/dev-tools/mage/gotest.go#L226-L262

kuisathaverat:
@andrewkroh at some point it was my implementation but as @ruflin said it is a lot of code (https://github.com/elastic/package-registry/pull/498#pullrequestreview-427181937) to generate a file that developers will not use in their local environments. The CI makes the transformation with a couple of commands. In that way the magefile is simple and the Jenkinsfile too

Issue #478, opened 2020-05-25
Change usage of `categories`


ruflin:
The current categories in the packages are `logs`, `metrics`. But instead these should be more like `database`, `cache` etc. 

The values for the dataset type which is `logs`, `metrics` and should not have to be set by the users as it should be possible to extract these from the datasets.

ruflin:
With https://github.com/elastic/package-registry/issues/482 moving forward for me it also clarifies where we should be going here:

* For `dataset.type` we already have it today inside each dataset manifest: https://github.com/elastic/package-registry/blob/master/testdata/package/default-pipeline/0.0.2/dataset/foo/manifest.yml#L4 We must validate that these can only be `logs` or `metrics` for now.
* If needed for querying, these types could also be exposed as a summary of all datasets on the package level

For the categories, I'm starting to think the category should be defined per dataset and not on the package level. And what we present on the package level for categories, is a summary of the categories in each dataset. I assume in most cases. all dataset of a single package will have the same category but it is not a requirement. @mostlyjason WDYT?

Following the above, we would have `category: foo` in each dataset. The nice part about this is that it also makes sure each dataset has only one category. These should then be validated against the valid categories. 

@mostlyjason Is this list of categories valid?
```
Cloud
Config management
Containers
AWS
Azure
Cointainer
CRM
Custom
Datastore
Elastic Stack
Google Cloud
Kubernetes
Languages
Message Queue
Monitoring
Network
Notification
Orchestration
OS & System
Productivity
Security
Support
Ticketing
Version Control
Web
```

mostlyjason:
Linking this issue so we can follow recommendations made there https://github.com/elastic/dev/issues/1429

ruflin:
I opened https://github.com/elastic/package-registry/pull/534 with the change. For now only package categories are adjusted.

ruflin:
Closing as https://github.com/elastic/package-registry/pull/534 was merged and old categories were removed.

PR #501, opened 2020-06-10
Add validation for dataset type

ruflin:
The valid dataset types at the moment are logs, metrics and events. events will be removed in the future but at the moment it is still used in the endpoint package. As soon as this is removed, the code here will be cleaned up.

With recent changes to the code, the datasets were not validated anymore during build or running the registry. This is now changed by adding validate for datasets every time a package is created.

Part of https://github.com/elastic/package-registry/issues/478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Generate index.json file. [#470](https://github.com/elastic/package-registry/pull/470)
 * Stream archived package content. [#472](https://github.com/elastic/package-registry/pull/472)
 * Generate package index.json files. [#479](https://github.com/elastic/package-registry/pull/479)
+* Add validation for dataset type. [#](https://github.com/elastic/package-registry/pull/)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Generate index.json file. [#470](https://github.com/elastic/package-registry/pull/470)
 * Stream archived package content. [#472](https://github.com/elastic/package-registry/pull/472)
 * Generate package index.json files. [#479](https://github.com/elastic/package-registry/pull/479)
-* Add validation for dataset type. [#](https://github.com/elastic/package-registry/pull/)
+* Add validation for dataset type. [#501](https://github.com/elastic/package-registry/pull/501)
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


ruflin:
@jonathan-buttner Would be great to get your feedback on this one as the endpoint package will need an update.

util/dataset.go
```diff
@@ -24,6 +24,13 @@ const (
 	DirIngestPipeline = "ingest-pipeline"
 )
 
+var ValidTypes = map[string]interface{}{
```
mtojek, line 27: I wonder if `ValidTypes` can be package visible only?

util/dataset.go
```diff
@@ -191,6 +202,11 @@ func (d *DataSet) Validate() error {
 	return nil
 }
 
+func (d *DataSet) ValidType() bool {
```
mtojek, line 205: Same question here: can this method be package visible?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Generate index.json file. [#470](https://github.com/elastic/package-registry/pull/470)
 * Stream archived package content. [#472](https://github.com/elastic/package-registry/pull/472)
 * Generate package index.json files. [#479](https://github.com/elastic/package-registry/pull/479)
+* Add validation for dataset type. [#](https://github.com/elastic/package-registry/pull/)
```
mtojek, line 22: PR reference is missing

util/dataset.go
```diff
@@ -24,6 +24,13 @@ const (
 	DirIngestPipeline = "ingest-pipeline"
 )
 
+var ValidTypes = map[string]interface{}{
+	"logs":    nil,
```
mtojek, line 28: This might cause problems in the future if someone would like to check if the value `logs` exists. They would get a `nil` value :) Maybe bool instead?

util/dataset.go
ruflin: yes, changed.

util/dataset.go
ruflin: yes, changed

CHANGELOG.md
ruflin: fixed

util/dataset.go
ruflin: I went an other approach and made the second part a String which is something we could also show in a UI if needed. Thinking of the case that someday we might have a `/types` endpoint if someone wants to know all the available dataset types.

util/dataset.go
```diff
@@ -24,6 +24,13 @@ const (
 	DirIngestPipeline = "ingest-pipeline"
 )
 
+var validTypes = map[string]string{
+	"logs":    "Logs",
```
mtojek, line 28: For string-to-string enums there is a different approach recommended: stringer (see: https://godoc.org/golang.org/x/tools/cmd/stringer). 

We can leave it as, it's not a full blown enum

PR #503, opened 2020-06-10
Add -dry-run flag to registry

ruflin:
The -dry-run flag allows to run the registry without starting the service. This has the nice benefit that the registry itself can be started to validate all the packages and directly checking the output for errors.

A test to the integration tests was added to use this behaviour.


magefile.go
```diff
@@ -132,6 +132,18 @@ func Test() error {
 }
 
 func TestIntegration() error {
+	// Build is need to make sure all packages are built
+	err := Build()
+	if err != nil {
+		return err
+	}
+
+	// Checks if the binary is properly run and does not return any errors
```
mtojek, line 141: nit: is properly running

main.go
```diff
@@ -33,6 +33,7 @@ const (
 var (
 	packagesBasePath string
 	address          string
+	dryRun           bool
```
mtojek, line 36: Dry run term might be confusing in terms of a web service (what is a registry doing if it's not playing a role of a web service).
The real function you've enabled here is package validation. What do you think about splitting code into phases (representing states): e.g. validation, running? If the service didn't start, you would know which phase did it fail.

ruflin: If we go the above route, how would the integration test then look like?

magefile.go
ruflin: fixed

main.go
mtojek: Building packages can be considered as `@BeforeClass` (a requirement for test) :) Let's say we have two test cases:

TestValidation:

```
Cmd args: "-state=validation"
Expected: app exits with code 0, message: VALIDATION OK or VALIDATION FAILED: Message why did it fail
```

TestRunning (without validation):

```
Cmd args: "-state=running"
Expected: process is running, web server exposed on port A
```

TestDefault (so including validation):

```
Cmd args: (none)
Expected: process is running, web server exposed on port A
```

My point is to enable testing on multiple levels the process startup. I'm not insisting on this change if you think this in an overkill here. I'm not convinced to applying the term "dry-run" to a web service.

ruflin: I see where you are going here but not sure we need this at the moment. Also not too happy with the naming of the flag. Any alternative suggestion?

mtojek: I think it's all about the name of the flag right now, so if we can't find a best option for both of us, let's keep it as is ;)

PR #504, opened 2020-06-10
Remove solution entry in package manifest

ruflin:
Based on the discussion in https://github.com/elastic/kibana/issues/67939 it seems like we do not need the solution entry in the manifest. This removes it.


ruflin:
@jonathan-buttner FYI as it should also be removed from the endpoint package I think if we move forward with this.

ruflin:
@jonathan-buttner Yes, we only validate fields we know about and ignore everything else.

PR #507, opened 2020-06-15
dry-run: validate required fields

mtojek:
Changes:
* while loading new dataset validate required fields (also in dry-run)


util/package.go
```diff
@@ -397,26 +402,123 @@ func (p *Package) LoadDataSets() error {
...
+		err = validateRequiredFields(datasetBasePath)
+		if err != nil {
+			return errors.Wrapf(err, "validating required fields failed (path: %s)", datasetPath)
+		}
+	}
+	return nil
+}
+
+// validateRequiredFields method loads fields from all files and checks if required fields are present.
+func validateRequiredFields(datasetPath string) error {
```
ruflin, line 428: Shouldn't this be part of the Dataset validation. Like this also now datasetBasePath has to be passed in.

mtojek: Fixed.

BTW I removed the constraint for not having `-` (dash) in dataset names. We have some in the `aws` integration.

ruflin: Hm, this is not good because it means the `-` will end up in the index name which breaks the indexing strategy. Or do I miss something here?

mtojek: `aws` integration:

```
ec2-logs
ec2-metrics
elb-logs
elb-metrics
s3_daily_storage
s3_request
```

I understand that we need to rename ones with `-` (dash). Is the underscore safe? Should we switch from dash to underscore in integrations?

ruflin: `_ ` are allowed. The [indexing strategy](https://github.com/elastic/kibana/blob/ff97a034eefe2f4302cd86f47846bb386937774d/x-pack/plugins/ingest_manager/dev_docs/indexing_strategy.md) uses the following patterns: `logs-*-*`. Because of this if there is a `-` in the dataset name, it will assume it is the namespace. All other chars except `-` are allowed.

ruflin: To add a bit more flavour to this: A package could be allowed to have a `-`. It is only the dataset that is not allowed to have one. But as in most cases (not all) we autogenerate the dataset name (currently id) out of the package name, it is better they don't.

mtojek: Thanks for the reference to docs. I will prepare a set of PRs to rename all invalid datasets.

mtojek: https://github.com/elastic/integrations/pull/76

ruflin:
@mtojek You also need to update the test packages :-(

PR #508, opened 2020-06-15
Do not require config.yml

mtojek:
Changes:
* do not require config.yml to be present
* update Dockerfile
* remove config.yml from the repository

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,5 +1,0 @@
-public_dir: "./public"
-
-cache_time.search: 10m
-cache_time.categories: 10m
-cache_time.catch_all: 10m
--- a/main.go
+++ b/main.go
@@ -117,6 +117,7 @@ func main() {
 func getConfig() (*Config, error) {
 	cfg, err := ucfgYAML.NewConfigWithFile(configPath)
 	if os.IsNotExist(err) {
+		log.Println(`Using default configuration options as "config.yml" is not available.`)
 		return &defaultConfig, nil
 	}
 	if err != nil {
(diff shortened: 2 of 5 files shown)


main.go
```diff
@@ -37,7 +39,7 @@ var (
 	configPath       = "config.yml"
 
 	defaultConfig = Config{
-		PublicDir:           "config.yml",
+		PublicDir:           "public",
```
ruflin, line 42: ups :-(

main.go
```diff
@@ -114,16 +116,18 @@ func main() {
 
 func getConfig() (*Config, error) {
 	cfg, err := ucfgYAML.NewConfigWithFile(configPath)
+	if os.IsNotExist(err) {
+		return &defaultConfig, nil
```
ruflin, line 120: Could you add a log message here that the default configs are used as no config was found?

config.yml
```diff
@@ -1,5 +0,0 @@
-public_dir: "./public"
```
ruflin, line 1: I would still like to keep this one to show users what configs are there. It is also part documentation.

mtojek: How about `config.reference.yml`?

ruflin: ++. Should we fall back to it or just have it there?

mtojek: There won't be any `config.yml` by default as it's not required, but there is will be a `config.reference.yml` if users would like to preview the configuration file.

ruflin: SGTM

main.go
mtojek: Fixed

config.yml
mtojek: Fixed

Issue #509, opened 2020-06-15
Add validation that package name/version are in line with the manifest


ruflin:
In https://github.com/elastic/package-storage/pull/63 we had the problem, that the content of package manifest was not fully aligned with the directory names and versions. We should have a check to validate this.

PR #510, opened 2020-06-15
Validate version consistency

mtojek:
Issue: https://github.com/elastic/package-registry/issues/509


util/package.go
```diff
@@ -316,9 +315,32 @@ func (p *Package) Validate() error {
...
+func (p *Package) validateVersionConsistency() error {
+	versionPackage, err := semver.Parse(p.Version)
+	if err != nil {
+		return errors.Wrap(err, "invalid version defined in manifest")
+	}
+
+	baseDir := filepath.Base(p.BasePath)
+	versionDir, err := semver.Parse(baseDir)
+	if err != nil {
+		return nil // package content is not rooted in version directory
```
ruflin, line 335: Do you expect us to support this long term or only at the moment?

util/package.go
```diff
@@ -316,9 +315,32 @@ func (p *Package) Validate() error {
...
+	err = p.validateVersionConsistency()
+	if err != nil {
+		return errors.Wrap(err, "version in manifest file is not consistent with path")
+	}
+
 	return p.ValidateDatasets()
 }
 
+func (p *Package) validateVersionConsistency() error {
+	versionPackage, err := semver.Parse(p.Version)
```
ruflin, line 327: You could use this one here: https://github.com/elastic/package-registry/blob/d4a6d00d7eed510b8630273acdf061ec69a47b58/util/package.go#L38

main.go
```diff
@@ -32,10 +32,9 @@ const (
 )
 
 var (
-	packagesBasePath string
```
ruflin, line 35: Seems like we don't use this anymore? My initial intention with this was to be able to pass it as a flag to to the docker container on which directories it should read.

util/package.go
mtojek: I planned to leave it there long term. I put this check just in case.

util/package.go
mtojek: No, I couldn't. The `Validate()` method is called by `Marshal()`, so the mentioned field is not filled yet.

main.go
mtojek: I the next step I have a plan to support multiple sources of packages (exactly for the docker environment). I think we can leave it as is.

Issue #474, opened 2020-05-18
Discuss: Move to zip instead of `tar.gz` for the format of the package.


ph:
Currently, the package registry uses the `tar.gz` as the compressed package format for shipping the packages down to Kibana. Sadly this format is not supported directly by Windows. It would be nice if we would move to a zip file format this would allow anyone to download the package on windows unzip it and inspect the content.

@ruflin I vaguely remember you mentioning that there were some issues in Kibana that prevent us from moving to ZIP.

ruflin:
We had zip support in the past but it was removed here: https://github.com/elastic/package-registry/pull/59 @jfsiii might chime in what works better and why on the Kibana side.

With the work @mtojek is doing on https://github.com/elastic/package-registry/pull/472 it will be pretty simple to support both. If we support both, we need to discuss if we should adjust the `/search` API output:

Example:
```
  {
    "description": "aws Integration",
    "download": "/epr/aws/aws-0.0.3.tar.gz",
    "icons": [
      {
        "src": "/package/aws/0.0.3/img/logo_aws.svg",
        "title": "logo aws",
        "size": "32x32",
        "type": "image/svg+xml"
      }
    ],
    "name": "aws",
    "path": "/package/aws/0.0.3",
    "title": "aws",
    "type": "integration",
    "version": "0.0.3"
  },
```

jfsiii:
I'd like to support both/multiple. `tar.gz` makes things easier on the Manager if it's on darwin or linux. `.zip` would be great for those on Windows.

I'm not sure about the specific API. Some options that come to mind are:

  * add a query param like `archive_format=zip` with a default and the `download` key would remain a `string`. 
  * move to an array of objects like `icons`

ruflin:
@jfsiii If you are thinking of an array, are you thinking of one with objects inside? If we go this path, it would allow us to add "metadata" to each download link potentially if needed so Kibana for example does not have to take the `suffix` to decide which download link to use but could do it on a specific field that tells it, it is a `tar.gz`.

ruflin:
I opened https://github.com/elastic/package-registry/pull/512 as a draft to get some initial feedback.

roncohen:
we should only support one format. We agreed that unless there's big complications, it should be ZIP. It's natively support in MacOS and Windows and Linux users will be able to manage.

ph:
Adding this to b2, we need to do this before we go out of beta.

ruflin:
[...]

PR #512, opened 2020-06-15
Add list of downloads to /search endpoint

ruflin:
As discussed in https://github.com/elastic/package-registry/issues/474 in the future we might have more then one format to download a package. To be prepared for this, downloads becomes an array.

At the moment download and downloads is available but as soon as Kibana switched over, the old endpoint should be removed.

### DIFF
--- a/docs/api/search-all.json
+++ b/docs/api/search-all.json
@@ -1,5 +1,16 @@
 [
   {
+<<<<<<< HEAD
+=======
+    "description": "Package with data sources",
+    "download": "/epr/datasources/datasources-1.0.0.tar.gz",
+    "downloads": [
+      {
+        "path": "/epr/datasources/datasources-1.0.0.tar.gz",
+        "type": "tar.gz"
+      }
+    ],
+>>>>>>> f5fea69e... Add list of downloads to /search endpoint
     "name": "datasources",
     "title": "Default datasource Integration",
     "version": "1.0.0",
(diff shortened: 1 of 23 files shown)


docs/api/search-all.json
```diff
@@ -11,6 +17,12 @@
   {
     "description": "Tests if no pipeline is set, it defaults to the default one",
     "download": "/epr/default-pipeline/default-pipeline-0.0.2.tar.gz",
+    "downloads": [
```
ruflin, line 20: @jfsiii WDYT about this format?

ph: Could we use the type as the key? Unless I am mistaken this should not be possible to have two artifacts of type tar.gz for the same thing?

ph: I am putting myself into a user would want to use curl and jq to fetch the artifact.

ruflin: I think there is a future with multiple artifacts for the same time. Lets assume we add quick intro movies to the packages in the future. By default, these would not be served in the package but directly from the CDN. But there should be a url for example `-full.tar.gz` that will also include the movie files. Same for some massive data, by default it would be streamed from the CDN, but it could also be downloaded in the package.

But I'm not happy with calling it `type` and the value `tar.gz` also seems odd. What I think is important is the Kibana can easily pick the right download link. Perhaps we are back at naming the downloads with a key? @jfsiii thoughts?

ph: I see your point for movies or other related artifacts.
I would say type is **OK**, we can version the API later on, iteration over perfection.

ruflin: @jfsiii If I get your comment right, you are good with moving forward with what we have in this PR but switch `type: tar.gz` to `type: tar`?

ruflin:
@jfsiii I would like to get your attention on comment https://github.com/elastic/package-registry/pull/512#discussion_r441375749 before I make an update. Let me know what you think works best.

docs/api/search-all.json
jfsiii: @ruflin Correct. Let's follow the `tar` example from https://artifacts-api.elastic.co and we can add `mime_type: "application/gzip"` or any additional info if we need it

ruflin: Adjusted.

ruflin:
@jfsiii I adjusted the PR. I suggest we get this in an follow up on the Kibana side to switch over to downloads. Then we remove the old `downloads` part from the registry.

@mtojek Are you ok with getting this in?

ruflin:
I filed as a follow up on the Kibana side: https://github.com/elastic/kibana/issues/69775

Issue #466, opened 2020-05-15
Encode Kibana dashboards


mtojek:
The `mage build` step for the package-storage project should encode Kibana dashboards. 

Files in this form will be stored in the package-storage. The package-storage will validate PRs to check if the format is consistent.

exekias:
Would this mean that package-storage dashboards look different than the ones in integrations? What's the reason to do that?

mtojek:
Let me explain starting from the end of the flow.

Kibana accepts objects (dashboards, visualizations, etc.) partially encoded.
The package-registry will not transform or modify any package content (no `mage build` anymore).
The package-storage will only validate package content (check if dashboards are partially encoded).
Here we come to the Integrations repository. We can either keep there Kibana files in the same form, encoded JSON for some fields, or decode them all. If we decode some structures and Git-version them, it's easier to review JSON files.

Actually, this is up to the Integrations team to choose between readability and consistency and decide which one fits better.

jsoriano:
> The reason why I do not care much about having decoded dashboards is that we've never edited them manually, in 98% cases we edit them using Kibana designer. This 2% covers small fixes, typos, etc. In general the structure of a dashboard is a mystery (all meta properties, doubled fields, etc.). Even if I edit the dashboard manually, I still have to import it to Kibana to verify.

In beats we have experienced both worlds, dashboards were not decoded first, and since some time ago we decode them (https://github.com/elastic/beats/pull/7224). In my opinion this was an important improvement in dashboards maintenance. Specially because it made much easier or even possible to make small fixes and to review small and not so small changes. Also because it makes the mysterious structure of the dashboards much more approachable :)

Something similar happened with pipelines, originally we had all of them in JSON, and supporting yaml (https://github.com/elastic/beats/pull/11209) was a big improvement for developing and reviewing some of the more complex ones. Elasticsearch only supports pipelines in JSON format, but we use yaml for development, and the tooling (filebeat setup in this case) takes care of the conversions. By the way, what takes care of converting pipelines from yaml to json in packages?

> I do not insist on this approach, I can live with decoded objects for sure :)

I'd want to remark that my point is not only about decoded vs encoded objects, it is also that I think that having a build step doesn't look unexpected, and we are probably going to need it in any case at some moment. And given that, why doing sacrifices in maintainability?

> * During daily development and maintenance a developer would need to run `build` step every time, they would like to reflect changes to a format compatible with package-registry.

[...]

Issue #505, opened 2020-06-10
Allow for multiple paths in "publicDir"


mtojek:
The aim of this task is to modify the Package Registry to allow for multiple different paths, similarly to the system PATH environment variable.

Justification:
When mounting the `/registry/public` directory, we need to prepare the source point:
1. Fetch package-storage.
2. Add the interesting ones (e.g. an integration a developer is currently working on)
3. Build all packages.
4. Mount the directory.

It would be much simpler, if the developer needs to prepare an additional custom public directory with built only single integration. No need for visiting package-storage if the target Docker image for registry already contains mandatory packages.

The registry would update version conflicts, the latter one in `publicDir` overrides previous revisions.

ruflin:
Assuming that we get completely rid of `mage build` and the registry can just read from directories directly, we could probably also get rid of the `public` name here as it is just `packagePaths`.

mtojek:
I'm going with this one.

PR #513, opened 2020-06-16
Do not use generator

mtojek:
Issue: https://github.com/elastic/package-registry/issues/466#issuecomment-642022815

Changes:
* remove generator
* copy resources from package paths (as there is `example`)
* encode all present Kibana dashboards


magefile.go
```diff
@@ -88,6 +87,73 @@ func fetchPackageStorage() error {
 		packageStorageRevision)
 }
 
+func buildPackages() error {
```
ruflin, line 90: Looking forward for this one to also disappear!

mtojek: Yes, we have a plan for this one: https://github.com/elastic/package-registry/issues/505

mtojek:
> LGMT!
> 
> Changelog 😆

This will result in a penalty action item to teach CI verifying this...

PR #515, opened 2020-06-17
Change versioning lib

ruflin:
The reason for the change is that the library from Mastermind also supports syntax like `~1.2.3`. This simplifies specifying the constraints.

Further changes to the versions are coming, but I wanted to first change the library to have this separate.


ruflin:
@kuisathaverat This build failed because of some junit reporting error?

kuisathaverat:
there are 4 test failing

PR #516, opened 2020-06-17
Remove support for Elasticsearch requirement

ruflin:
So far I have not seen a package that would need to specify requirements for an Elasticsearch version as the Elasticsearch and Kibana version are tied together in the minor. Because of this removing support for Elasticsearch requirements for now. We can readd it later if we need.

This also cleans up the existing packages to remove the mention of Elasticsearch and some other examples that were still around.

As a follow up, integrations packages should be updated to get rid of the requirement. But more changes will happen to the requirements after https://github.com/elastic/package-registry/pull/515 is merged so perhaps we rather wait with this a bit more.

This change should not break any existing packages.

### DIFF
--- a/ASSETS.md
+++ b/ASSETS.md
@@ -35,12 +35,8 @@ compatibility: [1.0.2, 2.0.1]
 os.platform: [darwin, freebsd, linux, macos, openbsd, windows]
 
 requirement:
-  elasticsearch:
-    versions: >7.0
   kibana:
     versions: >7.0
-  agent:
-    versions: >7.1
 
 # The order of the items listed here is the order they show up in the package overview.
 screenshots:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Change stream.* fields to dataset.* fields. [#492](https://github.com/elastic/package-registry/pull/492)
 * Remove `solution` entry support in package manfiest. [#504](https://github.com/elastic/package-registry/pull/504)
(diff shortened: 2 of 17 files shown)


ruflin:
@jonathan-buttner This will probably also need an update to the Endpoint package.

mtojek:
I think you should rollout this the other way round. First, make sure nobody uses/publishes the field, then remove it.

What about existing packages? I think @jonathan-buttner raised a similar issue a day ago (backwards incompatible change).

ruflin:
As the field has not effects I would consider this "nobody is using it" ;-) As it doesn't break any existing packages, I think it is fine to get this in first.

For the existing packages already released, I would not touch them as additional data can be there but will not be used.

mtojek:
> As the field has not effects I would consider this "nobody is using it" ;-) As it doesn't break any existing packages, I think it is fine to get this in first.

Usage:
https://github.com/elastic/integrations/blob/master/dev/import-beats/requirement.go#L30

> For the existing packages already released, I would not touch them as additional data can be there but will not be used.

I agree with this approach :)

ruflin:
@mtojek Will you take care of the integrations usage or should I open a PR there?

mtojek:
> @mtojek Will you take care of the integrations usage or should I open a PR there?

Sure, I can help here.

ruflin:
@mtojek Thanks

jonathan-buttner:
> @jonathan-buttner This will probably also need an update to the Endpoint package.

Thanks! I'll make sure our repo is updated 👍

ruflin:
@mtojek as https://github.com/elastic/integrations/pull/92 is merged, ok to get this one in?

Issue #517, opened 2020-06-17
[Meta] Planned changes to the manifest format


ruflin:
A few changes are planned to the manifest(s) format. This is to track these changes:

# Rename `datasources` to `config_templates` 

We remove the term datasources across the different components. This now also has to be removed in the packages and be renamed to config_templates:

  * [x] Update Kibana https://github.com/elastic/kibana/pull/69864
  * [x] Update registry https://github.com/elastic/package-registry/pull/514  https://github.com/elastic/package-registry/pull/570
  * [x] Packages https://github.com/elastic/package-storage/pull/94
  * [x] Integrations https://github.com/elastic/integrations/pull/132

# Cleanup terms

Cleanup terms used inside the registry.
* [x] Cleanup dataset terms https://github.com/elastic/package-registry/pull/514

# Requirements to constraints

We currently use requirements section. To align with the agent and other parts, it should be renamed to constraints. In line with this task, also the default dependencies in the packages should be udpated.

* [x] Update Registry  https://github.com/elastic/package-registry/pull/519
* [x] Update Integrations https://github.com/elastic/integrations/pull/133
* [x] Update Package storage  https://github.com/elastic/package-storage/pull/95
* [x] Remove legacy part in registry https://github.com/elastic/package-registry/pull/580
* [x] Introduce new default for all packages (^7.9.0) https://github.com/elastic/integrations/issues/151

# Update categories

* [x] Implement new categories in package-registry https://github.com/elastic/package-registry/pull/534
* [x] Remove existing categories in package-storage https://github.com/elastic/package-storage/pull/97
* [x] Update packages with new categories in integrations and release packages https://github.com/elastic/integrations/issues/135
* [x] Remove legacy package information in registry https://github.com/elastic/package-registry/pull/571

# Directory names cleanup

Some of the directory names contained `-`, others `_`. This also had an effect on how the config options are called and what paths users have to define. To only have one way, it was decided to use `_` everywhere and packages need to be adjusted.

* [x] Update directory names to be _ instead of - https://github.com/elastic/package-registry/issues/565
* [x] Update packages to also use the new name https://github.com/elastic/package-storage/pull/90
* [x] Change Kibana https://github.com/elastic/kibana/pull/70320
* [x] Update packages to only use the new names
* [x] Update integrations repo https://github.com/elastic/integrations/issues/150
* [ ] Validate all paths to not contain `-` if directory. https://github.com/elastic/package-registry/pull/572

# Ingest pipeline config option

[...]

PR #519, opened 2020-06-17
Add conditions as future replacement of requirements 

ruflin:
This change adds `conditions` as part of the package manifest to align with other parts in the stack. Currently it adds the future of conditions and does not remove requirements yet. Requirements will be removed as soon as packages are updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add list of downloads to /search endpoint. [#512](https://github.com/elastic/package-registry/pull/512)
 * Apply rule: first package found served. [#546](https://github.com/elastic/package-registry/pull/546)
 * Implement package watcher. [#553](https://github.com/elastic/package-registry/pull/553)
+* Add conditions as future replacement of requirements. [#519](https://github.com/elastic/package-registry/pull/519)
 
 ### Deprecated
 
--- a/testdata/generated/package/multiversion/1.1.0/index.json
+++ b/testdata/generated/package/multiversion/1.1.0/index.json
@@ -26,13 +26,13 @@
     "metrics"
   ],
   "release": "ga",
+  "conditions": {
+    "kibana.version": "\u003e6.7.0"
+  },
   "requirement": {
     "kibana": {
       "versions": "\u003e6.7.0"
     }
-  },
-  "conditions": {
-    "kibana.version": "\u003e6.7.0"
   },
   "assets": [
     "/package/multiversion/1.1.0/changelog.yml",
--- a/testdata/generated/package/multiversion/1.1.0/index.json
+++ b/testdata/generated/package/multiversion/1.1.0/index.json
@@ -31,6 +31,9 @@
       "versions": "\u003e6.7.0"
     }
   },
+  "conditions": {
+    "kibana.version": "\u003e6.7.0"
+  },
   "assets": [
     "/package/multiversion/1.1.0/changelog.yml",
     "/package/multiversion/1.1.0/manifest.yml",
--- a/testdata/package/example/1.0.0/manifest.yml
+++ b/testdata/package/example/1.0.0/manifest.yml
@@ -10,6 +10,9 @@ release: ga
 
 owner.github: "ruflin"
 
+conditions:
+  kibana.version: "~7.x.x"
+
 requirement:
   kibana:
     versions: ">=7.0.0"
--- a/util/package.go
+++ b/util/package.go
@@ -60,6 +60,7 @@ type Package struct {
 	versionSemVer   *semver.Version
 	Categories      []string         `config:"categories" json:"categories"`
 	Release         string           `config:"release,omitempty" json:"release,omitempty"`
+	Conditions      *Conditions      `config:"conditions,omitempty" json:"conditions,omitempty" yaml:"conditions,omitempty"`
 	Requirement     Requirement      `config:"requirement" json:"requirement"`
 	Screenshots     []Image          `config:"screenshots,omitempty" json:"screenshots,omitempty" yaml:"screenshots,omitempty"`
 	Assets          []string         `config:"assets,omitempty" json:"assets,omitempty" yaml:"assets,omitempty"`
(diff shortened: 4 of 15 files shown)


ruflin:
Throwing this in here as I found some additional conditions in an old file:
```
compatibility: [1.0.2, 2.0.1]
os.platform: [darwin, freebsd, linux, macos, openbsd, windows]
```

But these are probably more on the dataset level.

testdata/package/example/1.0.0/manifest.yml
```diff
@@ -10,6 +10,9 @@ release: ga
 
 owner.github: "ruflin"
 
+conditions:
```
ruflin, line 13: @ph @michalpristas I initially was thinking about aligning the syntax with the constraints on the agent. But it only complicates things. The reason is that the available conditions here are very limited and have very specific checks. Adding additional "non yaml" syntax only complicates things. So here is what I'm proposing at the moment.

As each `condition` can only be used once I made it a key/value pair instead of an array.

One thing I think we should align is the naming. I landed on conditions as I think it fits best in all the places. WDYT about renaming it also in the agent?

ph: +1 on using "conditions" instead of constraints, which would align with dynamic input.

Now, I am Okayish with using key values pair, for what you are proposing.

1. Its a limited set of fixed conditions.
2. The conditions are joined with an **AND** this mean all condition predicates must return true.

ph: Also, using key/value also mean you don't have to have the parser in Kibana and we don't have to share code in the registry. 

So since everything is defined upfront I do not see a concern if the syntax arent the same, it's easier to start fixed and limited than to be dynamic all the way.

ruflin:
Marked this as ready for review but it seems currently Github is flaky and has not picked up the most recent commits that I pushed :-(

testdata/generated/package/multiversion/1.1.0/index.json
```diff
@@ -31,6 +31,9 @@
       "versions": "\u003e6.7.0"
     }
   },
+  "conditions": {
+    "kibana.version": "\u003e6.7.0"
```
mtojek, line 35: I wonder why some fields are in such encoded format and some are not: `kibana.version: ">6.7.0  <7.6.0"`

util/package.go
```diff
@@ -97,6 +98,11 @@ type Requirement struct {
 	Kibana ProductRequirement `config:"kibana" json:"kibana,omitempty" yaml:"kibana"`
 }
 
+type Conditions struct {
+	KibanaVersion string `config:"kibana.version,omitempty" json:"kibana.version,omitempty" yaml:"kibana.version,omitempty"`
+	kibanaVersion *semver.Constraints
```
mtojek, line 103: I suggest to rename this field, the difference is in capitals. How about `kibanaConstraint` or `kibanaSemver`?

mtojek:
Is there any meta-issue to link this PR?

testdata/generated/package/multiversion/1.1.0/index.json
ruflin: The "problem" is the `>` char which gets converted. But I tested this quite some time ago and it works as expected even though it is converted.

ruflin:
Meta issue tracking this change is in https://github.com/elastic/package-registry/issues/517

util/package.go
ruflin: Renamed to `kibanaConstraint`.

ruflin:
@mtojek PR updated with the reviews. Could you take a look again?

util/package.go
```diff
@@ -185,8 +191,18 @@ func NewPackage(basePath string) (*Package, error) {
 
 	p.Downloads = []Download{NewDownload(*p, "tar")}
 
-	if p.Requirement.Kibana.Versions != "" {
-		p.Requirement.Kibana.semVerRange, err = semver.NewConstraint(p.Requirement.Kibana.Versions)
+	// If the new conditions are used, select them over the requirements
+	if p.Conditions != nil && p.Conditions.KibanaVersion != "" {
+		p.Conditions.kibanaConstraint, err = semver.NewConstraint(p.Conditions.KibanaVersion)
+		if err != nil {
+			return nil, errors.Wrapf(err, "invalid Kibana versions range: %s", p.Requirement.Kibana.Versions)
```
ycombinator, line 198: I think this should be `p.Conditions.KibanaVersion`?

util/package.go
```diff
@@ -185,8 +191,18 @@ func NewPackage(basePath string) (*Package, error) {
 
 	p.Downloads = []Download{NewDownload(*p, "tar")}
 
-	if p.Requirement.Kibana.Versions != "" {
-		p.Requirement.Kibana.semVerRange, err = semver.NewConstraint(p.Requirement.Kibana.Versions)
+	// If the new conditions are used, select them over the requirements
```
ycombinator, line 194: There is duplicated code in this if-else to account for the new and legacy settings. Looks like the intent is to use the new setting, `p.Conditions.KibanaVersion`. So why not do something like this instead?

```
p.setKibanaVersion()
if p.Conditions != nil && p.Conditions.KibanaVersion != "" {
   ...
}
```

```
func (p *Package) setKibanaVersion() {
  if p.Conditions != nil && p.Conditions.KibanaVersion != "" {
    return
  }

  if p.Requirement.Kibana.Versions != "" {
    if p.Conditions == nil {
      // initialize p.Conditions
    }
    p.Conditions.KibanaVersion = p.Requirement.Kibana.Versions
  }
}    
```

ruflin: @ycombinator The legacy code should only be around for 24-48h, so I was sloppy.

util/package.go
ruflin: fixed

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add list of downloads to /search endpoint. [#512](https://github.com/elastic/package-registry/pull/512)
 * Apply rule: first package found served. [#546](https://github.com/elastic/package-registry/pull/546)
 * Implement package watcher. [#553](https://github.com/elastic/package-registry/pull/553)
+* Add conditions as future replacement of requirements. [#519](https://github.com/elastic/package-registry/pull/519)
 
 ### Deprecated
```
ycombinator, line 45: Should requirements be deprecated?

ruflin: I will be removed in a few hours ;-)

ph:
post merge LGTM?

PR #520, opened 2020-06-18
Support validation ECS style dataset format

jonathan-buttner:
Once the dataset fields are included in ecs the yaml file will look like this:

```
---
- name: dataset
  title: dataset
  group: 2
  short: Fields describing the new indexing strategy
  description: >
    Fields describing the new indexing strategy <type>-<name>-<namespace>
  type: group
  fields:
    - name: type
      level: custom
      type: constant_keyword
      description: >
        Dataset type.

    - name: name
      level: custom
      type: constant_keyword
      description: >
        Dataset name.

    - name: namespace
      level: custom
      type: constant_keyword
      description: >
        Dataset namespace.
```

This results in the following `fields.yml`:
```
- name: dataset
  title: dataset
  group: 2
  description: Fields describing the new indexing strategy <type>-<name>-<namespace>
  type: group
  fields:
  - name: name
    level: custom
    type: constant_keyword
    description: Dataset name.
    default_field: false
  - name: namespace
    level: custom
    type: constant_keyword
    description: Dataset namespace.
    default_field: false
  - name: type
    level: custom
    type: constant_keyword
    description: Dataset type.
    default_field: false
- name: destination
  title: Destination
  group: 2
  description: 'Destination fields describe details about the destination of a packet/event.

    Destination fields are usually populated in conjunction with source fields.'
  type: group
  fields:
  - name: address
<more fields>
```

This will fail validation because we're searching for `dataset.*`.

This PR adds support for splitting the `dataset.*` string and looking inside the map. It finds `name: dataset` first and then pulls out the array `fields` and looks for the last part of the string (e.g. `namespace` or `type`).

This implementation is suupeeeer slow. So if you have better ideas that'd be great.

Basically I'd like to define the `dataset.*` fields in ecs style schema instead of using:

```
- name: dataset.type
  type: constant_keyword
  description: >
    Dataset type.
- name: dataset.name
  type: constant_keyword
  description: >
    Dataset name.
- name: dataset.namespace
  type: constant_keyword
  description: >
    Dataset namespace.
```

It doesn't look like this line: https://github.com/elastic/package-registry/blob/master/util/dataset.go#L282

Is flattening it like the way I hoped. It still results in:

```
{
  name: dataset
  fields: [
    {name: name, ...}, {name: type, ...}, {name: namespace, ...}
  ]
}
```

PR #521, opened 2020-06-18
Fix: skip fields level while validating

mtojek:
A bit different implementation of https://github.com/elastic/package-registry/pull/520 with test coverage.

### DIFF
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -277,17 +277,30 @@ func (d *DataSet) validateRequiredFields() error {
 		return errors.Wrapf(err, "walking through fields files failed")
 	}
 
-	// Flatten all fields
-	for i, fields := range allFields {
-		allFields[i] = fields.Flatten()
+	err = validateRequiredFieldsData(allFields)
+	if err != nil {
+		return errors.Wrapf(err, "validating required fields data failed")
 	}
+	return nil
+}
 
-	// Verify required keys
+func validateRequiredFieldsData(allFields []MapStr) error {
+	allFields = flattenFieldsData(allFields)
+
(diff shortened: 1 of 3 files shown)


mtojek:
> I really appreciate the tests!
> 
> I'm wondering if as a follow up we should also add two example yaml files to our testdata packages.

Hm.. we don't have any tests checking failed validation, do we? I think I will open an issue for this.

jonathan-buttner:
Thanks for taking a look @mtojek I think we're almost there.

I'm still getting this error:
```
Error: building dataset failed (path: /Users/jbuttner/proj/es/endpoint-package/out/packages/endpoint/0.4.0/dataset/alerts): error building dataset (path: alerts) in package: endpoint: validating required fields failed: validating required fields data failed: finding field failed (searchedName: dataset.type): field 'dataset.type' not found accessing config accessing config
make: *** [run-registry] Error 1
```

I think this is because `fields` is an array not a `MapStr`. When I printed the type I got this: `[]map[interface{}]interface{}]`

So the we'll need `tryToMapStr()` https://github.com/elastic/package-registry/blob/master/util/mapstr.go#L245 to handle `[]map[interface{}]interface{}`

The test should look like this:

```
func TestFlattenFieldsData_ObjectWithFields(t *testing.T) {
	data := []MapStr{
		{
			"dataset": MapStr{
				"fields": []MapStr{
					{
						"name": "name",
						"type": "constant_string",
					},
					{
						"name": "type",
						"type": "constant_string",
					},
				},
			},
		},
	}

	flattened := flattenFieldsData(data)
...
}
```

PR #524, opened 2020-06-18
Discuss: Remove public dir support

ruflin:
The public dir was needed in the past to build packages. But with the recent changes to the registry this has become obsolete. Instead the registry can be directly pointed at a directory with packages. In the package storage this directory is called `packages`, I suggest we use this as a default. The registry itself could have its packages in the same directory.

This PR is not complete, it is more to discuss the change.


handler.go
```diff
@@ -39,6 +40,7 @@ func catchAll(public http.FileSystem, cacheTime time.Duration) func(w http.Respo
 func determineResourcePath(r *http.Request, public http.FileSystem) (string, error) {
 	path := r.URL.Path
 
+	path = strings.Replace(path, "/package", "", 1)
```
ruflin, line 43: This change is odd but is needed as now all is at the top level

ruflin:
@mtojek I assume this also ties into your change you have planned for multiple paths?

mtojek:
Yes, it's draft currently: https://github.com/elastic/package-registry/pull/525

mtojek:
I planned to leave `PublicDir` for legacy purposes:

```golang
type Config struct {
	PublicDir           string        `config:"public_dir"` // left for legacy purposes
	PackagesPaths       []string      `config:"packages_paths"`
	CacheTimeSearch     time.Duration `config:"cache_time.search"`
	CacheTimeCategories time.Duration `config:"cache_time.categories"`
	CacheTimeCatchAll   time.Duration `config:"cache_time.catch_all"`
}
```

ruflin:
Directly closing this as https://github.com/elastic/package-registry/pull/525 already covers the idea from this PR .

@mtojek Lets try to get rid of all legacy before 7.9, we will build up enough over the next months :-)

PR #525, opened 2020-06-18
Support custom packages path

mtojek:
This PR add support for multiple paths with packages.

Issue: https://github.com/elastic/package-registry/issues/505


config.docker.yml
```diff
@@ -0,0 +1,7 @@
+public_dir: ./public
+packages_paths:
```
ruflin, line 2: I'm torn if we should call it `package_paths`?

magefile.go
```diff
@@ -33,15 +31,9 @@ var (
 	buildDir       = "./build"
 	storageRepoDir = filepath.Join(buildDir, "package-storage")
 	packagePaths   = []string{filepath.Join(storageRepoDir, "packages"), "./dev/packages/example/"}
-	tarGz          = true
 )
 
 func Build() error {
-	packagePathsEnv := os.Getenv("PACKAGE_PATHS")
```
ruflin, line 40: It would be nice to support this in the future directly in the registry to pass in an env variable for this.

main.go
```diff
@@ -37,7 +38,7 @@ var (
 	configPath = "config.yml"
 
 	defaultConfig = Config{
-		PublicDir:           "public",
+		PublicDir:           "./public", // left for legacy purposes
```
ruflin, line 41: Curious which legacy part you are referring to here? Do we need it somewhere?

mtojek: Every piece of software that mounts directory for packages, e.g. all testing environments where package path is mounted as `/public`.

magefile.go
mtojek: We can add it in the future, but the goal is to use the config file. If you prefer, we can create a new issue to read same configuration from environment variables, but not sure if there is a specific use case. In all cases I'm aware you can depend on the `config.yml` file.

config.docker.yml
mtojek: Renamed.

main.go
ruflin: I assume we should update these accordingly?

mtojek: No, it's for legacy purposes. Thanks for this line, I will work for configuration files defining `/public` directory.

The `package` directory is appended here: https://github.com/elastic/package-registry/blob/master/main.go#L126

ruflin: I think I miss what legacy purpose means in this context. All the registries and builds which are run are currently under our control so we can get rid of the `public` path there and remove it. But I assume I miss something?

My goal is that as soon as we hit 7.9, we don't have legacy code inside as we will build up more then enough in the future.

mtojek: It means that we need to remove all references to public from these places:
https://github.com/elastic/integrations/blob/master/testing/environments/snapshot.yml#L47
https://github.com/elastic/package-storage/blob/master/testing/environments/snapshot.yml#L43

and notify people to update their workplaces, because the package-registry will not support this anymore.

ruflin: I will open PR's to adjust these accordingly. I expect that all devs building packages at the moment update to the most recent changes at least with the updates of the snapshot build.

ruflin: I found the missing bit on my end. We don't provide any other default at the moment. My proposal would be to use `./packages` as default?

ruflin: I opened https://github.com/elastic/package-registry/pull/551 as an alternative proposal.

PR #526, opened 2020-06-18
Generate package index.json files for testing

ruflin:
We got rid of the generated files so we are able to get rid of the public directory and the build step. But on the testing side this removed one useful tool to show changes in our generated output. To get this back, I added a test that creates all index.json files for the testdata packages.

Doing this change I found also some generated files we don't use anymore. After this is merged, I want to clean up a bit on where we put our golden files and make sure we clean it up.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -122,14 +122,18 @@ func TestPackageIndex(t *testing.T) {
 
 // TestAllPackageIndex generates and compares all index.json files for the test packages
 func TestAllPackageIndex(t *testing.T) {
-	publicPath := "./testdata/public"
-	packagesBasePath := []string{publicPath + "/package"}
+	testPackagePath := filepath.Join("testdata", "package")
 
+	packagesBasePath := []string{testPackagePath}
 	packageIndexHandler := packageIndexHandler(packagesBasePath, testCacheTime)
 
 	// find all packages
-	dirs, err := filepath.Glob(packagesBasePath[0] + "/*/*")
-	assert.NoError(t, err)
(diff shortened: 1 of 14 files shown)


main_test.go
```diff
@@ -121,6 +123,40 @@ func TestPackageIndex(t *testing.T) {
 	}
 }
 
+// TestAllPackageIndex generates and compares all index.json files for the test packages
+func TestAllPackageIndex(t *testing.T) {
+	publicPath := "./testdata/public"
+	packagesBasePath := publicPath + "/package"
```
mtojek, line 129: I think you will have to adjust this according to the latest changes in master.

ruflin:
@mtojek Could you have a look again?

main_test.go
```diff
@@ -118,6 +120,40 @@ func TestPackageIndex(t *testing.T) {
 	}
 }
 
+// TestAllPackageIndex generates and compares all index.json files for the test packages
+func TestAllPackageIndex(t *testing.T) {
+	publicPath := "./testdata/public"
```
mtojek, line 125: Hm, does it still work like that? There should be no `public` directory anymore: https://github.com/ruflin/package-registry/tree/json-files/testdata . As there is no need for building, you should be good to go with sth like this:

```golang
packagesBasePaths := []string{"./testdata/package"}
```

ruflin: ups, I'm glad you spotted this. It worked because the dir was just empty so no files were generated and the old files were already there :-( Fixed now.

PR #528, opened 2020-06-18
Adding ecs style dataset package

jonathan-buttner:
Adding an example package for checking the ecs style dataset format.


util/package.go
```diff
@@ -333,7 +333,11 @@ func (p *Package) validateVersionConsistency() error {
 	}
 
 	if !versionPackage.Equal(versionDir) {
-		return fmt.Errorf("inconsistent versions (path: %s, manifest: %s)", versionDir.String(), p.versionSemVer.String())
+		manifestVer := "unknown"
```
jonathan-buttner, line 336: If the package version is off then `p.versionSemVer` will be nil. This causes a panic in the print below when the `p.versionSemVer.String()` is called.

ruflin: by off you mean not specified? I hoped this would even bubble up before we validate the version.

ruflin:
To update generated files you need to run `mage check ` and `go test . -generate`

util/package.go
jonathan-buttner: Sorry I should have been more clear. No the version is specified in the `manifest.yml` file but the directory name is different. So my example package had directory `0.0.1` but manifest version `0.0.2`. Not sure why it didn't populate the version field correctly though.

ruflin: Seems like you proofed that it works :-D +1 on error instead of panic.

PR #529, opened 2020-06-22
Use BasePackage for search output data

ruflin:
So far the `/search` output was manually built. This had the risk that the format of the package output and search output diverge but they should not. To prevent this, a BasePackage is introduced. The package is composed out of this BasePackage with additional fields.

This change does not affect the output, only the order of the output.

### DIFF
--- a/search.go
+++ b/search.go
@@ -163,28 +163,12 @@ func getPackageOutput(packagesList map[string]map[string]util.Package) ([]byte, 
 	}
 	sort.Strings(keys)
 
-	var output []map[string]interface{}
+	var output []util.BasePackage
 
 	for _, k := range keys {
 		parts := strings.Split(k, separator)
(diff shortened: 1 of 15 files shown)


search.go
```diff
@@ -163,28 +163,12 @@ func getPackageOutput(packagesList map[string]map[string]util.Package) ([]byte,
 	}
 	sort.Strings(keys)
 
-	var output []map[string]interface{}
+	var output []util.BasePackage
 
 	for _, k := range keys {
 		parts := strings.Split(k, separator)
 		m := packagesList[parts[0]][parts[1]]
```
mtojek, line 170: nit: is there is a need to check the number of parts here (in case if it's empty)?

ruflin: I also stumbled over this line when I did the change and I plan to follow up to make this nicer. There must be a better way here.

Issue #455, opened 2020-05-13
Discuss: Should registry remove 'removable' property?


jfsiii:
The packages on the registry are shipped with a `removable` property which Kibana uses to determine if the package is required / can be removed.

This feels like a Kibana concern. Why should the registry or package owner have this control over Kibana?

This also makes validation very important so that a package, especially a malicious one, isn't marked as unremovable.

I would like to explore storing this in Kibana. Is there a good reason to do it in the registry?

ruflin:
The advantage of having it in the package is that we can decide it `ad-hoc`. Meaning someone already using 7.8 and we ship now a package later where `removable: false`, we would not need to change Kibana. But it feels like a made up case which will not happen in reality. Why should a package when 7.8 shipped have been removable but not later on.

+1 on my end to store this in KB instead of the package. Also makes sure contributors don't mess around with this flag. But we will need a way down the road that other Kibana plugins can hook into this to specify also other packages, for example endpoint to not be removable.

Where did we land with the required packages? Is this specified in Kibana?

ruflin:
@jfsiii @paul-tavares @jonathan-buttner  I would like to move forward on this and store it in Kibana instaed of the package. I assume changes in KB would need to happen first?

jonathan-buttner:
> @jfsiii @paul-tavares @jonathan-buttner I would like to move forward on this and store it in Kibana instaed of the package. I assume changes in KB would need to happen first?

Moving it to Kibana sounds good to me. Yeah we'll probably want the changes to be in Kibana first. I don't think it matters for 7.8 for the endpoint package but would removing the `removable` field from the system package allow a user on 7.8 to remove the system package? Is that a concern?

paul-tavares:
I agree that this should be done in Kibana. Our (Endpoint) concern was/is that the Endpoint package should be a) installed by default, and b) can NOT be uninstalled, so regardless of approach, if that can be provide we're good 😃

ruflin:
@jonathan-buttner In 7.8 only the packages we have already built are available and also package-registry version will stay the same. I also assume we will not touch 7.8 code anymore, so this discussion only applies to 7.9.

@paul-tavares Could you take a lead on getting this into Kibana so we can remove it afterwards in the registry?

jfsiii:
@paul-tavares Ingest can prevent users from using its API or UI to uninstall a package, but that's it. With the right queries/permissions a user could still remove the dashboards, etc from ES

[...]

Issue #204, opened 2020-01-29
Versioning of the package registry API


ruflin:
Currently there is no versioning of the package registry API. So far the API is pretty small but we need to decide if we should version it from day one or tackle the issue when have our first breaking changes.

jfsiii:
From day one, or as early as possible, IMO.

We know the changes will occur; we might as well have a plan for how to handle it. We can default to v1/whatever to reduce burden on the user, but we should have some way to support versions.

ruflin:
Labeling this `beta1` as we need to come to a conclusion by then if we do it or not.

jfsiii:
I think https://apisyouwonthate.com/blog/api-versioning-has-no-right-way does a good job summarizing our options and the tradeoffs but most results for "API versioning strategy" will give a similar list.

We're already doing Global API Versioning using subdomains (`https://epr.elastic.co/`, `https://epr-experimental.elastic.co/`) and #560 explores the `/v1` path-based approach (but only for some routes).

It's also possible we could expand on the OpenAPI work started in https://github.com/elastic/package-registry/pull/490 and go for the ['commit to contracts'](https://apisyouwonthate.com/blog/commit-to-api-contracts) approach linked in that article. That's more flexible with less churn but would take longer to setup than global versioning.

I think my "anything is possible" ranked preferences are

1. Contracts via OpenAPI
2. global versioning via subdomain
3. global versioning via path
4. resource versioning

but different subdomains is very attractive and might actually be my first choice. Especially if we combine it with (2) by, say, publishing the OpenAPI spec from a known route on each domain.

jfsiii:
After the challenges listed at https://github.com/elastic/package-registry/pull/560#issuecomment-651955920 it seems relevant to mention the option of not using versions and only making backwards compatible changes

There's a good article on it at  https://apisyouwonthate.com/blog/api-evolution-for-rest-http-apis

This is the basic approach I've used at a few jobs. It puts more burden the API authors but keeps things simple for clients and consumers.

ruflin:
I closed https://github.com/elastic/package-registry/pull/560 for now and we stick with what we have at the moment.

I think it will be important that we properly use the manifest file `format_version` to indicate changes and also always send the Kibana version in the headers of all requests. This will make sure we could also make decision on this Kibana version if needed. Note: This does not replace the query param to the /search endpoint @ph Could you follow up on this.

jfsiii:
> and also always send the Kibana version in the headers of all requests

[...]

PR #532, opened 2020-06-22
Remove `removable` flag from package manifest

ruflin:
Based on the discussion in https://github.com/elastic/package-registry/issues/455 it is not up to the package to decide which packages can be removed but Kibana. Because of this, the removable flag is removed from the registry and the packages.

Closing https://github.com/elastic/package-registry/issues/455


ruflin:
This PR should only be merged if changes made it into Kibana. @paul-tavares Could you ping us here when this happened?

mtojek:
Did you check if the flag is used in the integrations repository?

ruflin:
@mtojek It is for the system package. Al others are set to true (the default): https://github.com/elastic/integrations/blob/0ae2f28fa667952f1b8c682ebec4163bbfe3bef6/dev/import-beats/packages.go#L26 I'll follow up on the integrations repo as soon as we get this in, but first work must happen in Kibana.

jfsiii:
https://github.com/elastic/kibana/pull/69761 is ready but without https://github.com/elastic/package-registry/issues/204 & the corresponding Kibana change we can't push changes independently

ruflin:
@jfsiii Not sure I follow why https://github.com/elastic/kibana/pull/69761 can't be merged as it only removes a field and I assume if the registry still ships it, it should not affect the Kibana?

If we would use the versioned API for every breaking change we do at the moment, we would already be at v25. This will of course change when we hit 7.9.

jfsiii:
@ruflin You're right https://github.com/elastic/kibana/pull/69761 can go in before this change, just not the other way around. I'll merge that PR now

> If we would use the versioned API for every breaking change we do at the moment, we would already be at v25.

Having _a_ strategy doesn't mean we need to use _that_ strategy but we can discuss that in the relevant issue

ruflin:
@jfsiii I kicked off https://github.com/elastic/package-registry/pull/560 related to the above.

ruflin:
Merging as https://github.com/elastic/kibana/pull/69906 is merged.

PR #534, opened 2020-06-22
Add additional categories

ruflin:
In the process of defining our integrations page, additional categories were defined: https://www.elastic.co/integrations This PR adds these categories to the registry.

It must be followed up to update all existing packages and then remove the old category definitions.

### DIFF
--- a/util/package.go
+++ b/util/package.go
@@ -26,7 +26,6 @@ var CategoryTitles = map[string]string{
 	"azure":             "Azure",
 	"cloud":             "Cloud",
 	"config_management": "Config management",
-	"container":         "Container",
 	"containers":        "Containers",
 	"crm":               "CRM",
 	"Custom":            "Custom",
--- a/util/package.go
+++ b/util/package.go
@@ -22,9 +22,35 @@ import (
 const defaultType = "integration"
 
 var CategoryTitles = map[string]string{
-	"logs":     "Logs",
-	"metrics":  "Metrics",
-	"security": "Security",
+	"aws":               "AWS",
+	"azure":             "Azure",
+	"cloud":             "Cloud",
+	"config_management": "Config management",
+	"container":         "Container",
+	"containers":        "Containers",
+	"crm":               "CRM",
+	"Custom":            "Custom",
+	"datastore":         "Datastore",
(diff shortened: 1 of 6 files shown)


ruflin:
@ph @mostlyjason @mtojek @andresrc The above means we should update all the packages with the related categories. Please reach out to @mostlyjason for the correct mapping.

@mostlyjason Can you validate the the category list is correct?

util/package.go
```diff
@@ -22,9 +22,35 @@ import (
...
 var CategoryTitles = map[string]string{
-	"logs":     "Logs",
-	"metrics":  "Metrics",
-	"security": "Security",
+	"aws":               "AWS",
+	"azure":             "Azure",
+	"cloud":             "Cloud",
+	"config_management": "Config management",
+	"container":         "Container",
+	"containers":        "Containers",
```
ruflin, line 30: @mostlyjason I stumbled over this one. I had somehow both, based on the integrations page I assume we only need `containers`?

ruflin: I removed it as I couldn't find it in the internal doc anymore.

mtojek:
@mostlyjason Please create an issue in the integrations repository with updated categories for packages. We can pick it up from there.

mostlyjason:
Thanks I left a comment on a separate thread discussing a new category field. I proposed that we move forward with this PR and come back later to update the categories if needed. Lets give a day or two to see if there are any objections.

util/package.go
```diff
@@ -22,9 +22,34 @@ import (
...
-	"logs":     "Logs",
-	"metrics":  "Metrics",
-	"security": "Security",
+	"aws":               "AWS",
+	"azure":             "Azure",
+	"cloud":             "Cloud",
+	"config_management": "Config management",
+	"containers":        "Containers",
+	"crm":               "CRM",
+	"Custom":            "Custom",
```
mostlyjason, line 31: all the others have a lowercase key name

ruflin: good catch, fixed.

ruflin:
@mostlyjason Merged this PR so we can move forward with the changes in the packages but happy to follow up if you find more issues in the category names.

Issue #451, opened 2020-05-12
Add a owner field in the package description


ph:
Currently it's hard to know which package belongs to which team. It would be nice to be able to know who is responsible for it: Platform, Service or SEIM. We could resolve that by either adding a top-level field `owner` or adding as metadata to the package.

Maybe the later is a better solution  because we do not know how or even if we will surface that to the end user.

cc @ruflin

ruflin:
I've been also thinking about an `author` or `contributors` field that exists in other package managers: https://docs.npmjs.com/files/package.json#people-fields-author-contributors It is a nice way to give "credit" to the creators and allow users to reach out to the. This will become especially relevant if we support community packages.

The two fields overlap in the beginning but might diverge. An author / contributor stays in, an owner might change. I wonder if we could mix the two.

Looking at the Google Play store, they have developer contact with a few fields as an example. We could do something like:

```
contact.website: discuss.*...
contact.github: @elastic/ingest-management
contact.name: ruflin 
contact.email: foo@bar.com
```

Fields are not required but could be filled in and are also part of the package. In our case, we can use it also for tracking the owner based on the Github entry which can be a single user or a group.

Not convinced about the naming yet :-(

ph:
Looking at rubygems which is similar to NodeJS and they do support **author** or **contributors**, I think is a more convenient and future proof to do  it. I can imagine tha author could be "Elastic" and the contributors you and I. I supposed when we open the registry we could have RBAC based on that and give contributors push access.

I would use something similar as this for each entry:

```
website: discuss.*...
source: https://github.com/elastic/...
name: ruflin 
email: foo@bar.com
```

ruflin:
What you mean for each entry? Could you share a full yaml on how it would look like?

ph:
```json
{
  "source": "http://github.com/elastic/mymodule",
  "website": "http://elastic.co",
  "author": {
    "name": "Elasticsearch",
    "email": "info@elastic.co"
  },
  "contributors": [
    {
      "name": "Ruflin",
      "email": "spam@elastic.co"
    },
    {
      "name": "PH",
      "email": "spam@elastic.co"
    }
  ]
}
```

Something like this, but this doesn't solve which team in Elastic own that module?
Do we allow a `_meta` fields for a package which we could use internally for toolings?

ruflin:
[...]

PR #536, opened 2020-06-22
Add support for owner filed in package manifest

ruflin:
In some cases it is important to know who the owner of a package is so this owner can be pinged for questions. In https://github.com/elastic/package-registry/issues/451 the different options around owner, authors, contributors were discussed. My take away from this discussions which lead to this PR:

* There are multiple authors over time
* Authors and contributors are mostly exchangeable terms
* There can only be 1 owner, even if it is a team.

This PR currently only implements `owner.github` because it is all we need. Later on we can add an array of `contributors` and addition owner fields if needed.

Closing https://github.com/elastic/package-registry/issues/451


ruflin:
@andresrc @mtojek If we move forward with this, I suggest we update the packages to have the owner team on Github inside this field.

mtojek:
To be honest, I'm against this change to prevent community people from pinging single developers instead of using team aliases. There is no plan for fixing this field if the person leaves the company and we won't fix already released packages.

I suggest to base on the repository owners here.

ruflin:
I used `owner.github` to exactly allow teams in here. For example `owner.github: elastic/integrations-platforms`.

dev/packages/example/reference/1.0.0/manifest.yml
```diff
@@ -28,6 +28,10 @@ license: basic
 # The default type is integration and will be set if empty.
 type: integration
 
+# Details about the owner of the package. Current only github link is supported.
+# A package can only have a single owner, the owner can be a user or a team.
```
ruflin, line 32: @mtojek I added a comment here to make it more obvious it can be a user or a team.

mtojek: Thanks for the explanation. I assume that team aliases will be set for all Elastic packages?

ruflin: Yes, I hope so. This was the initial driver for this in https://github.com/elastic/package-registry/issues/451 Not sure if there is a good way to add this to the generator as in Filebeat for example, some belong to SIEM, some to Platforms team.

ruflin:
After going through a few merge conflicts and rebases, this should now be ready again @mtojek

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Support multiple packages paths. [#525](https://github.com/elastic/package-registry/pull/525)
 * Added support for ecs style validation for dataset fields. [#520](https://github.com/elastic/package-registry/pull/520)
 * Use BasePackage for search output data. [#529](https://github.com/elastic/package-registry/pull/529)
+* Add support for owner filed in package manifest. [#535](https://github.com/elastic/package-registry/pull/535)
```
mtojek, line 35: nit: field

mtojek: nit: 536

ruflin: fixed. that is what happens if you try to predict the PR number ...

PR #538, opened 2020-06-22
Speed up build step

ruflin:
This speeds up the build and check mage command by not checking out all of the package-storage every time. Instead it checks if it already exists and if yes, resets it to the correct state.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -65,13 +65,15 @@ func Build() error {
 }
 
 func fetchPackageStorage() error {
-	err := os.RemoveAll(storageRepoDir)
-	if err != nil {
-		return err
-	}
 
-	err = sh.Run("git", "clone", "https://github.com/elastic/package-storage.git", storageRepoDir)
-	if err != nil {
+	// If storage directory does not exists, check it out
+	if _, err := os.Stat(storageRepoDir); os.IsNotExist(err) {
+		err = sh.Run("git", "clone", "--depth=1", "https://github.com/elastic/package-storage.git", storageRepoDir)
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		// catch other errors
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -65,26 +65,53 @@ func Build() error {
...
+		"--git-dir", filepath.Join(storageRepoDir, ".git"), "--work-tree", storageRepoDir,
+		"reset", "--hard")
 	if err != nil {
 		return err
 	}
 
-	err = sh.Run("git", "clone", "https://github.com/elastic/package-storage.git", storageRepoDir)
+	err = sh.Run("git",
+		"--git-dir", filepath.Join(storageRepoDir, ".git"), "--work-tree", storageRepoDir,
+		"clean", "-df")
```
mtojek, line 94: I'm not sure about this change. I used to fetch the repository to build directory and do not touch it. If I need to play with it, I would go to the package-storage. Do we need this adjustment?

magefile.go
```diff
@@ -65,26 +65,53 @@ func Build() error {
...
+	// If storage directory does not exists, check it out
+	if _, err := os.Stat(storageRepoDir); os.IsNotExist(err) {
+		err = sh.Run("git", "clone", "--depth=1", "https://github.com/elastic/package-storage.git", storageRepoDir)
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		// catch other errors
+		return err
+	}
```
mtojek, line 78: I suggest to not play with Git after cloning and simply return here.

magefile.go
ruflin: It's not a requirement. It just makes sure in case someone modified the build directory it will be reset / cleaned up.

magefile.go
ruflin: Agree, adjusted.

PR #540, opened 2020-06-23
Simplify directory name validation

ruflin:
The directory name / version validation had additional code inside because it was possible that the semver part was empty. By moving this to validate, this can not happen anymore.


ruflin:
@mtojek yes. The logic itself did not change. I only added a todo that we should reevaluate how we adjust this logic in a more predictable way.

PR #542, opened 2020-06-23
feat: add labels to the Docker images

kuisathaverat:
it adds labels to the Docker images, and change the namespace to push the PR Docker images to `docker.elastic.co/observability-ci/package-registry`, so now on `docker.elastic.co/package-registry/package-registry` there are only branches Docker images.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -11,6 +11,7 @@ pipeline {
     DOCKER_REGISTRY = 'docker.elastic.co'
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_IMG = "${env.DOCKER_REGISTRY}/package-registry/package-registry"
+    DOCKER_IMG_PR = "${env.DOCKER_REGISTRY}/observability-ci/package-registry"
   }
   options {
     timeout(time: 1, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -148,3 +154,25 @@ def cleanup(){
...
+  dir("${BASE_DIR}"){
+    dockerLogin(secret: "${env.DOCKER_REGISTRY_SECRET}",
+      registry: "${env.DOCKER_REGISTRY}")
+    sh(label: 'Build Docker image',
+      script: """docker build \
+        -t ${env.DOCKER_IMG_TAG} \
+        --label BRANCH_NAME=${env.BRANCH_NAME} \
+        --label GIT_SHA=${env.GIT_BASE_COMMIT} \
+        --label GO_VERSION=${env.GO_VERSION} \
+        --label TIMESTAMP=\$(date +%Y-%m-%d_%H:%M) \
```
ruflin, line 168: nice! I didn't release labels can be added without changing the Dockerfile itself.

kuisathaverat:
>We should probably clean up the old ones with the PR prefix?

I think we cannot delete images, I will as infra.

>Will it still build images for tags?

A tag is processed as a branch so yes it will generate a Docker image at `docker.elastic.co/package-registry/package-registry`

ruflin:
@kuisathaverat There is some magic to do it with some scripts as far as I remember, Infra will know.

PR #543, opened 2020-06-24
Introduce development mode

mtojek:
This PR introduce development mode. In this mode all package content is not served from memory cache.


mtojek:
> I wonder if we should set all `CacheTime*` variables to 0 if DevMode is enabled?

Not sure if it would be the expected outcome - ignore all CacheTime options in the configuration file, because the DevMode is enabled.

PR #551, opened 2020-06-24
Proposal: Introduce default path for packages

ruflin:
In the past, we had public as our default path for the build packages and things just worked out of the box. With removing the build step we got rid of public and introduced support for multiple package paths. In the package-storage we used `packages` as the directory name. I am proposing to use this as our recommended default. This means, if someone has a packages directory, the user can get started without having create a config file first. It also allows us to remove the legacy public directory.

Now more catch in this PR: I renamed the config to packages_paths from package_paths. This was the original name and I suggested to change it :facepalm: But now with having packages as the default this name probably makes more sense.

If we go forward with this, we need to update the different places where the public dir and package_paths is currently used.

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
@@ -39,7 +38,7 @@ var (
 	configPath = "config.yml"
 
 	defaultConfig = Config{
-		PublicDir:           "./public", // left for legacy purposes
+		PackagesPaths:       []string{"./packages"},
 		CacheTimeSearch:     10 * time.Minute,
 		CacheTimeCategories: 10 * time.Minute,
 		CacheTimeCatchAll:   10 * time.Minute,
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -39,7 +38,7 @@ var (
 	configPath = "config.yml"
 
 	defaultConfig = Config{
-		PublicDir:           "./public", // left for legacy purposes
+		PackagesPaths:       []string{"./packages"},
```
ruflin, line 41: @mtojek One reason I stumbled over making this the default is because of the reference config file :-)

ruflin:
@mtojek With our recent discussions around the Docker images and having `/packages/package-registry` as default path, I wonder if we should go a different direction. What if by default we apply `/packages/*` to fetch the array of directories with packages?  Then the user could either use this default or specify his own paths. The above would probably be its own config, something like `packages_path_pattern`. The nice part about this is that it removes all the multiple configs we use for the different branches at the moment.

ruflin:
Closing as this PR will not bring the benefits I initially hoped to simplify configuration. As soon as directories are automatically discovered, the user can't decide on the order anymore which is important for our production / staging / snapshot deployments.

PR #552, opened 2020-06-24
Add support for ES index-template configs

ruflin:
In some cases, the Elasticsearch index template for a dataset needs more configuration options then just the mappings that can be shipped as the fields.yml. To make this possible, this introduces the option to configure some settings in the dataset manifest. The naming is `elasticsearch.index-template` to be in line with the directory names. Inside settings and mappings are supported as free form. Anything else is ignored.

This is not supported yet on the Kibana side. There this should either be put into a separate component template or merge on creation.

### DIFF
--- a/testdata/package/reference/1.0.0/dataset/reference/manifest.yml
+++ b/testdata/package/reference/1.0.0/dataset/reference/manifest.yml
@@ -10,9 +10,9 @@ release: beta
 type: logs
 
 elasticsearch:
-  index-template.mappings:
+  index_template.mappings:
     dynamic: false
-  index-template.settings:
+  index_template.settings:
     index.lifecycle.name: reference
 
 # This is the ingest pipeline which should be used. If none is define,
--- a/testdata/package/reference/1.0.0/dataset/reference/manifest.yml
+++ b/testdata/package/reference/1.0.0/dataset/reference/manifest.yml
@@ -9,6 +9,12 @@ release: beta
 # Allowed values: logs, metrics, events
 type: logs
 
+elasticsearch:
+  index-template.mappings:
+    dynamic: false
+  index-template.settings:
+    index.lifecycle.name: reference
+
 # This is the ingest pipeline which should be used. If none is define,
 # it checks if a default pipeline exists.
 #ingest_pipeline: default
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -36,11 +36,12 @@ type Dataset struct {
 	Type string `config:"type" json:"type" validate:"required"`
 	Name string `config:"name" json:"name,omitempty" yaml:"name,omitempty"`
 
-	Title          string   `config:"title" json:"title" validate:"required"`
-	Release        string   `config:"release" json:"release"`
-	IngestPipeline string   `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
-	Streams        []Stream `config:"streams" json:"streams,omitempty" yaml:"streams,omitempty" `
-	Package        string   `json:"package,omitempty" yaml:"package,omitempty"`
+	Title          string         `config:"title" json:"title" validate:"required"`
+	Release        string         `config:"release" json:"release"`
+	IngestPipeline string         `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
+	Streams        []Stream       `config:"streams" json:"streams,omitempty" yaml:"streams,omitempty" `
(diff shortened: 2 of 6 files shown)


ruflin:
@skh @neptunian How would we implement this best on the Kibana side?

util/dataset.go
```diff
@@ -32,13 +32,14 @@ var validTypes = map[string]string{
...
-	Release        string   `config:"release" json:"release"`
-	Type           string   `config:"type" json:"type" validate:"required"`
-	IngestPipeline string   `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
-	Streams        []Stream `config:"streams" json:"streams,omitempty" yaml:"streams,omitempty" `
-	Package        string   `json:"package,omitempty" yaml:"package,omitempty"`
+	ID             string         `config:"id" json:"id,omitempty" yaml:"id,omitempty"`
+	Title          string         `config:"title" json:"title" validate:"required"`
+	Release        string         `config:"release" json:"release"`
+	Type           string         `config:"type" json:"type" validate:"required"`
+	IngestPipeline string         `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
```
ycombinator, line 39: Now that we have an `Elasticsearch` field, should `IngestPipeline` be moved under it?

util/dataset.go
```diff
@@ -84,6 +85,11 @@ type Os struct {
 	Windows interface{} `config:"windows" json:"windows,omitempty" yaml:"windows,omitempty"`
 }
 
+type Elasticsearch struct {
+	IndexTemplateSettings map[string]interface{} `config:"index-template.settings" json:"index-template.settings,omitempty" yaml:"index-template.settings,omitempty"`
```
ycombinator, line 89: Nit: casing. For ingest pipelines we use `ingest_pipeline` (underscore-delimited) but for index templates we use `index-template` (hyphen-delimited).

util/dataset.go
ruflin: Good point. I would say yes!

util/dataset.go
ruflin: Glad you bring this up. I remember we had in the past a discussion that we should standardise. You have any preference?

ycombinator: tl;dr: I lean towards underscore.

When it comes to defining these fields in JSON or YAML I don't think it matters much, especially since we would never start a field name with a delimiter (otherwise hyphens could be problematic in YAML). 

When it comes to deserializing these fields into language constructs, hyphen could be problematic in some languages as it might represent subtraction. 

So I lean towards underscore.

util/dataset.go
ycombinator: Will this pose a problem in terms of BWC for existing manifests?

ruflin: Yes and no. After 7.9 we must stop doing these changes but until then we can still update old ones. The way I plan such changes are:

* Update registry and support both format. Read in and output.
* Update Kibana and packages / integrations scripts
* Remove legacy code

@ycombinator Any help getting this changes in is appreciated ;-)

util/dataset.go
ruflin: Let's clean it up and go with underscore!

ruflin: I filed https://github.com/elastic/package-registry/issues/565 for this and will clean up the PR.

ruflin:
@ycombinator PR updated, would be great if you could have a look again.

testdata/package/reference/1.0.0/dataset/reference/manifest.yml
```diff
@@ -9,6 +9,12 @@ release: beta
 # Allowed values: logs, metrics, events
 type: logs
 
+elasticsearch:
+  index_template.mappings:
+    dynamic: false
+  index_template.settings:
+    index.lifecycle.name: reference
```
ycombinator, line 16: Nit: you could break out the `index_template` key into it's own level and nest `mappings` and `settings` under it, like you did in `testdata/package/yamlpipeline/1.0.0/dataset/log/manifest.yml`.

util/dataset.go
ycombinator: Maybe I'm missing something about the plan but shouldn't there be an `IngestPipeline` field in the `Elasticsearch` struct below?

ycombinator: Oh I see, the plan is to introduce it in #564.

ycombinator:
Left a nitpick. PR needs rebase due to conflicts.

testdata/package/reference/1.0.0/dataset/reference/manifest.yml
ruflin: I actually didn't do it on purpose to make sure I test that both options work.

PR #553, opened 2020-06-24
Implement package watcher

mtojek:
This PR introduces a package watcher that can be used in dev mode.

Thanks to this update, packages are served from memory cache if there are no important changes in the filesystem.


main.go
```diff
@@ -100,6 +102,8 @@ func main() {
 	if err := server.Shutdown(ctx); err != nil {
 		log.Fatal(err)
 	}
+
+	util.ClosePackageWatcher()
```
ruflin, line 106: What about using a `defer util.ClosePackageWatcher()` on line 77 instead? Like this we can keep the logic together and the watcher is not closed when it is not started.

util/package_watcher.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/pkg/errors"
+	"github.com/radovskyb/watcher"
+)
+
+const watcherPollingPeriod = time.Second
+
+var (
+	w *watcher.Watcher
+
+	watchedPackages []Package
```
ruflin, line 21: Why not use the same package list cache from packages.go? It seems the other one is not used if this is enabled.

util/package_watcher.go
```diff
@@ -0,0 +1,83 @@
...
+
+	for _, p := range packagePaths {
+		err = w.AddRecursive(p)
+		if err != nil && !os.IsNotExist(err) {
+			log.Fatal(errors.Wrapf(err, "watching directory failed (path: %s)", p))
+		}
+	}
+
+	go func() {
+		go w.Start(watcherPollingPeriod)
```
ruflin, line 44: Why is this started inside the go routine? Could this go on line 42?

mtojek: Actually, it blocks. I can move the whole line up: `go w.Start(...)`

ruflin: jep, that is what I mean. I just saw routine in routine which set off alarm bells :-)

util/package_watcher.go
mtojek: Fixed.

main.go
mtojek: Fixed.

util/package_watcher.go
mtojek: Fixed :)

PR #556, opened 2020-06-25
Add a very simple benchmark tests for NewPackage

ruflin:
I plan to play around a bit with the package creation and I am curious on what impact these changes might have on the benchmarks, because of this adding a simple one here.

Current output on my machine:

```
pkg: github.com/elastic/package-registry/util
BenchmarkNewPackage-8               1624            729592 ns/op
PASS
ok      github.com/elastic/package-registry/util        1.411s
````

### DIFF
--- a/util/package_test.go
+++ b/util/package_test.go
@@ -126,3 +126,10 @@ func TestValidate(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkNewPackage(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		_, err := NewPackage("../testdata/package/reference/1.0.0")
+		assert.NoError(b, err)
+	}
+}

### CODE REVIEW

File: util/package_test.go [hunk 1]
```diff
@@ -126,3 +126,10 @@ func TestValidate(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkNewPackage(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		_, err := NewPackage("../testdata/package/reference/1.0.0")
```
mtojek, line 132: Can you also add a benchmark for NewPackageWithResources?

ruflin: Haha, interesting you ask. Because that is the one I plan to remove :-D And see what the difference in performance is.

mtojek: Please check then, if it doesn't break any validation or package_index handler.

ph:
> nit: it's always good to store somewhere these results. Maybe a dedicated directory in the repo?

Not sure its required, you can compare between branch?

ruflin:
I added it to the PR / commit description so it can be found again. If we get more serious about benchmarks, we should find a place for it.

Issue #535, opened 2020-06-22
Too many open files for many .tar.gz downloads


ruflin:
For so load tests I run the following command:

```
wrk -c 200 -d 30 -t 20 http://localhost:8080/epr/cisco/cisco-0.1.2.tar.gz
```

Unfortunately quite a few errors were returned. Digging into it I found the following errors in the registry logs:

```
2020/06/22 15:20:36 archiving package path 'public/package/cisco/0.1.2' failed: 1 error: processing package path 'public/package/cisco/0.1.2' failed: fcntl: too many open files
```

The problem is that we build the `.tar.gz` currently on the fly and rebuild it every time a request comes in. Few potential solutions:

* Limit the number of parallel requests
* Cache file paths for a certain time
* Cache .tar.gz for a certrain time

Caching the full .tar.gz is probably the most effective approach as it eliminates all file accesses.

ruflin:
@mtojek WDYT would be the best approach to solve the above?

mtojek:
First of all, please post any reports from the load test, so we can review them.

A couple of questions:
1. How did you run the registry?
2. Was it on Linux/Mac/under Docker?
3. What is the system limit for open files (hence the error). Keep in mind that if you fire a load test from the same host as the target, you will use more descriptors (client, server).

I would leave caching to be the last possible solution and focus on optimizing process of building the package (visiting resources might be faster). Remember, there is CDN in between.

ruflin:
* The registry is run locally on OS X
* kern.maxfilesperproc: 24576

Even though there is a CDN in between, it is way too easy to reach this limit. Caching / prebuilding is in our case is simple as these assets for now don't change over time.

mtojek:
You can also verify if descriptors not leaking anywhere.

mtojek:
@jsoriano This issue might be relevant for https://github.com/elastic/package-registry/pull/703

mtojek:
No error occurrences recently, so let me resolve this one.

PR #559, opened 2020-06-25
Cache build .tar.gz file

ruflin:
This introduced a cache for the tar.gz files. The code is not finished yet and hacky but it works. This also helps with https://github.com/elastic/package-registry/issues/535

Few additional notest to the code. Having the .tar.gz cached also means they could be prebuilt. This could be used during a "warmup" of the registry or in case an external storage is mounted in, the prebuilt .tar.gz could be used instead. Also these are cached across restarts.

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -5,8 +5,11 @@
 package main
 
 import (
+	"io/ioutil"
 	"log"
 	"net/http"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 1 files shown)


artifacts.go
```diff
@@ -56,14 +59,42 @@ func artifactsHandler(packagesBasePaths []string, cacheTime time.Duration) func(
...
-		err = archiver.ArchivePackage(w, archiver.PackageProperties{
+		properties := archiver.PackageProperties{
 			Name:    packageName,
 			Version: packageVersion,
 			Path:    packagePath,
-		})
-		if err != nil {
-			log.Printf("archiving package path '%s' failed: %v", packagePath, err)
+		}
+		// TODO: find proper directory, make it a config option
```
mtojek, line 67: I think it needs to be a more advanced concept. We should define an upper limit of total cache size and remove based on LRU rules. Otherwise, we can easily loose space there.

artifacts.go
```diff
@@ -56,14 +59,42 @@ func artifactsHandler(packagesBasePaths []string, cacheTime time.Duration) func(
...
+		basePath := "/tmp/epr"
+		path := basePath + "/" + properties.Name + "-" + properties.Version + ".tar.gz"
+
+		// TODO: Add dev option to skipt caching part
```
mtojek, line 71: I'd like to see this condition implemented. You can use devmode package, there are some methods exposed.

artifacts.go
```diff
@@ -56,14 +59,42 @@ func artifactsHandler(packagesBasePaths []string, cacheTime time.Duration) func(
...
+		// The nice thing here is that it also works across restarts, meaning pre-building
+		// is an option.
+		_, err = os.Stat(path)
+		// If file does not exists, it builds the tar.gz and throws it into a file so it can be served again later.
```
mtojek, line 75: nit: does not exist

ruflin:
Closing this for now as things work!

PR #560, opened 2020-06-25
Introduce v1 API version

ruflin:
This introduces a v1 prefix for all the API calls. The only calls that are not prefixed by v1 are the calls to download the .tar.gz files. As packages should not change, also path to packages should not change.

This change also makes it obvious to me that we have the prefix defined in too many places. To simplify this change, I do some previous cleanup PRs.


ruflin:
@mtojek @jfsiii This is an initial attempt at introducing a versioned API according to https://github.com/elastic/package-registry/issues/204 Please have a look. I did not prefix everything by v1, like the .tar.gz downloads.

jfsiii:
Having some endpoints prefixed and others not is not common in my experience and complicates the client code.

Ideally, all the routes are relative to the same root, whether that's `http://some.tld/v1` or `http://v1.some.tld`

ph:
@jfsiii So you would also add the artifact under the same v1 prefix?

jfsiii:
@ph I'd treat it the same as the other endpoints, yes. I'd prefer a global API version over a resource- or endpoint-level version.

But I'm also confused about the role/visibility of `/epr` or what the drawbacks are to versioning it. A while back, we changed Kibana to read the `download` value instead of "knowing" how to access an archive because, iirc, it was an implementation detail and not for direct access. 

Either way, having all endpoints relative to the same root (be it `http://some.tld/v1` or `http://v1.some.tld`) is easier for clients and my preference

ph:
@ruflin Looking at this into more details I agree with @jfsiii to scope everything, including download path, into a v1 prefix, actually, this make things simpler. Let's say, that we had nginx in front of the registry we would just add a "prefix" to the API.

Worse case we have more cached packages, but again we can cross the rubicon if it's really a problem.

ruflin:
There is a bit a problem here with `/epr/*`. The structure follows the standard structure for our download stats and it seems it is not very easy to adjust. That is why the downloads are under this exact location. So versioning this one will break the stats.

jfsiii:
It feels like this discussion should be in #204 so it's not lost in this PR, but since we're here

Is the stats restriction for subdomains as well, or just paths?

We currently have download urls like

1. `epr.elastic.co/epr/*`
2. `epr-experimental.elastic.co/epr/*`

Do we get stats for (2)? If so, we could do `epr-v1.elastic.co/epr/*` (which seems to be the pattern) , or `v1.epr.elastic.co/epr/*`.

ph:
Okay, let's try to be pragmatic in those issues. I think the three of us would prefer to have the download stats under some v1 concepts. I personally would like that we use **path** instead of **CNAME**, it's easier to manage both at the infra side (DNS) and implementation of client. 

Now there is a problem for the download status server when you use a v1 prefix, I think this is wrong behavior and it should be fixed in the download stats logic.

What can we do now? I think we don't need to rush to merge this PR. We can assume that if **no prefix** is added we are at v1. 
Also, we are stabilizing v1, this mean we should start making incremental changes and aim to not break it. I don't think we would add v2 routes before 8.

I don't think @jfsiii's work with OpenAPI would be impacted by this?

This gives us a few releases to convince to adapt the download stats.

ruflin:
Closing this PR as for now not moving forward with this.

PR #564, opened 2020-06-29
Introduce `elasticsearch.ingest_pipeline.name` as config option

ruflin:
In https://github.com/elastic/package-registry/pull/552 `elasticsearch` as config block is introduced. As discussed there, it also makes sense to use this for the Ingest Pipeline config to have all Elasticsearch related parts in one place.

For now no package uses this change and Kibana does not support it. As both ways are supported, this is not a breaking change yet. As a follow up, the package generation must be updated to adjust for this option and Kibana must be adapted to only support the new option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Apply rule: first package found served. [#546](https://github.com/elastic/package-registry/pull/546)
 * Implement package watcher. [#553](https://github.com/elastic/package-registry/pull/553)
 * Add conditions as future replacement of requirements. [#519](https://github.com/elastic/package-registry/pull/519)
+* Introduce `elasticsearch.ingest_pipeline.name` as config option. [#](https://github.com/elastic/package-registry/pull/)
 
 ### Deprecated
 
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -83,6 +83,10 @@ type Elasticsearch struct {
 	IndexTemplateMappings map[string]interface{} `config:"index_template.mappings" json:"index_template.mappings,omitempty" yaml:"index_template.mappings,omitempty"`
 }
 
+type Elasticsearch struct {
+	IngestPipelineName string `config:"ingest_pipeline.name,omitempty" json:"ingest_pipeline.name,omitempty" yaml:"ingest_pipeline.name,omitempty"`
+}
+
 type fieldEntry struct {
 	name  string
 	aType string
@@ -157,7 +161,18 @@ func (d *Dataset) Validate() error {
 		return fmt.Errorf("type is not valid: %s", d.Type)
 	}
 
-	if d.IngestPipeline == "" {
+	if d.Elasticsearch != nil && d.Elasticsearch.IngestPipelineName == "" {
+		// Check that no ingest pipeline exists in the directory except default
+		for _, path := range paths {
+			if filepath.Base(path) == "default.json" || filepath.Base(path) == "default.yml" {
+				d.Elasticsearch.IngestPipelineName = "default"
+				// TODO: remove because of legacy
+				d.IngestPipeline = "default"
+				break
+			}
+		}
+		// TODO: Remove, only here for legacy
+	} else if d.IngestPipeline == "" {
 		// Check that no ingest pipeline exists in the directory except default
 		for _, path := range paths {
 			if filepath.Base(path) == "default.json" || filepath.Base(path) == "default.yml" {
--- a/util/dataset.go
+++ b/util/dataset.go
@@ -87,10 +87,7 @@ type Variable struct {
 type Elasticsearch struct {
 	IndexTemplateSettings map[string]interface{} `config:"index_template.settings" json:"index_template.settings,omitempty" yaml:"index_template.settings,omitempty"`
 	IndexTemplateMappings map[string]interface{} `config:"index_template.mappings" json:"index_template.mappings,omitempty" yaml:"index_template.mappings,omitempty"`
-}
-
-type Elasticsearch struct {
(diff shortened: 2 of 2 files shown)


util/dataset.go
```diff
@@ -84,6 +85,10 @@ type Os struct {
 	Windows interface{} `config:"windows" json:"windows,omitempty" yaml:"windows,omitempty"`
 }
 
+type Elasticsearch struct {
+	IngestPipelineName string `config:"ingest_pipeline.name,omitempty" json:"ingest_pipeline.name,omitempty" yaml:"ingest_pipeline.name,omitempty"`
```
ruflin, line 89: @ycombinator Not 100% happy with the name yet. Any input?

ycombinator: I think just `IngestPipeline` is enough? Even looking at Filebeat fileset's `manifest.yml`s, I don't think we've ever needed to define any sub-properties of ingest pipelines?

util/dataset.go
```diff
@@ -154,7 +159,18 @@ func (d *DataSet) Validate() error {
...
 	}
 
-	if d.IngestPipeline == "" {
+	if d.Elasticsearch != nil && d.Elasticsearch.IngestPipelineName == "" {
+		// Check that no ingest pipeline exists in the directory except default
+		for _, path := range paths {
+			if filepath.Base(path) == "default.json" || filepath.Base(path) == "default.yml" {
+				d.Elasticsearch.IngestPipelineName = "default"
+				// TODO: remove because of legacy
+				d.IngestPipeline = "default"
```
ycombinator, line 168: Consider making the bare strings, `"default"`, `"default.json"`, and `"default.yml"` into constants. Especially `"default"`, since it is used in a couple of places.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add list of downloads to /search endpoint. [#512](https://github.com/elastic/package-registry/pull/512)
 * Apply rule: first package found served. [#546](https://github.com/elastic/package-registry/pull/546)
 * Implement package watcher. [#553](https://github.com/elastic/package-registry/pull/553)
+* Introduce `elasticsearch.ingest_pipeline.name` as config option. [#](https://github.com/elastic/package-registry/pull/)
```
ycombinator, line 42: Maybe add a note to the Deprecated section of the changelog about deprecating `ingest_pipeline` in favor of `elasticsearch.ingest_pipeline`?

util/dataset.go
```diff
@@ -32,13 +32,14 @@ var validTypes = map[string]string{
...
-	Release        string   `config:"release" json:"release"`
-	Type           string   `config:"type" json:"type" validate:"required"`
-	IngestPipeline string   `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
-	Streams        []Stream `config:"streams" json:"streams,omitempty" yaml:"streams,omitempty" `
-	Package        string   `json:"package,omitempty" yaml:"package,omitempty"`
+	ID             string         `config:"id" json:"id,omitempty" yaml:"id,omitempty"`
+	Title          string         `config:"title" json:"title" validate:"required"`
+	Release        string         `config:"release" json:"release"`
+	Type           string         `config:"type" json:"type" validate:"required"`
+	IngestPipeline string         `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
```
ycombinator, line 39: Consider adding a godoc comment above this that starts with `// Deprecated: `.

util/dataset.go
```diff
@@ -154,7 +159,18 @@ func (d *DataSet) Validate() error {
 		return fmt.Errorf("type is not valid: %s", d.Type)
 	}
 
-	if d.IngestPipeline == "" {
+	if d.Elasticsearch != nil && d.Elasticsearch.IngestPipelineName == "" {
```
ycombinator, line 162: There is duplication of code inside this `if` block and it's sibling `else if` block. Looks like the intent is to prefer `d.Elasticsearch.IngestPipelineName` over `d.IngestPipeline`. How about defining a helper method, something like:

```
func (d *DataSet) ingestPipeline() string {
    if d.Elasticsearch != nil && d.Elasticsearch.IngestPipelineName != "" {
        return d.Elasticsearch.IngestPipelineName
    }

    return d.IngestPipeline
}
```

Then calling it here as using it's return value in the condition of the `if`. Then you should be able to get rid of the duplication.

ycombinator:
PR needs rebase due to conflicts.

util/dataset.go
ruflin: There are 2 properties for ingest pipeline I had in mind:

* Path to the file locally
* Name defined in the manifest instead of generated name in the package manager

That is why I made it an object to leave this door open.

util/dataset.go
ruflin: done.

util/dataset.go
ruflin: done, good idea.

util/dataset.go
ruflin: As the code goes away very soon, I kept it "delete friendly". Do you think it is still worth cleaning up?

ruflin:
@ycombinator could you have a look at my updates and comments?

util/dataset.go
ycombinator: No, I was not aware it was going away soon when I made the clean up comment. Okay to leave it as-is.

CHANGELOG.md
ycombinator: I'm guessing this is also not needed as the old way is going away very soon?

Issue #530, opened 2020-06-22
Remove `event` dataset.type support from registry


ruflin:
Currently the registry supports `dataset.type: event` as the endpoint package still uses it. This should be removed as soon as the endpoint package is not using it anymore.

ruflin:
@jonathan-buttner Not sure what the best path forward here is. Should we potentially remove the old packages that still use event?

jonathan-buttner:
> @jonathan-buttner Not sure what the best path forward here is. Should we potentially remove the old packages that still use event?

Yeah it's probably easiest to just delete our old packages. I'll need probably a day or so to get everything ready for that. I need a couple PRs to go in:

https://github.com/elastic/kibana/pull/69589 once that's merged I can rebuild our package without the `events` dataset and do a release, then we can delete everything before that version

ruflin:
++

ruflin:
PR is open here: https://github.com/elastic/package-registry/pull/567

ruflin:
Closing as PR was merged.

PR #567, opened 2020-06-29
Remove `dataset.type: event` as supported type.

ruflin:
This event type was only used by the endpoint package and it got removed. Because of this, it can be removed from the list.


ruflin:
This closes https://github.com/elastic/package-registry/issues/530

PR #568, opened 2020-06-29
Cleanup: Packages and package naming

ruflin:
Just a few things I stumbled over browsing through the code:

* Instead of using `[]Package`, this introduces a `Packages` type. This will allow us to build more tooling like locks and similar on top of Packages.
* Rename `aPackage` to `p` as we use it in several other places.

No functional changes.

### DIFF
--- a/package_index.go
+++ b/package_index.go
@@ -60,7 +60,7 @@ func packageIndexHandler(packagesBasePaths []string, cacheTime time.Duration) fu
 		w.Header().Set("Content-Type", "application/json")
 		cacheHeaders(w, cacheTime)
 
-		aPackage, err := util.NewPackageWithResources(packagePath)
+		p, err := util.NewPackageWithResources(packagePath)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


package_index.go
```diff
@@ -60,15 +60,15 @@ func packageIndexHandler(packagesBasePaths []string, cacheTime time.Duration) fu
 		w.Header().Set("Content-Type", "application/json")
 		cacheHeaders(w, cacheTime)
 
-		aPackage, err := util.NewPackageWithResources(packagePath)
+		p, err := util.NewPackageWithResources(packagePath)
```
mtojek, line 63: I'd rather stick to this explanation: https://www.codereadability.com/i-n-k-single-letter-variable-names/

ruflin: I rather stick to consistency across the registry code for now. 

I would add to this list when using inside a method to reference the type like `func (p Package) ...`. But yes, personal preference.

PR #514, opened 2020-06-16
Update package output structure

ruflin:
This change contains several changes to the output generated for the package API endpoint:

* dataset.id is renamed to dataset.name
* datasources is renamed to config_templates. Currently only in the output, packages can follow later.
* config_templates (aka datasources) do not pregenerate the content anymore to simply config building in Kibana
* template_content is removed

Further change is done to use `Dataset(s)` instead of `DataSet(s)` in the code for consistency.

PR #569, opened 2020-06-29
Add OpenAPI schemas for responses

jfsiii:
## Summary

- Created OpenAPI schemas for the response types by looking at the `*Handler` functions and  `util/*.go` `struct`s 
- Changed `internal` query param from a `string` to a `boolean` with a default value of `false`
- Added the same schema for the previously undocumented `experimental` param

It might be easiest to review by using an online viewer like https://editor.swagger.io/ and importing the URL to the raw file https://raw.githubusercontent.com/jfsiii/package-registry/cd276b32f7c82bc01dbbf380684c789ca3672eb7/openapi.yml or copying & pasting it in and reviewing the docs it generates from the spec file

Screen Shot 2020-06-29 at 4 49 05 PM

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -54,6 +54,8 @@ paths:
                       count: 17
       operationId: get-categories
       description: List of the existing package categories and how many packages are in each category
+      parameters:
+        - $ref: '#/components/parameters/experimentalPackageParam'
   /search:
     get:
       summary: Search packages
@@ -85,11 +87,8 @@ paths:
           in: query
           name: package
           description: 'Filters by a specific package name, for example mysql. In contrast to the other endpoints, it will return by default all versions of this package.'
-        - schema:
-            type: string
-          in: query
-          name: internal
-          description: 'This can be set to true, to also list internal packages. This is set to false by default.'
+        - $ref: '#/components/parameters/internalPackageParam'
+        - $ref: '#/components/parameters/experimentalPackageParam'
   '/package/{package}/{version}':
     get:
       summary: GET package info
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -112,3 +115,268 @@ paths:
         in: path
         required: true
         description: Version of the package
+components:
+  schemas:
+    IndexData:
```
jfsiii, line 120: The schema names (`IndexData`, `Category`, etc) are based off a golang `struct` with the same name

openapi.yml
```diff
@@ -112,3 +115,268 @@ paths:
...
+          type: string
+    Requirement:
+      title: Requirement
+      type: object
+      properties:
+        Kibana:
+          $ref: '#/components/schemas/ProductRequirement'
+      required:
+        - Kibana
+    Dataset:
```
jfsiii, line 173: I'll update this and other types now that https://github.com/elastic/package-registry/pull/514/files#diff-7dea786222588c32c19238bffffee9c2 is merged

jfsiii:
From least to most sound

0. **Memory, habit, etc:** Author must remember to update and/or reviewers must remember to check.
1. **Generate mock server from spec and run the tests against it**: At least the spec as good as the tests.
2. **Have our golang server import the spec and use that for attaching handlers, dealing w/req params, etc**: For the places we use it, we'll agree with the published spec.
3. **Generate *real* server from the spec**: It can't be out-of-date

I think (1) shouldn't be bad. Lots of options at https://openapi.tools/#mock. I've not used https://github.com/danielgtaylor/apisprout but it caught my eye. Or we could use a parser like https://github.com/getkin/kin-openapi and create the mock another way. This applies to (2) as well.

I suspect (3) is too far off to think about, but https://github.com/deepmap/oapi-codegen seemed like a solid project

PR #571, opened 2020-06-30
Remove support for logs/metrics categories

ruflin:
The metrics and logs categories are not supported anymore. This removes support for it. But it can only be merged after https://github.com/elastic/package-storage/pull/97 landed.


ruflin:
@mtojek Yes, the categories are available here: https://github.com/elastic/package-registry/blob/master/util/package.go#L23

PR #572, opened 2020-06-30
Validate that directory do not contain -

ruflin:
As we try to ensure no directories contain -, we can also check for this. This adds a check.

The additional NewPackage method with assets was removed and merged into NewPackage for now. This means NewPackage is a bit more heavy but we have all the validation in place. It also means, the memory consumption to keep all packages is a bit higher. For the number of packages we have today, this should be ok.

Instead I plan a future refactoring that serves the /search endpoint only from the info of the BasePackage. So not all package information with the assets has to be kept in memory.

This removes all the - assets in the packages.


ruflin:
Fixed commit message and merged ;-)

PR #575, opened 2020-07-01
Fix that first package wins

ruflin:
In the search endpoint, follow up packages with the same name and version did overwrite the existing package. But we defined in the past that the first package found should win. This does a fix in the search endpoint and adds tests for it.

Long term I think the fix should not be in search but getPackages should already return a unique list, but this one here was easier to fix.

Also tests with multiple package paths were added so we see in case we break it in the future.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -32,7 +32,7 @@ var (
 )
 
 func TestEndpoints(t *testing.T) {
-	packagesBasePaths := []string{"./testdata/package"}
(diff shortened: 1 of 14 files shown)


ruflin:
@mtojek For the categories one etc. I'm not so sure. But I plan a fix on getPackages which will solves these automatically. Also there it is not a direct problem as the package info is not fully used. It is a problem, if the same version has different categories which would be unexpected.

mtojek:
I think that the method `util.GetPackages` should work as expected, but it's better to verify it.

PR #582, opened 2020-07-02
Switch to production branch from package-storage

ruflin:
As the master branch will not contain any packages anymore in the future, the packages used to run the registry locally must be adjusted. See https://github.com/elastic/package-storage/pull/116 for more details.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -55,7 +55,8 @@ func fetchPackageStorage() error {
 
 	// If storage directory does not exists, check it out
 	if _, err := os.Stat(storageRepoDir); os.IsNotExist(err) {
-		return sh.Run("git", "clone", "--depth=1", "https://github.com/elastic/package-storage.git", storageRepoDir)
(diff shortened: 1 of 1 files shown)


ruflin:
@mtojek I was not sure if we should get the packages from production or "production + staging + snapshot" for running it locally here. In theory, all packages also in snapshot should be compatible and never break the registry. WDYT?

magefile.go
```diff
@@ -55,7 +55,8 @@ func fetchPackageStorage() error {
 
 	// If storage directory does not exists, check it out
 	if _, err := os.Stat(storageRepoDir); os.IsNotExist(err) {
-		return sh.Run("git", "clone", "--depth=1", "https://github.com/elastic/package-storage.git", storageRepoDir)
+		return sh.Run("git", "clone", "--depth=1", "--single-branch", "--branch", "production", "https://github.com/elastic/package-storage.git", storageRepoDir)
```
ph, line 58: glad to see --depth=1

PR #583, opened 2020-07-02
Remove packages from registry Docker build

ruflin:
The Dockerfile for the registry itself should not contain any packages. Instead it should be empty and there are other distributions with packages, see elastic/package-storage#86. This removes the packages from the default Docker build.

Few additional changes in the this PR:

* Split up the Dockerfile in two stages to reduce the size of the image. Thanks @jsoriano for the contribution
* Switch over to production packages instead of master
* Select /packages as default path

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This image contains the package-registry binary.
-# It expects packages to be mounted under /packages or have a config file loaded into /package-registry/config.yml
+# It expects packages to be mounted under /packages/package-registry or have a config file loaded into /package-registry/config.yml
 
 # Build binary
 ARG GO_VERSION=1.14.2
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,44 +1,37 @@
-# This Dockerfile allows to build the package-registry and packages together.
-# It is decoupled from the packages and the registry even though for now both are in the same repository.
-ARG GO_VERSION=1.14.2
-FROM golang:${GO_VERSION}
-
-# Get dependencies
-RUN \
-    apt-get update \
-      && apt-get install -y --no-install-recommends zip rsync \
-      && rm -rf /var/lib/apt/lists/*
-
-# Copy the package registry
-COPY ./ /home/package-registry
-WORKDIR /home/package-registry
-
-ENV GO111MODULE=on
-RUN go get -u github.com/magefile/mage
-# Prepare all the packages to be built
-RUN mage build
+# This image contains the package-registry binary.
+# It expects packages to be mounted under /packages or have a config file loaded into /package-registry/config.yml
 
 # Build binary
+ARG GO_VERSION=1.14.2
+FROM golang:${GO_VERSION} AS builder
+
+ENV GO111MODULE=on
+COPY ./ /package-registry
+WORKDIR /package-registry
 RUN go build .
 
-# Move all files need to run to its own directory
-# This will become useful for staged builds later on
-RUN mkdir -p /registry/public # left for legacy purposes
-RUN mkdir -p /registry/packages/package-storage
-RUN mv package-registry /registry/
-RUN cp -r build/package-storage/packages/* /registry/packages/package-storage/
-RUN cp config.docker.yml /registry/config.yml
+
+# Run binary
+FROM centos:7
+
+# Get dependencies
+RUN yum install -y zip rsync && yum clean all
+
+# Move binary from the builder image
+COPY --from=builder /package-registry/package-registry /package-registry/package-registry
 
 # Change to new working directory
-WORKDIR /registry
+WORKDIR /package-registry
 
-# Clean up files not needed
-RUN rm -rf /home/package-registry
+# Get in config which expects packages in /packages
+COPY config.docker.yml /package-registry/config.yml
 
+# Start registry when container is run an expose it on port 8080
 EXPOSE 8080
-
 ENTRYPOINT ["./package-registry"]
+
 # Make sure it's accessible from outside the container
 CMD ["--address=0.0.0.0:8080"]
 
 HEALTHCHECK --interval=1s --retries=30 CMD curl --silent --fail localhost:8080/health || exit 1
+
--- a/config.docker.yml
+++ b/config.docker.yml
@@ -1,6 +1,9 @@
 public_dir: ./public
+
+# If users want to add their own packages, they should be put under
+# /packages/package-registry or the config must be adjusted.
 package_paths:
-  - /packages
+  - /packages/package-registry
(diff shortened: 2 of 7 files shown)


ruflin:
Opening as draft as the change is not complete and it needs to be checked if this maybe effect the integrations repository.

ruflin:
@mtojek I checked the integrations repo and it seems this will not effect it as in all places I could find, you mount your own volume and your own config with your own packages.

ruflin:
I also opened https://github.com/elastic/integrations/pull/156 to attach integrations to a specific version of the registry.

Dockerfile
```diff
@@ -9,35 +9,28 @@ RUN \
...
-# Move all files need to run to its own directory
-# This will become useful for staged builds later on
-RUN mkdir -p /registry/public # left for legacy purposes
-RUN mkdir -p /registry/packages/package-storage
-RUN mv package-registry /registry/
-RUN cp -r build/package-storage/packages/* /registry/packages/package-storage/
-RUN cp config.docker.yml /registry/config.yml
+# Move binary to its own directory and clean up
+RUN mv package-registry /package-registry/
+RUN rm -rf /package-registry/tmp
```
jsoriano, line 22: A suggestion about this Dockerfile.

Each `RUN` command moving or removing files creates its own layer, duplicating the affected files, this increases the size of the final image. And we are also including the layers from the golang image. The image built with this Dockerfile takes 843MB.

For these cases we can use [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/), for example the following Dockerfile generates an image based on Debian 10 (same base image as golang:1.14.2) with the package-registry, rsync and zip of 128MB:
```
# This Dockerfile allows to build the package-registry and packages together.
# It is decoupled from the packages and the registry even though for now both are in the same repository.
ARG GO_VERSION=1.14.2
FROM golang:${GO_VERSION} AS builder

# Build binary
ENV GO111MODULE=on
COPY ./ /package-registry
WORKDIR /package-registry
RUN go build .

FROM debian:10.4

# Get dependencies
RUN \
    apt-get update \
      && apt-get install -y --no-install-recommends zip rsync \
      && rm -rf /var/lib/apt/lists/*

# Move binary from the builder image
COPY --from=builder /package-registry/package-registry /package-registry/package-registry

# Change to new working directory
WORKDIR /package-registry

# Get in config which expects packages in /packages
COPY config.docker.yml /package-registry/config.yml

# Start registry when container is run an expose it on port 8080
EXPOSE 8080
ENTRYPOINT ["./package-registry"]

# Make sure it's accessible from outside the container
CMD ["--address=0.0.0.0:8080"]

HEALTHCHECK --interval=1s --retries=30 CMD curl --silent --fail localhost:8080/health || exit 1
```

Regarding the base image, on other projects we are using the `centos:7` image, maybe we want to use the same image here. For that, we would have to replace the debian specific lines with something like this:
```
...
FROM centos:7

# Get dependencies
RUN yum install -y zip rsync && yum clean all
...
```

ruflin: @jsoriano 100x times nicer what you proposed. I copy / pased it and pushed a new commit. Also switched to centos, no preference on my end.

I really appreciate the copy / paste friendly comment you made, thank you for the effort!

ruflin:
Update the PR with all recommendations, should now be ready for a round of reviews. 

@jsoriano Your contribution will directly make our production images much smaller as they depend on this on here.

Dockerfile
```diff
@@ -1,44 +1,37 @@
...
-RUN mkdir -p /registry/packages/package-storage
-RUN mv package-registry /registry/
-RUN cp -r build/package-storage/packages/* /registry/packages/package-storage/
-RUN cp config.docker.yml /registry/config.yml
+# Move binary from the builder image
+COPY --from=builder /package-registry/package-registry /package-registry/package-registry
 
 # Change to new working directory
-WORKDIR /registry
+WORKDIR /package-registry
```
mtojek, line 24: Keep in mind that changing this path may break depending docker-compose definitions:

https://github.com/elastic/integrations/blob/master/testing/environments/snapshot.yml#L46
https://github.com/elastic/integrations/blob/master/testing/environments/kubernetes/snapshot.yml#L193

config.docker.yml
```diff
@@ -1,6 +1,6 @@
 public_dir: ./public
 package_paths:
-  - ./packages/package-storage
+  - /packages
```
mtojek, line 3: Let me explain why was there a reference to the `package-storage`.

In integrations we mount the second path for packages, so that we have:

```
./packages/package-storage
./packages/integrations
```
so that all packages are stored in the same `/packages` directory. Using a single directory prevents from simply loading another path with packages without replacing existing ones.

If you plan to change the current behavior here, I believe you need to update the testing/environment in integrations.

main_integration_test.go
```diff
@@ -23,6 +23,12 @@ import (
 // and the setup command works as expected.
 func TestSetup(t *testing.T) {
 
+	// Mage build is needed to pull in the packages from package-storage
```
mtojek, line 26: I don't think this is correct.

`test` shouldn't build any code. That's why you've dependencies between targets (not sure if mage supports it) or you call it `build test`

config.docker.yml
ruflin: My assumption was this keeps working as in https://github.com/elastic/integrations/blob/master/testing/environments/snapshot.yml#L47 you overwrite the default config?

Dockerfile
ruflin: Got it. Assuming we get https://github.com/elastic/integrations/pull/156 in first, we could roll it out in two steps? I will update https://github.com/elastic/integrations/pull/156 to also adjust the k8s one to a fix registry.

main_integration_test.go
ruflin: Updated to only use fetchPackageStorage instead to prevent the building of the binary.

mtojek:
For reference: https://github.com/elastic/package-registry/pull/367#discussion_r412894410 We talked about multi-stage built images a couple of weeks ago. I'm glad we have it finally!

main_integration_test.go
mtojek: I'm afraid that this is something we won't skip.

ruflin: @mtojek Not sure I can follow your last comment. What part will we not skip?

mtojek: not to fetch packages

config.docker.yml
ruflin: Updated to use `/packages/package-registry` as default and added a note to the dockerfile and config what users should do about it.

PR #585, opened 2020-07-03
Remove build trigger from package-storage:master

ruflin:
package-storage master is replaced by the 3 release branches. These have a fixed version of the registry attached so no need to trigger builds anymore here.


ruflin:
@kuisathaverat can you think of other places where we need to update things.

kuisathaverat:
@ruflin There is another change related I will make, but it is more for consistency between pipelines in the package-storage, I will make a PR in a few

ruflin:
@kuisathaverat Thanks. Should this PR wait to be merged or move forward?

PR #597, opened 2020-07-22
Remove development mode

mtojek:
The development mode appeared to be buggy and relatively slow (due to docker volumes). As it has been replaced with https://github.com/elastic/integrations/pull/187 , we can remove it from here.


mtojek:
CI looks happy, so I will merge this cleanup, self-approved.

PR #600, opened 2020-07-27
Template path was not filled in when stream disabled

ruflin:
The template path was not filled in when a stream was disabled by default. The issue was that the template path addition was inside the if clause but it should not be. This PR fixes this.


ph:
@ruflin was this pushed online?

EricDavisX:
double checking PH's question above  ^ want to know when we can see this from 7.9 usage, being explicit helps - the team members don't all know the flow of fixes from EPR into the ELK stack usage

ruflin:
It is available in all 3 environments now.

PR #602, opened 2020-07-27
Prepare release v0.7.1

ruflin:


ruflin:
Oh, there was a v0.7.0 release? Seems like we missed to update the changelog? I now also see the tag. Seems like this becomes 0.7.1. then?

ruflin:
After getting in all the changelogs for v0.7.0 I updated this PR to cut v0.7.0 and v0.7.1 at the same time. Turns out v0.7.0 is already shipped.

ruflin:
@mtojek My thinking is that I will update all 3 releases after to v0.7.1. Not sure why production is still on v0.6.0. @ph might know?

PR #612, opened 2020-07-29
Update testing to run against 7.9.0-SNAPSHOT

ruflin:
Based on https://github.com/elastic/integrations/pull/202 standard registry tests should run against 7.9. We should make sure in the future it is run against not only 7.9 but all the different versions.


ruflin:
Made this ready for review. We should follow up to start testing the registry against multiple stack versions as with changes we should not break any of them.

PR #628, opened 2020-08-27
Ship packages as zip instead of tar.gz

### LINKED ISSUE #474: Discuss: Move to zip instead of `tar.gz` for the format of the package.

ruflin:
This is a breaking change and first needs changes on the Kibana side. The new registry and the new version of Kibana should be either rolled out in sync or Kibana supports both for a bit.

Closes #474


ruflin:
Opening this as ready for review even though CI will be red on the integration tests. The reason is that Kibana does not support zip at the moment.

mtojek:
Do you think it's feasible to prepare an archiver abstraction and support two modes (zip, tar.gz)?

ruflin:
@mtojek Definitively possible but what would be the advantage of having the abstraction if we only plan to support one?

mtojek:
Last time we also planned to support single one :) In the description you've written that we can't push it until Kibana supports the zip format. With an appropriate abstraction we don't need to wait for this.

ruflin:
I realised, only Kibana can handle both as it is a single entry in the registry output (`download`). Lets not add abstraction for it before we are sure we will need it in the future. 

As soon as https://github.com/elastic/kibana/pull/76197 is merged I'll trigger CI again to see if it goes green.

mtojek:
The implementation looks good to me, let's wait for the CI to pass all tests.

ruflin:
jenkins, test this

ruflin:
With 7.10 docker images from Kibana updated, this now seems to pass CI too and should be ready to be merged. @jfsiii @mtojek Would be good to get from both of you the 👍

archiver/archive.go
```diff
@@ -87,24 +80,20 @@ func ArchivePackage(w io.Writer, properties PackageProperties) (err error) {
...
 
-	err = tarWriter.Flush()
-	if err != nil {
-		return errors.Wrap(err, "flushing tar writer failed")
-	}
-
-	err = gzipWriter.Flush()
+	err = zipWriter.Flush()
 	if err != nil {
 		return errors.Wrap(err, "flushing gzip writer failed")
```
mtojek, line 85: nit: zip writer

mtojek (approved):
single nit-pick only

archiver/archive.go
ruflin: fixed.

PR #629, opened 2020-08-27
Rename image src to path and have src as the original value from the …

ruflin:
…manifest

All the configs specified in the manifest file by the user should be exposed without changes in our JSON output. In the case of images, the src was overwritten by the path variable. This means what was in the manifest and outputted by JSON was not a 1:1 match.

It is ok to add additional fields in the JSON output but the original fields should never be touched.

This is a breaking change and need adjustements in Kibana.

### DIFF
--- a/util/package.go
+++ b/util/package.go
@@ -107,6 +107,7 @@ type Owner struct {
 
 type Image struct {
 	Src   string `config:"src" json:"src" validate:"required"`
+	Path  string `config:"path" json:"path"`
 	Title string `config:"title" json:"title,omitempty"`
 	Size  string `config:"size" json:"size,omitempty"`
 	Type  string `config:"type" json:"type,omitempty"`
@@ -167,13 +168,13 @@ func NewPackage(basePath string) (*Package, error) {
 
 	if p.Icons != nil {
 		for k, i := range p.Icons {
-			p.Icons[k].Src = i.getPath(p)
+			p.Icons[k].Path = i.getPath(p)
 		}
 	}
 
 	if p.Screenshots != nil {
 		for k, s := range p.Screenshots {
-			p.Screenshots[k].Src = s.getPath(p)
+			p.Screenshots[k].Path = s.getPath(p)
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


ruflin:
@ycombinator This addresses the most urgent issue that we overwrite variables from the manifest. A follow up discussion is needed on what convention we follow for naming the different fields and standardise on from where the path is relative. I picked `path` for now as we use it the same way in other places too.

ruflin:
@skh @jfsiii I expect this would effect a few parts in Kibana to switch over to `path` from `src` for icons and screenshots.

ruflin:
@jfsiii This is a breaking change for Kibana. Do you see a way to potentially support both options for a while. Something like if `path` does not exist, still use `src` and clean it up later.

jfsiii:
@ruflin yes, I believe that's right. I created https://github.com/elastic/kibana/issues/76380 to track it. I'll leave for you & @ph to prioritize but it should :tm: be a quick & safe change

Could you add the new `path` property to the OpenAPI spec under https://github.com/ruflin/package-registry/blob/1e1106a48ab0fb309d245c90d3fcccdee4d2a06c/openapi.yml#L149-L150

ruflin:
I added the path in the openapi spec, thanks for the reminder.

For my own reference: Here the PR to test this with: https://github.com/elastic/kibana/pull/76434

util/package.go
```diff
@@ -107,6 +107,7 @@ type Owner struct {
 
 type Image struct {
 	Src   string `config:"src" json:"src" validate:"required"`
+	Path  string `config:"path" json:"path"`
```
mtojek, line 110: Does this change need a dependency update in integrations repository? Same question for package-storage.

mtojek (approved):
LGTM

util/package.go
ruflin: I don't think so as this only changes the output generated by the registry and nothing on the package side itself. Anything specific you were thinking of?

mtojek: I was thinking about the case in which the Kibana expects a missing property.

ruflin: This should be solved by https://github.com/elastic/kibana/pull/76434

ruflin:
Tested again with 7.10.0-SNAPSHOT of Kibana and seems to work as expected.

PR #632, opened 2020-09-02
Remove public_dir config

ruflin:
This config is replaced by package_paths and not needed anymore.


ruflin:
@mtojek Do we still use this somewhere?

mtojek (approved):

PR #633, opened 2020-09-02
Rename version to service.version in index handler

ruflin:
To be in line with ECS, version has been renamed to service.version: https://www.elastic.co/guide/en/ecs/current/ecs-service.html Even though this is marked as a breaking change, I do not assume this will break any current usage of the registry.


mtojek (approved):
LGTM

PR #634, opened 2020-09-02
Update testing environment to 7.10

ruflin:
Moving forward we will make some breaking changes in the registry. Some of these are only supported by Kibana >= 7.10. Because of this the testing environment needs to be updated.

### DIFF
--- a/testing/environments/snapshot.yml
+++ b/testing/environments/snapshot.yml
@@ -3,7 +3,7 @@
 version: '2.3'
 services:
   elasticsearch:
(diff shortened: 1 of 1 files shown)


ruflin:
@mtojek @EricDavisX @ycombinator @mdelapenya I assume the above will also apply in other places ...

mdelapenya:
We have a 7.9.x branch and master, so probably we should eventually update to use 7.10.0-SNAPSHOT

ruflin:
Self approved as I need this for other PR's to pass.

ruflin:
@ycombinator For awareness.

PR #640, opened 2020-09-24
fix: Update docker registry path

mdelapenya:


ruflin:
@mdelapenya Do you know if this code is still run from somewhere. I glad you stumbled over this as we should not publish images to my private account anymore as all has its own place by now.

mdelapenya:
I'm not sure about this. What I do know is about the e2e tests or the clusters, and we do not use this image.

I found it after a wide search across all our repos looking for the outdated URL.

ruflin:
@kuisathaverat @JoshMock You know more?

I'm tempted to remove all the code and see what breaks 😇

JoshMock:
Staging and prod use `docker.elastic.co/package-registry/distribution`, so anything going to the `employees/ruflin` namespace is not being used for anything mission critical any more.

It also [looks like](https://container-library.elastic.co/r/employees/ruflin/package-registry) there are no published tags for this anyway.

It's probably fine to remove entirely.

ruflin:
I opened https://github.com/elastic/package-registry/pull/642 to completely remove it. If CI goes green, will get it in and it will make this PR obsolete. Thanks @mdelapenya for spotting.

kuisathaverat (approved):

kuisathaverat:
upss!!! obsolete 😃

PR #644, opened 2020-10-02
Fix: expose proper EPR version

mtojek:
I noticed that EPR is exposed with unreleased version (0.13.1). This PR fixes it.


mtojek:
self-approved

PR #649, opened 2020-10-29
Set cache headers for 404 and 400 to 0

ruflin:
Fastly caches 404 headers by default. This could cause issues when rolling out a new registry and a path was requested to early. This sets the cache time to 0.

Fastly docs: https://docs.fastly.com/en/guides/http-status-codes-cached-by-default

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -34,7 +34,7 @@ func cacheHeaders(w http.ResponseWriter, cacheTime time.Duration) {
 
 func noCacheHeaders(w http.ResponseWriter) {
 	w.Header().Add("Cache-Control", "max-age=0")
-	w.Header().Add("Cache-Control", "private, no store")
+	w.Header().Add("Cache-Control", "private, no-store")
 }
 
 func jsonHeader(w http.ResponseWriter) {
--- a/handler.go
+++ b/handler.go
@@ -17,10 +17,12 @@ import (
 var errResourceNotFound = errors.New("resource not found")
 
 func notFoundError(w http.ResponseWriter, err error) {
+	cacheHeaders(w, 0)
 	http.Error(w, err.Error(), http.StatusNotFound)
 }
 
 func badRequest(w http.ResponseWriter, errorMessage string) {
+	cacheHeaders(w, 0)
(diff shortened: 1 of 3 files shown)


kuisathaverat (approved):

handler.go
```diff
@@ -17,10 +17,12 @@ import (
 var errResourceNotFound = errors.New("resource not found")
 
 func notFoundError(w http.ResponseWriter, err error) {
+	cacheHeaders(w, 0)
```
mtojek, line 20: I think you need to change the internal implementation. The recommended for Cache-Control value for not cachable resources is:
`Cache-Control: no-cache, no-store, must-revalidate`

see: https://developer.mozilla.org/pl/docs/Web/HTTP/Headers/Cache-Control (Preventing caching)

ruflin: WDYT about using "private"? https://docs.fastly.com/en/guides/configuring-caching#do-not-cache

mtojek: Hm.. "private" is to inform the intermittent servers not to cache data, because they will be cached by browsers. To prevent caching on the browser level, AFAIK you need to add `Max-Age: 0`.

mtojek: Referring to the Mozilla docs, `Cache-Control: no-store` should be sufficient at this point:


>no-store
>The response may not be stored in any cache. Although other directives may be set, this alone is the only directive you need in preventing cached responses on modern browsers. max-age=0 is already implied. Setting must-revalidate does not make sense because in order to go through revalidation you need the response to be stored in a cache, which no-store prevents.

ruflin: Fastly does not respect `no-store`:

```
Fastly does not currently respect no-store or no-cache directives. Including either or both of these in a Cache-Control header has no effect on Fastly's caching decision, unless you alter this behavior using custom VCL.
```

ruflin:
This now sets
```
Cache-Control: max-age=0
Cache-Control: "private, no store"
```

This should work for browser and especially Fastly too.

jonathan-buttner (approved):

mtojek (approved):

PR #655, opened 2020-11-19
Allow `template_path` field on inputs

jen-huang:
Follow up to https://github.com/elastic/package-spec/pull/79. Kibana needs `template_path` to ascertain which input template file to read from to build the agent YAML. This PR lets the registry serve that field at the input level, if defined.


jalvz (approved):
thanks!

mtojek (changes requested):
Could you please add a new test package with relevant data stream to the `testdata/package` directory? You can test it with `go test .`. You will also need to regenerate expected data with `go test . -generate`.

Please also add a CHANGELOG entry.

jen-huang:
@mtojek Thanks, I added `input_level_templates` test package and updated changelog.

mtojek (approved):

PR #656, opened 2020-11-20
Add traces as valid type

jalvz:
Follow up on https://github.com/elastic/package-spec/pull/78


ruflin (commented):
Change LGTM, could you add a changelog entry?

Would be good if @ycombinator or @mtojek could also have a look.

mtojek (commented):
Could you please add a new test package with relevant data stream to the `testdata/package` directory? You can test it with `go test .`. You will also need to regenerate expected data with `go test . -generate`.

mtojek:
As @ruflin has already mentioned, please add a CHANGELOG.

mtojek (approved):
Free free to merge it if the CI build goes green.

PR #657, opened 2020-12-03
Add ilm policy field to data stream

nnamdifrankie:
Issue: elastic/security-team#548

This PR adds the ilm_policy field to data stream manifest file. The ilm_policy is required to specify an alternate ilm policy to override the policy specified by the type field e.g. logs, metrics.

### `ilm_policy` field

This feature is needed so that we can get more flexibility when defining data streams within our packages, specifically when it comes to ilm_policy. Currently, the ilm_policy is tied to the type of the data stream. We want to add a field where we can specify an ilm_policy to override this behavior. There will be a subsequent PR to the Fleet code that will install the components according to this new ilm_policy field.

Our specific use case that is driving this is to define a data stream that is meant to collect documents that are used for diagnostics/telemetry. There will be a separate telemetry server that will query this particular data stream and collect the documents. As a result, we want the ilm_policy to be fairly aggressive as we do not need the documents to stay in storage for very long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
 * Add input-level `template_path` field. [#655](https://github.com/elastic/package-registry/pull/655)
 * Add "hidden" field to data stream. [#660] (https://github.com/elastic/package-registry/pull/660)
+* Add "ilm_policy" field to data stream. [#657] (https://github.com/elastic/package-registry/pull/657)
+
 ### Deprecated
 
 ### Known Issue
--- a/config.yml
+++ b/config.yml
@@ -1,3 +1,4 @@
 package_paths:
   - ./testdata/package
   - ./build/package-storage/packages
+  - /Users/fejoh/Documents/WORK/SOURCE/endpoint-package/out/packages
--- a/config.yml
+++ b/config.yml
@@ -1,4 +1,3 @@
 package_paths:
   - ./testdata/package
   - ./build/package-storage/packages
-  - /Users/fejoh/Documents/WORK/SOURCE/endpoint-package/out/packages
--- a/testdata/generated/categories-experimental.json
+++ b/testdata/generated/categories-experimental.json
@@ -22,7 +22,7 @@
   {
     "id": "custom",
     "title": "Custom",
-    "count": 11
+    "count": 12
   },
   {
     "id": "message_queue",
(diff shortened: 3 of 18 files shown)


mtojek (commented):
When you're modifying the package spec, you need to prepare a PR to the package-spec repository first (https://github.com/elastic/package-spec). We use it to discuss manifest properties there. Once it's approved, you can propagate your changes to other repositories.

nnamdifrankie:
> When you're modifying the package spec, you need to prepare a PR to the package-spec repository first

@mtojek Can you point me to the file I need to change over there? This is my first time in this repositories.

mtojek:
The package specification is written using JSON schema. The entrypoint for schema is: https://github.com/elastic/package-spec/blob/master/versions/1/ and the data stream manifest is here: https://github.com/elastic/package-spec/blob/master/versions/1/data_stream/manifest.spec.yml (I suppose you'd like to modify that file).

Please also add a sample package to: https://github.com/elastic/package-spec/tree/master/code/go/internal/validator/test/packages

...and enable the package directory in: https://github.com/elastic/package-spec/blob/master/code/go/pkg/validator/validator_test.go

To quickly ramp up in this area I suggest to review already closed pull requests. The 99% of them are spec extensions.

ruflin:
For reference, here is the PR for the package spec change: https://github.com/elastic/package-spec/pull/90

@nnamdifrankie I would also like to ask you here to split up hidden and ilm policy into two parts. Don't forget the changelog.

mtojek:
... and please also update the PR description.

mtojek (commented):
The missing part in this PR are test packages I mentioned in this comment: https://github.com/elastic/package-registry/pull/657#issuecomment-738169296

Please add a test package similar to one defined in the package spec and create expected test results (`go test . -generate`). To verify if tests are passing correctly, you can use `go test .`.

kevinlog:
@mtojek thanks for the info.

we added the test packages in this PR and the other: https://github.com/elastic/package-registry/pull/660

I realize that when I merge one, I'll need to rebase the other, but just looking for approval so we can move forward.  Let me know if there's anything else

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
 * Add input-level `template_path` field. [#655](https://github.com/elastic/package-registry/pull/655)
+* Add ilm_policy field to data stream. [#657] (https://github.com/elastic/package-registry/pull/657)
```
mtojek, line 17: nit: `ilm_policy` or "ilm_policy"

mtojek (approved):
LGTM!

@kevinlog I don't have any more comments. Thank for your contributions and all adjustments. Let me know if you need more help with releasing your changes (for reference: https://github.com/elastic/package-registry#release)

PR #658, opened 2020-12-03
Support input and stream condition

nchaulet:
## Description 

Related to https://github.com/elastic/package-spec/issues/86  and https://github.com/elastic/integrations/pull/411

@ycombinator added the support for `condition` at the input and stream level for this change to be available to agent it need to be available in the package registry so Kibana can see it.

This PR expose the `condition` property both at the input level and the stream level

Should we had test data?  I tested with the system package from https://github.com/elastic/integrations/pull/411 and it work as expected


ruflin:
It would be good to add a test package with an example of this in https://github.com/elastic/package-registry/tree/master/testdata/package if possible

This PR will need a changelog entry.

mtojek (changes requested):
As @ruflin already stated, please try to add the `condition` field to some test packages, so we can see it present in HTTP responses. To regenerate expected test output use the `-generate` switch (`go test . -generate`).

nchaulet:
@elasticmachine merge upstream

mtojek:
> @elasticmachine merge upstream

I'm not quite convinced that it works here :) used to perform it manually.

CHANGELOG.md
```diff
@@ -12,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
-* Add input-level `template_path` field. [#655](https://github.com/elastic/package-registry/pull/655)
+- Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
```
mtojek, line 16: Out of curiosity - why did you refactored the CHANGELOG :) ? I'm not complaining, but maybe you can push it in a separate PR? /cc @ruflin

testdata/generated/search-all.json
```diff
@@ -69,6 +69,16 @@
     "download": "/epr/input_level_templates/input_level_templates-1.0.0.zip",
     "path": "/package/input_level_templates/1.0.0"
   },
+  {
+    "name": "input_stream_condition",
+    "title": "Input and stream condtion package",
```
mtojek, line 74: nit: typo

mtojek (approved):
I'm wondering if the package-registry shouldn't filter package list based on the condition (just a thought).

LGTM, although let's double-check it with @ycombinator or @ruflin .

CHANGELOG.md
nchaulet: Oh sorry I did not check, I think my editor format on save did this, I will revert this

nchaulet:
While working on this PR I am wondering if we really need a new variables for input condition or if we could put this in the agent template, I started the discussion here I would love to have other thoughts on that https://github.com/elastic/package-spec/issues/86#issuecomment-740152447

nchaulet:
Closed as we will be using agent template for condition as discussed here https://github.com/elastic/package-spec/issues/86

PR #660, opened 2020-12-08
Add hidden field to data stream

nnamdifrankie:
Issue: elastic/security-team#548

This PR adds hidden property to data stream. The hidden field allows fleet to mark the data stream as hidden elastic/elasticsearch#63987 . This hidden feature is needed for issue 548 to hide the data stream from views.

### `hidden` field

Further, the hidden field will correspond with this ES change: elastic/elasticsearch#63987 . We want to give the flexibility for package authors to denote a data stream as hidden for purposes such as this where a data stream is not intended for app specific usage and should be hidden to reduce user confusion.

Our specific use case, again, has to do with a data stream that is intended to be queried by a separate telemetry server. As a result, we want the data stream to be hidden to reduce user confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
 * Add input-level `template_path` field. [#655](https://github.com/elastic/package-registry/pull/655)
-* Add hidden field to data stream. [#660] (https://github.com/elastic/package-registry/pull/660)
+* Add "hidden" field to data stream. [#660] (https://github.com/elastic/package-registry/pull/660)
 ### Deprecated
 
 ### Known Issue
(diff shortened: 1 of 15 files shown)


mtojek (commented):
I have a similar comment (see: https://github.com/elastic/package-registry/pull/657#pullrequestreview-547952048) for this PR.

kevinlog:
@mtojek added the test package, see this comment https://github.com/elastic/package-registry/pull/657#issuecomment-742081774

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Add "traces" as legal event type. [#656](https://github.com/elastic/package-registry/pull/656)
 * Add input-level `template_path` field. [#655](https://github.com/elastic/package-registry/pull/655)
-
+* Add hidden field to data stream. [#660] (https://github.com/elastic/package-registry/pull/660)
```
mtojek, line 17: nit: `hidden` or "hidden" 

in this case a hidden field may sound confusing :)

mtojek (approved):
LGTM!

PR #663, opened 2020-12-10
Bump changelog and release for 0.15.0

kevinlog:
This is required for the latest fields in the Endpoint package: https://github.com/elastic/endpoint-package/pull/118

"hidden" field is required to hide a `data stream` used for telemetry
"ilm_policy" field is required to define a custom `ilm_policy` used for telemetry


mtojek (approved):
Ship it!

Issue #653, opened 2020-11-02
Reduce total time of booting up the registry


mtojek:
As the number of packages is growing the package-registry needs more time to validate and load all packages before starting the HTTP server.

There are multiple ways of fixing this issue, including disabling package verification, loading packages once needed or exposing two separate endpoints for health and readiness.

mtojek:
I looked into the source code - 

@ruflin do you think we can disable the validation in package-registry (or add a switch for that in package-storage Docker images)? Packages are validated on the Integrations level, I didn't see any situation in which the package-registry notified about the problem. It will result in decreasing the total boot time.

The other option, separate endpoints for health and readiness, will play better with k8s orchestrator, but it won't decrease the deployment time, so I would vote for the first option.

ruflin:
If we have a "distribution" package with all the packages backed in, ++ on disabling the validation and doing it during the Docker build.

I think there is still value for validation in the case the registry is run to read in packages from a volume where validation was not done before. Could we make validation just a flag?

I assume package-registry still has its own validation or did we switch over the package-spec?

mtojek:
> If we have a "distribution" package with all the packages backed in, ++ on disabling the validation and doing it during the Docker build.

Yes, that's the plan.

> I think there is still value for validation in the case the registry is run to read in packages from a volume where validation was not done before. Could we make validation just a flag?

Yes, I was thinking about the flag, which is disabled in the distribution mode (as all packages have been already validated).

> I assume package-registry still has its own validation or did we switch over the package-spec?

It's still there as we need to research if there (where) are validation gaps, so we can switch to package-spec. In general the package-spec catches 99% package errors.

ruflin:
Would be great to get rid of the package-registry own implementation for validation and have parity with package-spec.

mtojek:
Let me split into half and work on the flag first.

Regarding switching to the package-spec - we have to remember that this will introduce another upstream dependency in the package-registry, 3 branches of the package-storage and integrations. In other words, every commit pushed to the package-spec will trigger an avalanche of updates. I wonder if we can louse coupling here or simply accept it (/cc @ycombinator ).

ruflin:
At the moment the "problem" is that each commit means a release (more or less) and in general a package-spec change is driven by a very specific need to the release has to happen immeditately. I wonder if over time the releases become less frequent? I personally like that we don't magically update but avalanche of updates is not great. Automation possible maybe?

ycombinator:
[...]

PR #667, opened 2021-01-04
Switch for disabling validation

mtojek:
Issue: https://github.com/elastic/package-registry/issues/653

This PR introduces a switch for disabling package validation. In public image builds of package-storage the validation will be shifted from runtime to build time.

Changes:
- added switch to disable package validation
- refactored `Validate()` methods to become idempotent

Unit tests are passing.

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,6 +46,7 @@ func init() {
 	flag.StringVar(&address, "address", "localhost:8080", "Address of the package-registry service.")
 	// This flag is experimental and might be removed in the future or renamed
 	flag.BoolVar(&dryRun, "dry-run", false, "Runs a dry-run of the registry without starting the web service (experimental)")
+	flag.BoolVar(&util.EnablePackageValidation, "validate", false, "Validate package content")
 }
 
 type Config struct {
@@ -60,6 +61,10 @@ func main() {
 	flag.Parse()
 	log.Println("Package registry started.")
 	defer log.Println("Package registry stopped.")
+
+	if dryRun {
+		util.EnablePackageValidation = true // dry-run enables the package validation
+	}
 
 	config := mustLoadConfig()
 	packagesBasePaths := getPackagesBasePaths(config)
--- a/main.go
+++ b/main.go
@@ -46,7 +46,7 @@ func init() {
 	flag.StringVar(&address, "address", "localhost:8080", "Address of the package-registry service.")
 	// This flag is experimental and might be removed in the future or renamed
 	flag.BoolVar(&dryRun, "dry-run", false, "Runs a dry-run of the registry without starting the web service (experimental)")
-	flag.BoolVar(&util.EnablePackageValidation, "validate", false, "Validate package content")
+	flag.BoolVar(&util.PackageValidationDisabled, "disable-package-validation", false, "Disable package content validation")
 }
 
 type Config struct {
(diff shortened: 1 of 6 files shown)


ruflin (commented):
Should it be enabled or disabled by default? I wonder if we have less reported issues if we enable it by default. If disabled by default, the registry might have issues because packages are not valid. 

Everyone just getting started with the registry and running it for some of the packages should have validation enabled. The only place we need to disable it is for the production environment because we know there, we already have validation in place. 

That is why I would switch the defaults.

mtojek:
I think we can go this way too, might be more intuitive and backwards compatible. I adjusted the PR and it's description.

main.go
```diff
@@ -46,6 +46,7 @@ func init() {
 	flag.StringVar(&address, "address", "localhost:8080", "Address of the package-registry service.")
 	// This flag is experimental and might be removed in the future or renamed
 	flag.BoolVar(&dryRun, "dry-run", false, "Runs a dry-run of the registry without starting the web service (experimental)")
+	flag.BoolVar(&util.PackageValidationDisabled, "disable-package-validation", false, "Disable package content validation")
```
ruflin, line 49: Quite a long name for a flag but no better suggestion.

util/data_stream.go
```diff
@@ -151,22 +151,11 @@ func NewDataStream(basePath string, p *Package) (*DataStream, error) {
...
-	if strings.Contains(d.Dataset, "-") {
-		return fmt.Errorf("data stream name is not allowed to contain `-`: %s", d.Dataset)
-	}
-
-	if !d.validType() {
-		return fmt.Errorf("type is not valid: %s", d.Type)
+		return nil, err
 	}
 
 	if d.Elasticsearch != nil && d.Elasticsearch.IngestPipelineName == "" {
```
ruflin, line 161: This is kind of a mix between validation and building the data model. I wonder if we should clean it up later or keep it as is. It would probably mean doing parts of it twice.

ruflin (approved):
LGTM

main.go
mtojek: It's one of these feature flag that you should be discouraged to use :)

util/data_stream.go
mtojek: I'm not sure if I understand your concerns. Any code that modifies a package or a datastream is placed in constructors (e.g. `NewDataStream`). Are you referring to this part of code: https://github.com/elastic/package-registry/blob/b20850de456bf1a0e363163792a7178a4d0c3358/util/data_stream.go#L161-L183 ? If so, then we should come back here and refactor, as the `DataStream.IngestPipeline` is marked as deprecated.

I suppose we can strip this part of at all, unless there is Kibana logic using it.

ruflin: What I initially assume could be an issue does not seem to be one after quickly checking the code. I was worried that for example the pipeline checks do not happen if only `Validate` is run. But this is not the case as before validate is run, the constructor is used which runs all these additional checks always to build the actual package object. So all good.

mtojek: Yeah, I know the design is not really clear as we have to sneak the condition into the `Validate` function, but I didn't see any `Option` to disable validation at `Unpack`.

ruflin: I think I know the person who built this initially and is to blame for the design 😆

PR #671, opened 2021-01-18
Add streaming category

ChrsMark:
This PR adds `streaming` category useful to tag `stan` package, or `NATS Streaming`.
stan PR: https://github.com/elastic/integrations/pull/532
stan docs: https://github.com/nats-io/stan.go

### DIFF
--- a/util/package.go
+++ b/util/package.go
@@ -44,6 +44,7 @@ var CategoryTitles = map[string]string{
 	"os_system":         "OS & System",
 	"productivity":      "Productivity",
 	"security":          "Security",
+	"streaming":         "Data Streaming"
 	"support":           "Support",
 	"ticketing":         "Ticketing",
 	"version_control":   "Version Control",
--- a/util/package.go
+++ b/util/package.go
@@ -44,7 +44,7 @@ var CategoryTitles = map[string]string{
 	"os_system":         "OS & System",
 	"productivity":      "Productivity",
 	"security":          "Security",
-	"streaming":         "Data Streaming"
+	"streaming":         "Data Streaming",
 	"support":           "Support",
 	"ticketing":         "Ticketing",
 	"version_control":   "Version Control",


util/package.go
```diff
@@ -44,6 +44,7 @@ var CategoryTitles = map[string]string{
 	"os_system":         "OS & System",
 	"productivity":      "Productivity",
 	"security":          "Security",
+	"streaming":         "Data Streaming"
```
mtojek, line 47: Missing trailing comma

mtojek (commented):
Is this change coordinated with product team or just your suggestion? I'm wondering if there is some work needed on the Kibana side.

ChrsMark:
> Is this change coordinated with product team or just your suggestion? I'm wondering if there is some work needed on the Kibana side.

It's just a suggestion to enrich the tags to better match the specific integration since this is one is better identified as a streaming platform rather than a message queue.

mtojek:
> It's just a suggestion to enrich the tags to better match the specific integration since this is one is better identified as a streaming platform rather than a message queue.

@ruflin @skh do you think it's safe to push this change (won't blow up anything in Kibana)?

mtojek:
@ChrsMark  once we agree on this change, you'll have to update the package-spec as well: https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L64 (then update elastic-package on the spec, but I can help you with this).

ruflin:
These categories are synced with https://www.elastic.co/integrations so I would like to pull in @asawariS to chime in if we can just add a category here.

mtojek:
> These categories are synced with https://www.elastic.co/integrations so I would like to pull in @asawariS to chime in if we can just add a category here.

Fun fact: these tiles hyperlink to Beats documentation (a bit irrelevant).

asawariS:
@ruflin - Thanks for flagging. I defer to @tbragin @mostlyjason and other PMs on the tags here. Even though the tags on elastic.co/integrations are not directly currently pulled from the EPR, we eventually want them to be. I agree that there needs to be some process / discussion before new tags are added -- I think PM/Eng is best suited to define that process. we will just consume the final decision on the web page. 

> Fun fact: these tiles hyperlink to Beats documentation (a bit irrelevant). 

@mtojek good catch. When the page was created, most integrations (via agent) didn't exist, so we pointed to Beats modules in most places, etc. There was also the case of do we point people to beta features (not supported in prod) or the GA module version. 

Can point me to the appropriate link in the docs for Nats integration. I will discuss with PMs and take point on updating the links for this and other integrations that are now available via Fleet/Agent, if that's the route we go with.

mtojek:
> Can point me to the appropriate link in the docs for Nats integration. I will discuss with PMs and take point on updating the links for this and other integrations that are now available via Fleet/Agent, if that's the route we go with.

The only (so far best) place you can find documentation for an integration without booting up Kibana is Github:

https://github.com/elastic/integrations/tree/master/packages/nats

mostlyjason:
We have a policy to coordinate changes to these categories through ECS. @jamiehynds already has an open PR for updates to these categories in ECS so I recommending syncing with him on this request https://github.com/elastic/ecs/pull/958. I'd recommend moving this request to the ECS repo to coordinate it.

ruflin:
@mostlyjason These are NOT the same categories. What is in the package-spec is what is also on the elastic.co/integrations page: https://github.com/elastic/package-registry/blob/3489847b9372cfe2dc4b4220b2dedd6ced75bb28/util/package.go#L27 There have been quite a few historical discussions on this (private links, please ping me).

The question here is: Should we add it to our categories and if yes, it means we must also add it to /integrations. package-spec is currently the source of truth for categories and I think it should stay the source of truth.

tbragin:
What in addition to "NATS Streaming" be part of the "streaming" category? I'd prefer to be quite conservative in adding new categories, we already have quite a few: https://www.elastic.co/integrations As it stands, I think we can just leave "NATS Streaming" in the "Kubernetes" category, no?

ChrsMark:
Ok it seems that we only have this one for now in this category and I can only think of Redis Streaming if we decide to add it in the feature. So leaving it out sounds good to me @tbragin, thanks! I will update the package's description properly to indicate the streaming functionality as well as tag it with k8s category.

mtojek:
Thank you Christos for checking this. I will resolving the issue.

Issue #672, opened 2021-01-20
Package registry doesn't load the package that has accidental .DS_Store file in the root.


aleksmaus:
Package registry doesn't load the package that has accidental .DS_Store file in the root.
It fails silently and doesn't output any error log. 
It should at least log the cause why the package can't be loaded.

ycombinator:
@aleksmaus, thanks for filing this issue and the corresponding PR to fix it in the package registry.

I'm curious, however, to know what steps you took to load a package with a `.DS_Store` file into the package registry. I checked the `.gitignore` files in all the relevant branches of https://github.com/elastic/package-storage and I see that `.DS_Store` is already mentioned in there. So I assume you ran into this issue while developing and serving a package with the package registry locally? Would you mind posting the steps you took to run into this issue? Perhaps there's tooling we can fix or enhance to detect and clean up such files during local development. Thanks!

aleksmaus:
the prevention you mentioned would be nice, but would not work in my case, because I just browsed to the local packages dir with Finder app (that could create the .DS_Store file) and then it took a bit of time to figure why my package stopped loading. The ```warning``` log message that I added in the PR would be helpful anyways.

ycombinator:
Gotcha. Agreed then that when the package registry loads the package is the earliest time we could catch this issue. Thanks for explaining!

PR #673, opened 2021-01-21
Fix the package not loading if it has an accidental file left in the package root directory.

aleksmaus:
Fix the package not loading if it has an accidental file in the package root directory.
Log the error and load the remaining sibling directories.

Addresses https://github.com/elastic/package-registry/issues/672

Screen Shot 2021-01-20 at 9 11 18 PM

### DIFF
--- a/util/packages.go
+++ b/util/packages.go
@@ -74,14 +74,13 @@ func getPackagePaths(allPaths []string) ([]string, error) {
 			if info.IsDir() {
 				log.Printf("%-20s\t%10s\t%s", dirs[0], dirs[1], path)
 				foundPaths = append(foundPaths, path)
-			} else {
-				// Not expected file, return nil in order to continue processing sibling directories
-				// Fixes an annoying problem when the .DS_Store file is left behind and the package
-				// is not loading without any error information
-				log.Printf("error: unexpected file: %s", path)
-				return nil
+				return filepath.SkipDir
 			}
-			return filepath.SkipDir // don't need to go deeper
+			// Not expected file, return nil in order to continue processing sibling directories
+			// Fixes an annoying problem when the .DS_Store file is left behind and the package
+			// is not loading without any error information
+			log.Printf("error: unexpected file: %s", path)
+			return nil
 		})
 		if err != nil {
 			return nil, errors.Wrapf(err, "listing packages failed (path: %s)", packagesPath)
--- a/util/packages.go
+++ b/util/packages.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


util/packages.go
```diff
@@ -74,6 +74,12 @@ func getPackagePaths(allPaths []string) ([]string, error) {
 			if info.IsDir() {
 				log.Printf("%-20s\t%10s\t%s", dirs[0], dirs[1], path)
 				foundPaths = append(foundPaths, path)
+			} else {
+				// Not expected file, return nil in order to continue processing sibling directories
+				// Fixes an annoying problem when the .DS_Store file is left behind and the package
+				// is not loading without any error information
+				log.Printf("error: unexpected file: %s", path)
+				return nil
```
ruflin, line 82: I wonder if we should return an error instead. This is at first inconvenient for the developer working with the package because he has to fix it but it assures no "invalid" files are inside the package.

@mtojek @ycombinator Do we have any validation in place in `elastic-package check` that would detect these files?

ruflin (commented):
Can you add a changelog entry?

util/packages.go
mtojek: I'm wondering how did @aleksmaus run into this issue? Have you been using the elastic-package to boot up the stack? Did you interact with OSX file browser?

@ruflin This directory is not part of a package (look at the path). It's a bug in package-registry. Speaking about validation, there is such one in place, but there are places in which it's not a bug but a feature, like e.g. a package root. It's hard to detect these files as the end-user may include additional READMEs, license notes, etc. 

For reference: https://github.com/elastic/package-spec/blob/master/versions/1/spec.yml#L8 

@aleksmaus I don't think we need to fail the entire application in case of noticing strange files. It will be more convenient to the end-user to skip this files. Could you try this piece of code?

```go
if info.IsDir() {
    log.Printf("%-20s\t%10s\t%s", dirs[0], dirs[1], path)
    foundPaths = append(foundPaths, path)
    return filepath.SkipDir
}
return nil
```

aleksmaus: Mac OS finder creates these pesky .DS_Store files for Finder app. 
This change doesn't fail the whole app, it keeps loading the package versions. 
Before the change, stumbling on such file (not dir code path) was returning filepath.SkipDir which skips all the siblings directories. 
In my case the directory layout was 
```
.DS_Store
0.1.0
0.1.2
```

And the existing code was stumbling upon .DS_Store file first and since it was not directory the default code path was always returning filepath.SkipDir

From documentation of filepath.Walk: 
```
 If the function returns SkipDir when invoked on a non-directory file, Walk skips the remaining files in the containing directory.
 ```
 
 Sure can update the code  the sample you provided, it's pretty much the same what my code does.
 Would still like to keep the log statement pointing at the error during load.

**Comment by aleksmaus** [comment]
@mtojek @ruflin is this good to merge now?

File: util/packages.go [hunk 2]
```diff
@@ -74,8 +74,13 @@ func getPackagePaths(allPaths []string) ([]string, error) {
 			if info.IsDir() {
```
ycombinator, line 74: I would take this opportunity to make this check a bit stronger.

If the execution reaches this line, we know that we should be looking at a directory whose name is a version. So how about we not only check if `info.IsDir()` but also check that `dirs[1]` [is a semver](https://godoc.org/github.com/Masterminds/semver#StrictNewVersion)?  That way we handle not only the unexpected _file_ case but also the unexpected _folder_ case as well.

util/packages.go
```diff
@@ -74,8 +74,13 @@ func getPackagePaths(allPaths []string) ([]string, error) {
...
 				log.Printf("%-20s\t%10s\t%s", dirs[0], dirs[1], path)
 				foundPaths = append(foundPaths, path)
+				return filepath.SkipDir
 			}
-			return filepath.SkipDir // don't need to go deeper
+			// Not expected file, return nil in order to continue processing sibling directories
+			// Fixes an annoying problem when the .DS_Store file is left behind and the package
+			// is not loading without any error information
+			log.Printf("error: unexpected file: %s", path)
```
ycombinator, line 82: Nit: consider changing this to a `warning` instead of an `error` and also mention that we're simply ignoring the unexpected file. That way the person looking at these logs knows this is not a severe problem to worry about.

aleksmaus:
Updated the log message and added additional check if dir is not semver.
Screen Shot 2021-01-26 at 8 13 13 AM

ycombinator (approved):
LGTM. Thanks for finding and fixing this!

ycombinator:
@aleksmaus Apologies for not catching this earlier, but would you mind making a follow up PR to add a CHANGELOG entry for the change in this PR? Thank you.

https://github.com/elastic/package-registry/blob/master/CHANGELOG.md

PR #674, opened 2021-01-26
Add "dataset_is_prefix" field to data_stream, if present.

skh:
This change exposes the new field `dataset_is_prefix` so that it can be used by Kibana/EPM.

The field was added to package-spec in https://github.com/elastic/package-spec/pull/102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add "dataset_is_prefix" field to data stream. [#674] (https://github.com/elastic/package-registry/pull/674)
+
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


ycombinator:
Hi @skh, could you try running `mage format` in the root folder?

ycombinator (changes requested):
Could you add a CHANGELOG entry for this change, please? Thanks!

https://github.com/elastic/package-registry/blob/master/CHANGELOG.md

ycombinator (approved):
LGTM.

ruflin:
Would be nice to have a follow up adding this to one of the test packages: https://github.com/elastic/package-registry/tree/master/testdata/package Then we can see how the actual output looks.

skh:
> Would be nice to have a follow up adding this to one of the test packages

https://github.com/elastic/package-registry/pull/676

PR #680, opened 2021-01-29
Updating README

ycombinator:
Updating README to add more tasks to do after cutting a new package registry release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,11 +126,8 @@ Afterwards the service can be started:
 
 The following endpoints exist:
 
-* prod, no CDN: https://epr.ea-web.elastic.dev
 * prod, CDN: https://epr.elastic.co
-* staging, no CDN: https://epr-staging.ea-web.elastic.dev
 * staging, CDN: https://epr-staging.elastic.co
-* snapshot, no CDN: https://epr-snapshot.ea-web.elastic.dev/
 * snapshot, CDN: https://epr-snapshot.elastic.co/
 
 An dev registry is running on `https://epr-staging.elastic.co/`. This is updated from time to time to be in sync with master.
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ The following endpoints exist:
 * prod, CDN: https://epr.elastic.co
 * staging, no CDN: https://epr-staging.ea-web.elastic.dev
 * staging, CDN: https://epr-staging.elastic.co
-* experimental, no CDN: https://epr-experimental.ea-web.elastic.dev/
-* experimental, CDN: https://epr-experimental.elastic.co/
+* snapshot, no CDN: https://epr-snapshot.ea-web.elastic.dev/
+* snapshot, CDN: https://epr-snapshot.elastic.co/
 
 An dev registry is running on `https://epr-staging.elastic.co/`. This is updated from time to time to be in sync with master.
 
--- a/README.md
+++ b/README.md
@@ -191,4 +191,7 @@ New versions of the package registry need to be released from time to time. The 
 
 CI automatically creates a new Docker image which will be available under `docker.elastic.co/package-registry/package-registry:vA.B.C` a few minutes after creating the tag.
 
-As a new registry is normally released to bring new features to the package-storage, follow the docs in the [Package Storage](https://github.com/elastic/package-storage#update-package-registry-for-a-distribution) repository on how to update the distributions.
+After the new registry Docker image is available, update the following projects that consume it:
+- Integrations: Update the version of the Package Registry Docker image as shown in this [sample PR](https://github.com/elastic/integrations/pull/581).
+- Package Storage: Follow the [documentation](https://github.com/elastic/package-storage#update-package-registry-for-a-distribution) repository on how to update the distributions.
+- Kibana: Do this only after all Package Storage distributions have been updated and released as new Docker images. Update the version of the Package Registry Distribution Docker image as shown in this [sample PR](https://github.com/elastic/kibana/pull/89776).
--- a/README.md
+++ b/README.md
@@ -124,11 +124,16 @@ Afterwards the service can be started:
 
 **Deployment**
 
-The following endpoints exist:
+The following **active** endpoints exist:
 
 * prod, CDN: https://epr.elastic.co
 * staging, CDN: https://epr-staging.elastic.co
 * snapshot, CDN: https://epr-snapshot.elastic.co/
+
+Additionally, the following **frozen** endpoints exist and are **no longer updated**:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,4 +191,7 @@ New versions of the package registry need to be released from time to time. The
 
 CI automatically creates a new Docker image which will be available under `docker.elastic.co/package-registry/package-registry:vA.B.C` a few minutes after creating the tag.
 
-As a new registry is normally released to bring new features to the package-storage, follow the docs in the [Package Storage](https://github.com/elastic/package-storage#update-package-registry-for-a-distribution) repository on how to update the distributions.
+After the new registry Docker image is available, update the following projects that consume it:
+- Integrations: Update the version of the Package Registry Docker image as shown in this [sample PR](https://github.com/elastic/integrations/pull/581).
+- Package Storage: Follow the [documentation](https://github.com/elastic/package-storage#update-package-registry-for-a-distribution) on how to update the distributions.
+- Kibana: Do this only after all Package Storage distributions have been updated and released as new Docker images. Update the version of the Package Registry Distribution Docker image as shown in this [sample PR](https://github.com/elastic/kibana/pull/89776).
```
ruflin, line 197: Good question on how often we should update these? Each time a stack release goes out, we should tag a version of the package-storage which then also can be used in Kibana. This makes it more readable. Unfortunately we didn't do a great job at this so far :-(

README.md
```diff
@@ -130,8 +130,8 @@ The following endpoints exist:
 * prod, CDN: https://epr.elastic.co
 * staging, no CDN: https://epr-staging.ea-web.elastic.dev
```
mtojek, line 131: I wonder if we shouldn't keep only endpoints with CDN.

README.md
```diff
@@ -130,8 +130,8 @@ The following endpoints exist:
...
 * staging, CDN: https://epr-staging.elastic.co
-* experimental, no CDN: https://epr-experimental.ea-web.elastic.dev/
-* experimental, CDN: https://epr-experimental.elastic.co/
```
mtojek, line 134: Honestly I would leave it here (as it still exists/live), but would add a description on why is it deprecated.

README.md
mtojek: What do you think about changing this dependency in Kibana to look at the latest one? We don't expect any breaking changes on the EPR side.

README.md
ruflin: I think for debugging purpose it is quite nice to have these documented.

README.md
ruflin: I quite like tying it to a specific version and then update is manually. This is especially true for the release branches where the testing should be fully predictable.

mtojek: Yeah, but Kibana depends on the package-storage, not package-registry. We don't have a consistent versioning there and we don't need it.

README.md
```diff
@@ -191,4 +191,7 @@ New versions of the package registry need to be released from time to time. The
 
 CI automatically creates a new Docker image which will be available under `docker.elastic.co/package-registry/package-registry:vA.B.C` a few minutes after creating the tag.
 
-As a new registry is normally released to bring new features to the package-storage, follow the docs in the [Package Storage](https://github.com/elastic/package-storage#update-package-registry-for-a-distribution) repository on how to update the distributions.
+After the new registry Docker image is available, update the following projects that consume it:
+- Integrations: Update the version of the Package Registry Docker image as shown in this [sample PR](https://github.com/elastic/integrations/pull/581).
```
mtojek, line 195: This is needed only for scripts (import-beats, update-package-storage). It's not serious crime if we forget about bumping it there.

README.md
ruflin: > We don't have a consistent versioning there and we don't need it.

As mentioned above, I think we should have tags for the production ones. Not for the release itself but exactly for testing or running on prem in the future.

README.md
ycombinator: As discussed off-PR, we remove them from this particular README.

README.md
ycombinator: In that case, there is also a `7.9` endpoint IIRC. Should we document that here as well?

ruflin: yes

README.md
ycombinator: Just so I'm super clear, @ruflin, you're referring to adding another Docker tag on the `package-storage` image besides `snapshot`, `production`, etc. that refers to the `package-registry` version that image was built with, correct?

If so, it looks like we already add a *label* for this information: https://github.com/elastic/package-storage/blob/5314869e2f6bc01d37b8652f7bda89248950b3a4/Dockerfile#L8

But since images can't be pulled down using labels, it's just a matter of also adding the same label as a tag as well, e.g. `snapshot-0.17.0`, `production-0.17.0`, etc.?

ruflin: No :-) What I'm looking for is a tag that is aligned with the stack release like we have for 7.9.0: https://github.com/elastic/package-storage/tree/7.9.0 So this means, we should have created a tag `7.11.0` on the day of the 7.11 Elastic Stack release. This is documented here: https://github.com/elastic/package-storage#tags

ycombinator: Ah okay, thanks for clarifying :) So this will need to be a completely manual process or can we make it part of the release manager CI job?

ruflin: Would be great to have it automated. How are tags in other repos created?

ycombinator: I've created an infra issue to ask about automating this and tagged you both on it.

README.md
ycombinator: Removed in 19c82f59e313373c59b61f358ceeded7f59655fa.

README.md
ycombinator: Added in 2e0681981a785386de076b46b64fd57089b2d2e5.

mtojek (approved):

README.md
ruflin: @ycombinator I think there are already issues and @kuisathaverat is working on this already.

kuisathaverat: > Would be great to have it automated. How are tags in other repos created?

It depends, we have two kinds:
* The tag is created manually on the repo and it triggers a release process
* A job is triggered in the CI that creates a tag and make a release process

We are working on automerge the package-storage PRs and trigger the deployment automatically https://github.com/elastic/package-storage/issues/592, I think is not the same.

ruflin: @ycombinator @kuisathaverat Sorry, I mixed two things here. You are right.

ycombinator: I'm going to merge this PR with what we have here. If/when we have automation, we can always come back to this README and remove the `Kibana` line or update it to reflect the reality of that time.

PR #681, opened 2021-02-02
Add base data stream into package

kaiyan-sheng:
(just a draft PR for my small poc)
This PR is to adjust the code to read the new config option `expand_data_stream` at the package level. When `expand_data_stream` is set to true, abstract data stream info and store them into the `/search` API endpoint result.

For example with `expand_data_stream: true` in AWS package, `/search` API response will look like this:
```
[
  {
    "name": "aws",
    "title": "AWS",
    "version": "0.3.18",
    "release": "beta",
    "description": "AWS Integration",
    "type": "integration",
    "download": "/epr/aws/aws-0.3.18.zip",
    "path": "/package/aws/0.3.18",
    "icons": [
      {
        "src": "/img/logo_aws.svg",
        "path": "/package/aws/0.3.18/img/logo_aws.svg",
        "title": "logo aws",
        "size": "32x32",
        "type": "image/svg+xml"
      }
    ],
    "base_data_streams": [
      ...
      {
        "type": "metrics",
        "dataset": "aws.ec2_metrics",
        "title": "AWS EC2 metrics",
        "release": "beta",
        "description": "AWS EC2 Integration",
        "icons": [
          {
            "src": "../img/logo_aws_ec2.svg",
            "path": "",
            "title": "logo aws ec2",
            "size": "32x32",
            "type": "image/svg+xml"
          }
        ]
      },
      ...
    ]
  }
]
```

### DIFF
--- a/util/data_stream.go
+++ b/util/data_stream.go
@@ -42,6 +42,9 @@ type DataStream struct {
 	Title   string `config:"title" json:"title" validate:"required"`
 	Release string `config:"release" json:"release"`
 
+	Description string  `config:"description" json:"description"`
+	Icons       []Image `config:"icons,omitempty" json:"icons,omitempty" yaml:"icons,omitempty"`
+
 	// Deprecated: Replaced by elasticsearch.ingest_pipeline.name
 	IngestPipeline string         `config:"ingest_pipeline,omitempty" config:"ingest_pipeline" json:"ingest_pipeline,omitempty" yaml:"ingest_pipeline,omitempty"`
 	Streams        []Stream       `config:"streams" json:"streams,omitempty" yaml:"streams,omitempty" `
@@ -53,6 +56,17 @@ type DataStream struct {
 
 	// Local path to the package dir
 	BasePath string `json:"-" yaml:"-"`
+}
+
+type BaseDataStream struct {
+	// Name and type of the data stream. This is linked to data_stream.dataset and data_stream.type fields.
+	Type    string `config:"type" json:"type" validate:"required"`
+	Dataset string `config:"dataset" json:"dataset,omitempty" yaml:"dataset,omitempty"`
+
+	Title       string  `config:"title" json:"title" validate:"required"`
+	Release     string  `config:"release" json:"release"`
+	Description string  `config:"description" json:"description"`
+	Icons       []Image `config:"icons,omitempty" json:"icons,omitempty" yaml:"icons,omitempty"`
 }
 
 type Input struct {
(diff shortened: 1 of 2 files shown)


util/data_stream.go
```diff
@@ -55,6 +58,17 @@ type DataStream struct {
 	BasePath string `json:"-" yaml:"-"`
 }
 
+type BaseDataStream struct {
```
mtojek, line 61: I was wondering why is it a "base" data stream. I got the idea, but can't relate this name with meaning.

util/data_stream.go
```diff
@@ -97,19 +111,19 @@ type fieldEntry struct {
 	aType string
 }
 
-func NewDataStream(basePath string, p *Package) (*DataStream, error) {
+func NewDataStream(basePath string, p *Package) (*DataStream, *BaseDataStream, error) {
```
mtojek, line 114: The `util` package contains structs and methods which can be used in other repositories. I woudn't change signatures of exported methods.

util/data_stream.go
```diff
@@ -42,6 +42,9 @@ type DataStream struct {
 	Title   string `config:"title" json:"title" validate:"required"`
 	Release string `config:"release" json:"release"`
 
+	Description string  `config:"description" json:"description"`
```
mtojek, line 45: In general `BasePackage` and `DataStream` correspond to their format in package-spec. I wouldn't extend them this way. If you need to add more fields to view, but not manifest, I would consider creating new structures (possibly extending ones in spec).

util/data_stream.go
```diff
@@ -55,6 +58,17 @@ type DataStream struct {
 	BasePath string `json:"-" yaml:"-"`
 }
 
+type BaseDataStream struct {
+	// Name and type of the data stream. This is linked to data_stream.dataset and data_stream.type fields.
+	Type    string `config:"type" json:"type" validate:"required"`
+	Dataset string `config:"dataset" json:"dataset,omitempty" yaml:"dataset,omitempty"`
```
mtojek, line 64: Looks like redundant information?

util/package.go
```diff
@@ -67,20 +67,23 @@ type Package struct {
...
+	Title           *string           `config:"title,omitempty" json:"title,omitempty" yaml:"title,omitempty"`
+	Version         string            `config:"version" json:"version"`
+	Release         string            `config:"release,omitempty" json:"release,omitempty"`
+	Description     string            `config:"description" json:"description"`
+	Type            string            `config:"type" json:"type"`
+	Download        string            `json:"download" yaml:"download,omitempty"`
+	Path            string            `json:"path" yaml:"path,omitempty"`
+	Icons           []Image           `config:"icons,omitempty" json:"icons,omitempty" yaml:"icons,omitempty"`
+	Internal        bool              `config:"internal,omitempty" json:"internal,omitempty" yaml:"internal,omitempty"`
+	BaseDataStreams []*BaseDataStream `config:"base_data_streams,omitempty" json:"base_data_streams,omitempty" yaml:"base_data_streams,omitempty"`
```
mtojek, line 86: `BaseDataStream` contains redundant data that you can find in package. I think we should keep them only in package to prevent the response body from increasing.

BTW base data streams look to me more like data stream previews or sth (just looking for a different name :)

mtojek (commented):
TL;DR Remember, the spec goes first :)

There are few steps to push this change:

1. Package-spec.
2. API design of package-registry.
3. Implementation of package-registry.

Re 1:
I think we all agree with `expand_data_streams`. Do you want to introduce images/screenshots on the data stream level? You can open a PR to package-spec and we can discuss it there.

Re 2:
I would consider two approaches:
1. Introduce data streams previews (`BaseDataStream`) which contain as min information as possible comparing to the owning package.

Re 3:
Once we agree on 1. and 2., this is a good moment to go with implementation of the package-registry.

ruflin (changes requested):
Lets put this on hold until we completed the discussion on package-spec.

PR #688, opened 2021-04-12
Fix: define synthetics type

mtojek:
This PR defines the missing "synthetics" type.

Spotted in: https://github.com/elastic/package-storage/pull/1140


mtojek:
self-approved

PR #691, opened 2021-04-26
Bump elastic stack version automation

v1v:
## What does this PR do?

Script to bump the versions:

```diff
## I changed the sh script to exist after the sed
$ ./.ci/bump-stack-version.sh 8.0.0-3beeb19e
Update stack with version 8.0.0-3beeb19e
diff --git a/testing/environments/snapshot.yml b/testing/environments/snapshot.yml
index dbbade2f..f79c5e95 100644
--- a/testing/environments/snapshot.yml
+++ b/testing/environments/snapshot.yml
@@ -3,7 +3,7 @@
 version: '2.3'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.0.0-3beeb19e-SNAPSHOT
     healthcheck:
       test: ["CMD", "curl", "-f", "-u", "elastic:changeme", "http://127.0.0.1:9200/"]
       retries: 300
@@ -20,7 +20,7 @@ services:
     - "ELASTIC_PASSWORD=changeme"
 
   kibana:
-    image: docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT
+    image: docker.elastic.co/kibana/kibana:8.0.0-3beeb19e-SNAPSHOT
     healthcheck:
       test: "curl -f http://localhost:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null"
       retries: 600
~
```

This PR is one part of the required changes to enable the automation. 

## Why is it important?

Handle the stack versions on a PR basis to ensure breaking changes are handled correctly. Unfortunately we cannot use `dependabot` since those versions are handled differently.

## Issues

Similar to https://github.com/elastic/beats/pull/25148

## Follow ups

- Enable automerge with `mergify`?

### DIFF
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Given the stack version this script will bump the version.
+#
+# This script is executed by the automation we are putting in place
+# and it requires the git add/commit commands.
+#
+# Parameters:
+#	$1 -> the version to be bumped. Mandatory.
+#	$2 -> whether to create a branch where to commit the changes to.
+#		  this is required when reusing an existing Pull Request.
+#		  Optional. Default true.
+#
+set -euo pipefail
+MSG="parameter missing."
+VERSION=${1:?$MSG}
+CREATE_BRANCH=${2:-true}
+
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
+
+if [ "${OS}" == "darwin" ] ; then
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(image: docker\.elastic\.co/.*):[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1:${VERSION}#g" testing/environments/snapshot.yml
+
+echo "Commit changes"
+if [ "$CREATE_BRANCH" = "true" ]; then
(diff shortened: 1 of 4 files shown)


testing/environments/snapshot.yml
```diff
@@ -3,7 +3,7 @@
 version: '2.3'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.10.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT
```
mtojek, line 6: Same story as in https://github.com/elastic/package-storage/pull/1205

.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Given the stack version this script will bump the version.
```
mtojek, line 3: It seems that the same script will land in multiple places. Is it possible to keep it in a single public one instead of maintaining multiple copies?

v1v: I see more cons:
1) Each project might differ.
2) The bump script will not be easy to access and all the issues related to how to test, how to version.
3) Each project branch might potentially differ in how to bump those versions, even if not now it could happen

Maybe the script should be moved to the build system instead, and delegate that particular functionality to the project itself. But that's something we can iterate on it in the future if needed.

We can revisit this in the future to refactor if needed, what do you think?

PR #695, opened 2021-05-10
[Fleet] Allow to specify kibana.version when fetching categories

nchaulet:
## Description 

Related to https://github.com/elastic/kibana/issues/94582 

The categories count in Kibana is sometimes invalid has some package could not exists depending on the kibana version.

This PR fix it by allowing to pass the `kibana.version` (same parameter name used in the search handler) query params to the category handler, and filtering package that have the correct kibana version.

## Examples 

```
➜  ~ http http://localhost:8080/categories\?experimental\=true\&kibana.version\=7.11.0 | jq '.[] | select(.id =="elastic_stack")'
{
  "id": "elastic_stack",
  "title": "Elastic Stack",
  "count": 1
}
➜  ~ http http://localhost:8080/categories\?experimental\=true | jq '.[] | select(.id =="elastic_stack")'
{
  "id": "elastic_stack",
  "title": "Elastic Stack",
  "count": 3
}
``

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Support filtering /categories using `kibana.version` query param [#695](https://github.com/elastic/package-registry/pull/695)
+
 ### Deprecated
 
 ### Known Issues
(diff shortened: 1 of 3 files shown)


mtojek (commented):
Thank you for working on the fix, nice finding BTW.

Would you mind covering relevant endpoints with unit tests? See: https://github.com/elastic/package-registry/blob/master/main_test.go#L32

Please also update the CONTRIBUTING guide.

Issue #698, opened 2021-06-22
Disable integration tests for Package Registry


mtojek:
There are contract tests present and enabled in the repository, so it seems that we don't need to keep integration tests against the whole stack (Kibana, Fleet, etc.). Let's disable them.

ruflin:
What is the concern here? Are these flaky? Taking too long?

mtojek:
Please take a look at reference thread with @jsoriano and @v1v.

Long story short - we're looking for opportunities to simplify updating references to  Elastic stack images. In this case integration tests against Kibana don't give anything much value as we have API contract tests.

ruflin:
Got it, thanks for the details. I'm hesitant to fully remove the tests as I think it still provides some value. A dev opening a PR against the repo will see on the PR itself that something is not working instead of seeing it only later through e2e tests or similar.

Let me propose an alternative. In general we are pretty good in testing the "current" version but sometimes lack on testing against older version. We know the oldest version that the package registry should work with is 7.10 (please double check) so this is the version we always test against. If we break one of the version that are currently under development, I'm pretty sure we will quickly know because of all the other test systems.

mtojek:
> Got it, thanks for the details. I'm hesitant to fully remove the tests as I think it still provides some value. A dev opening a PR against the repo will see on the PR itself that something is not working instead of seeing it only later through e2e tests or similar.

With this file we have covered all endpoints and, I think, edge cases: https://github.com/elastic/package-registry/blob/master/main_test.go . If a dev opens a PR against this repo, one of these contract tests would fail when the response body is different.

> Let me propose an alternative. In general we are pretty good in testing the "current" version but sometimes lack on testing against older version. We know the oldest version that the package registry should work with is 7.10 (please double check) so this is the version we always test against.

Do you think that there is an edge case which won't be caught with contract tests, but might be possible to catch with legacy testing? Personally I consider contract tests sacred and this is the first "blocker" for any changes in API. Keep in mind that package-registry doesn't keep any state or logic which can modify its behavior.

cc @jsoriano

jsoriano:
>> In this case integration tests against Kibana don't give anything much value as we have API contract tests.

> I'm hesitant to fully remove the tests as I think it still provides some value.

@ruflin I agree that these tests provide value, so tests like this should exist, but not in this repository. Tests in this repository should be focused on ensuring that the APIs (and their behaviours) don't unexpectedly change.

[...]

PR #699, opened 2021-06-24
Disable integration tests

mtojek:
Fixes: https://github.com/elastic/package-registry/issues/698
Closes: https://github.com/elastic/package-registry/pull/691

This PR disables integration tests in the Package Registry repository. It removes all logic related to it. I have also removed the vendor directory to be consistent with other repositories.


testing/environments/local.yml
```diff
@@ -1,26 +0,0 @@
...
-    depends_on:
-      elasticsearch:
-        condition: service_healthy
-
-  elasticsearch:
-    ports:
-      - "127.0.0.1:9200:9200"
-
-  package-registry:
-    build: ../..
```
jsoriano, line 19: Is there still some test or something that tries to build the registry container?

magefile.go
```diff
@@ -171,23 +149,3 @@ func Clean() error {
...
 	return os.RemoveAll("package-registry")
 }
-
-func Vendor() error {
-	fmt.Println(">> mod - updating vendor directory")
-
-	err := sh.RunV("go", "mod", "tidy")
-	if err != nil {
-		return err
-	}
```
jsoriano, line 181: We probably still want to check that `go mod tidy` doesn't produce changes.

jsoriano (commented):
> I have also removed the vendor directory to be consistent with other repositories.

:+1:

testing/environments/local.yml
mtojek: Currently, it's only in the Jenkinsfile: https://github.com/elastic/package-registry/blob/master/.ci/Jenkinsfile#L164

magefile.go
mtojek: Fixed, thanks

testing/environments/local.yml
jsoriano: I think this is enough, I only wanted to check that there is something checking that the container image can be built.

jsoriano (approved):
:sparkles: I love when code can be removed :slightly_smiling_face:

NOTICE.txt
```diff
@@ -1,5 +0,0 @@
-Elastic package-registry
```
ruflin, line 1: Is this file removed on purpose?

ruflin (commented):
* Can we separate the changes around moving vendor directories and removing integration testing? 
* For the compose files: What is the new recommended way to test a registry change during development?

NOTICE.txt
mtojek: Yes, as there is also LICENSE file: https://github.com/elastic/package-registry/blob/master/LICENSE.txt

ruflin: For example in Beats repo we have both files and I remember there was a reason for this. Not sure if we still need both? @urso you might remember?

mtojek:
> Can we separate the changes around moving vendor directories and removing integration testing?

Ok, I can do this.

> For the compose files: What is the new recommended way to test a registry change during development?

It's the same as it was before, as integration tests were redundant. There are endpoint tests defined here: https://github.com/mtojek/package-registry/blob/disable-integration-tests/main_test.go#L47-L67

NOTICE.txt
mtojek: I see, let's clarify if we need this file. It isn't present in any package related repo (integrations, elastic-package, package-spec, etc.)

mtojek:
I extracted the vendor part, so this PR went down to 11 files. @ruflin what do you think should be included in this PR to get it into the master branch?

mtojek:
> For the testing and compose file, I was referring to manual testing. Lets say I have a PR and I want to check it quickly with an "actual" Kibana setup. Before I used the compose files to build the container and spin it up. As the compose files are not in anymore, does something similar still exist? I personally like to provide devs an easy way to manually test their changes.

I used to start it with "go run ." and then point Kibana to the local EPR instance.

ruflin:
I assume in the above scenario, you run Kibana from source? How do you run it from a released version without having to setup all the stack manually?

mtojek:
I think that it isn't necessary to run against the Kibana from source. You can docker-build the package-registry locally (with packages in the `build` directory) and then docker-tag it accordingly.

Issue #670, opened 2021-01-11
Package registry redesign


ycombinator:
This issue is a placeholder to capture requirements - current pain points + desired enhancements - for a redesigned package registry. A number of discussions have happened on this topic privately; this issue is the place to centralize them as as well as bring more visibility and discoverability to them.

Summarizing from various private discussions:
- The current package registry works quite well but we foresee some potential scalability, performance, and reliability issues as the number of packages continues to grow. 
  - We need to come up with a design that would allow the package registry process to start up in low and constant amount of time.
  - Ideally the size of the package registry image and the memory used by the package registry process would not grow proportionally with the number of packages being served by the registry.
    - Big packages should be supported https://github.com/elastic/package-spec/issues/135.
  - The current package registry bears two distinct responsibilities under the hood: that of providing a search API for packages and that of acting as a static file server for packages. There might be an opportunity to decouple these responsibilities from each other to address the performance and scalability concerns mentioned above, but also to increase the reliability by delegating each responsibility to appropriate technologies.
- There will be a need for private / air-gapped package registries, especially for deployments that are subject to regulatory approval with very strict security/access requirements. https://github.com/elastic/integrations/issues/1178
- The registry will ideally support packages from multiple sources, not just Elastic-developed packages.
- The ability to serve as a source of truth for our marketing integrations page https://www.elastic.co/integrations
  - The ability to render landing pages and docs
  - Web-scale through a CDN if not already

ycombinator:
@mtojek @ruflin @masci @andresrc please feel free to edit the issue description and add more requirements if I've missed any.

Next step is to come up with a proposal doc for a registry redesign that would address all the requirements (and link to it from this issue) so we can all review and iterate on it.

ruflin:
Adding to the list:

* API to publish packages: This partially falls into the multiple sources but having an API would remove the need to support multiple sources
* Search backend: You have it scoped under separate responsibility / split up registry. Lets not make early decision on this but what we need for sure is a search backend as currently we load everything into memory to have quick results. This will not work forever.

ycombinator:
I'm going to start working on a proposal draft for this one. I'll post the link here (Google doc) as soon as I have something to share and review.

In the mean time, if there are more use cases / requirements for the package registry that are not currently satisfiable, please keep adding them here as comments. Thanks!

ycombinator:
As I started working on a proposal for this issue, I realized it's quite a broad scope covering everything from scalability of the package registry to security aspects. 

[...]

PR #702, opened 2021-07-20
Add APM instrumentation

jsoriano:
In https://github.com/elastic/package-registry/issues/670 we are planning to introduce some refactors, in part motivated by performance issues. Introduce some instrumentation to evaluate what parts of the service require more performance tunning. And in the future to be able to compare the impact of the applied changes.

With this instrumentation we can collect traces such as this one:
Captura de pantalla de 2021-07-20 19-36-32

Checklist:
- [x] Fix `/categories`, tests failing.
- [x] Add docs.


util/packages.go
```diff
@@ -25,20 +27,26 @@ type Packages []Package
 // The list is stored in memory and on the second request directly served from memory.
 // This assumes changes to packages only happen on restart (unless development mode is enabled).
 // Caching the packages request many file reads every time this method is called.
-func GetPackages(packagesBasePaths []string) (Packages, error) {
+func GetPackages(ctx context.Context, packagesBasePaths []string) (Packages, error) {
+	span, ctx := apm.StartSpan(ctx, "GetPackages", "custom")
+	defer span.End()
```
axw, line 32: Given that the function just returns cached results if not already loaded, maybe it's not worth creating a span for this? It'll mostly be extremely fast, and when it's not then all the time will be attributed to GetPackagesFromFilesystem

categories.go
```diff
@@ -26,129 +28,163 @@ type Category struct {
...
 
-			// Skip experimental packages if flag is not specified
-			if p.Release == util.ReleaseExperimental && !experimental {
-				continue
-			}
+	return filter, nil
+}
+
+func (filter categoriesFilter) Filter(ctx context.Context, packages util.Packages) map[string]util.Package {
+	span, _ := apm.StartSpan(ctx, "FilterPackages", "custom")
```
axw, line 98: I would suggest changing all the span types from "custom" to "app". This should work better with the "time spent by span type" chart in APM:

image

Transaction time is countered under "app", and if you add spans with type "app" then they will add to this. i.e. it's best practice (probably not written down anywhere) to use "app" for spans related to application code.

axw (approved):
Looks good from the Go Agent usage angle, just a couple of minor comments.

categories.go
mtojek, line 98: nit: I assume it's `error` swallowed. Is it safe here?

mtojek (approved):
LGTM as long as CI is happy. Do we need to modify Dockerfiles in Package Storage too?

util/packages.go
jsoriano: Good point. Currently the packages are cached in memory, and only reloaded on startup, but this is something that can change in the near future. There may be a need to "reload" packages, what could affect this method, or they may be cached in some shared external store.
Said that, we can add the span when this method does more things.

categories.go
jsoriano: Changed :+1: 

Btw, when initializing the service ([here](https://github.com/elastic/package-registry/pull/702/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R88)), I am using "backend.init" as transaction type, is this ok?
Are there some guidelines or list of recommended values for transaction and span types?

jsoriano:
> Do we need to modify Dockerfiles in Package Storage too?

Yes, next step would be to see where we send the data, and configure the registries used in production.

categories.go
axw: That is okay, and I don't think it particularly matters. HTTP requests will be given the transaction type "request". I'm not aware of any guidance for transaction types for this kind of program initialisation.

PR #703, opened 2021-07-30
Serve packages archived (as zip) on storage

jsoriano:
Draft exploring the idea of serving packages stored directly as zip files, so users receive the same packages as stored in the backend.

An "`Indexer"` interface is introduced to allow to provide different implementations for sources of packages.
List of packages is not global anymore, allowing better isolation of test cases. 

This is done in context of #670.

Checklist:
- [ ] Refactor the artifacts handler so it uses the paths from the indexer.
- [ ] Refactor the indexer so it offers a more efficient way to look for specific packages and versions (probably needed for previous point).
- [ ] Some conditional code is added to decide how to handle packages depending on how they are stored, make it more explicit and safer in general.
- [ ] In the filesystem indexer, collect the `Package` objects from the zip files without extracting them to disk.


util/fs.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	switch {
+	case info.IsDir():
+		return NewExtractedPackageFileSystem(path)
+	case strings.HasSuffix(path, ".zip"):
+		return NewZipPackageFileSystem(path)
```
mtojek, line 41: That reminded me: https://github.com/golang/tools/blob/master/godoc/vfs/zipfs/zipfs.go

mtojek (commented):
The idea of indexer seems to be the right direction, but it also shows the spaghetti code we have here. 

I understand that this will change the way we store packages in Package Storage - zip archives instead of raw files? Please keep in mind that we'll have to adjust our tooling to support this distribution form and also inform other parties about the initiative (teams contributing directly to Package Storage like endpoint or APM).

There are also some features we use thank to raw files:
* we can see differences between next package revisions ([PR changes](https://github.com/elastic/package-storage/commit/9fbb2f13fc114c9bdf85e02e640dcb9cfab9cb08))
* "grepping" is easy

I don't think it's impossible to plan replacements for them.

util/fs.go
jsoriano: Oh, I was hoping to find something like this :slightly_smiling_face: With the new interfaces added in 1.16 these things should be more common.

main.go
```diff
@@ -60,6 +60,10 @@ type Config struct {
 	CacheTimeCatchAll   time.Duration `config:"cache_time.catch_all"`
 }
 
+type Indexer interface {
+	GetPackages(context.Context, *util.GetPackagesOptions) (util.Packages, error)
+}
+
```
mtojek, line 66: This is just a loose suggestion, I don't want to interfere or stop you from making the progress.

I guess you can extract all interfaces and domain classes to a separate package, so they can be used to implement more sophisticated Indexers and Storages (more or less similarly to https://github.com/elastic/beats/tree/master/libbeat/beat).

jsoriano: I like to have the interfaces defined where they are used :slightly_smiling_face: Implementations don't really need it to implement it.

If we find that there may be more users of Indexers, we can reconsider and move the interface to a separate package (an alternative can also be to duplicate the definition, what can be fine in many cases too).

jsoriano:
@mtojek I am opening this for review.

I have updated the branch with recent changes related to categories and migrated the new code to follow the refactor, API tests pass now as they are in master.

I have also hidden the new feature behind an experimental flag, so we don't have surprises by trying to serve additional files. Still with this refactor we would have API and storage less coupled, and we would have the package filtering in a single place.

We can continue later with further optimizations or refactors as the one to use `io.FS` implementations, I have added them as postponed tasks in the description.

static.go
```diff
@@ -8,26 +8,56 @@ import (
...
-			log.Printf("stat package path '%s' failed: %v", r.URL.Path, err)
+			badRequest(w, "invalid package version")
+			return
+		}
 
+		fileName, ok := vars["name"]
+		if !ok {
+			badRequest(w, "missing file name")
+			return
+		}
```
mtojek, line 45: nit: this can extracted to separate function for readability purposes

search.go
```diff
@@ -80,179 +70,46 @@ func newSearchFilterFromParams(r *http.Request) (searchFilter, error) {
...
-	p.BasePackage.BasePolicyTemplates = updatedBasePolicyTemplates
-	return p
-}
-
-func getPackageOutput(ctx context.Context, packagesList map[string]map[string]util.Package) ([]byte, error) {
+func getPackageOutput(ctx context.Context, packages util.Packages) ([]byte, error) {
 	span, ctx := apm.StartSpan(ctx, "GetPackageOutput", "app")
 	defer span.End()
 
-	separator := "@"
```
mtojek, line 240: I'm not sure if this logic wasn't placed here due to performance reasons. Maybe double-check it with @ruflin ?

util/fs.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (fs *extractedPackageFileSystem) Close() error { return nil }
+
+// zipPackageFileSystem provides utils to access files in a zipped package.
+type zipPackageFileSystem struct {
+	root   string
+	reader *zip.ReadCloser
+}
+
+func NewZipPackageFileSystem(p *Package) (*zipPackageFileSystem, error) {
```
mtojek, line 65: Things are mixed here: exposed function with private structure.

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
+		return nil
+	}
+	fsBuilder := func(p *Package) (PackageFileSystem, error) {
+		return NewExtractedPackageFileSystem(p)
+	}
+	return &FileSystemIndexer{
+		paths:     paths,
+		label:     "FileSystemIndexer",
+		walkerFn:  walkerFn,
+		fsBuilder: fsBuilder,
```
mtojek, line 101: maybe just:

```
fsBuilder: NewExtractedPackageFileSystem,
```

?

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
+func (i *FileSystemIndexer) getPackagePaths(packagesPath string) ([]string, error) {
 	var foundPaths []string
-	for _, packagesPath := range allPaths {
-		log.Printf("Packages in %s:", packagesPath)
-		err := filepath.Walk(packagesPath, func(path string, info os.FileInfo, err error) error {
-			relativePath, err := filepath.Rel(packagesPath, path)
-			if err != nil {
-				return err
-			}
+	err := filepath.Walk(packagesPath, func(path string, info os.FileInfo, err error) error {
```
mtojek, line 205: You can also consider WalkDir as it might be a bit more efficient for raw file system.

mtojek (commented):
I left some comments, but nothing really breaking the idea. Good job!

I'm not sure if you have done it, but maybe you can try to build a Docker image with real content from package storage to check the feature (including compressed packages).

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
+
+	// Builder to access the files of a package in this indexer.
+	fsBuilder FileSystemBuilder
+}
+
+var walkerIndexFile = errors.New("file should be indexed")
+
+// NewFileSystemIndexer creates a new FileSystemIndexer for the given paths.
+func NewFileSystemIndexer(paths ...string) *FileSystemIndexer {
+	walkerFn := func(basePath, path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
jsoriano: Current implementation doesn't complain if the directories don't exist, only if there are no packages, this behaviour is maintained here. We can reconsider this in a follow up.

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
+		label:     "FileSystemIndexer",
+		walkerFn:  walkerFn,
+		fsBuilder: fsBuilder,
+	}
+}
 
-// GetPackages returns a slice with all existing packages.
+// NewZipFileSystemIndexer creates a new ZipFileSystemIndexer for the given paths.
+func NewZipFileSystemIndexer(paths ...string) *FileSystemIndexer {
+	walkerFn := func(basePath, path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
jsoriano: Added error handling for the walker errors. If a directory doesn't exist, it is ignored, to keep current behaviour.

util/packages.go
```diff
@@ -5,6 +5,7 @@
 package util
```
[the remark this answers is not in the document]
jsoriano: Agree, this `util` package contains too many things, I will leave this for the end, or for a follow up :)

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
+// Options can be used to filter the returned list of packages. When no options are passed
+// or they don't contain any filter, no filtering is done.
 // The list is stored in memory and on the second request directly served from memory.
 // This assumes changes to packages only happen on restart (unless development mode is enabled).
 // Caching the packages request many file reads every time this method is called.
-func GetPackages(ctx context.Context, packagesBasePaths []string) (Packages, error) {
-	if packageList != nil {
-		return packageList, nil
+func (i *FileSystemIndexer) GetPackages(ctx context.Context, opts *GetPackagesOptions) (Packages, error) {
+	if i.packageList == nil {
```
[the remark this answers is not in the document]
jsoriano: Adding an explicit init method for indexers.

util/fs.go
```diff
@@ -0,0 +1,207 @@
...
+	f, err := fs.Open(name)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return ioutil.ReadAll(f)
+}
+
+// virtualPackageFileSystem provide utils for package objects that don't correspond to
```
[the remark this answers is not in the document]
jsoriano: This is used now in tests, it can be used too as base for filesystems that instantiate packages without actually having them on disk.

util/fs.go
```diff
@@ -0,0 +1,207 @@
...
+	Glob(pattern string) (matches []string, err error)
+	Close() error
+}
+
+// extractedPackageFileSystem provides utils to access files in an extracted package.
+type extractedPackageFileSystem struct {
+	path string
+}
+
+func NewExtractedPackageFileSystem(p *Package) (*extractedPackageFileSystem, error) {
```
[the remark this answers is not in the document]
jsoriano: Making type public.

main.go
```diff
@@ -112,14 +115,23 @@ func initServer() *http.Server {
 
 	config := mustLoadConfig()
 	packagesBasePaths := getPackagesBasePaths(config)
-	ensurePackagesAvailable(ctx, packagesBasePaths)
+
+	indexers := []Indexer{
+		util.NewFileSystemIndexer(packagesBasePaths...),
+	}
+	if serveZip {
```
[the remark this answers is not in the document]
jsoriano: I think this will be safe, I only added it last minute as a safety net to don't have it enabled by default in production by now :slightly_smiling_face: Not a strong opinion, so as you prefer :slightly_smiling_face:

util/fs.go
```diff
@@ -0,0 +1,207 @@
...
+type zipFileSeeker struct {
+	fs.File
+
+	reader *zip.ReadCloser
+	path   string
+}
+
+// Seek implements the seeker interface for zip files. This is inefficient, it shouldn't
+// be frequently used.
+func (f *zipFileSeeker) Seek(offset int64, whence int) (n int64, err error) {
```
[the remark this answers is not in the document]
jsoriano: Yes, this is hacky. Full story is that I wanted to use [`http.ServeContent`](https://pkg.go.dev/net/http#ServeContent) because it handles content type and so on, but this requires the file reader to implement `io.ReadSeeker`, an interface that zip file readers don't implement. So I added a minimal implementation for it.

I can give a thought to serve the file with the proper content type without `http.ServeContent`, what wouldn't require this hack, but would require to duplicat some things that this helper does.

static.go
jsoriano: Will add params handling to its own function.

search.go
jsoriano: To what logic do you refer? To the whole search/filtering logic? I am moving it to a common place so all handlers have consistent behaviour, and we have a single place to look at to make performance optimizations if needed :slightly_smiling_face:

util/packages.go
jsoriano: `NewExtractedPackageFileSystem` returns the specific type `*ExtractedPackageFileSystem`, fsBuilder expects a builder that returns the `FileSystemBuilder` interface.

search.go
mtojek: I meant the sorting logic below.

util/packages.go
mtojek: Ok, I see now. My intention was to get rid of the additional proxy function due to lang limitations. If it isn't possible, feel free to ignore this comment.

util/fs.go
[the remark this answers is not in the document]
mtojek: Ok, let's leave it for now. Thank you for the explanation.

util/packages.go
jsoriano: Good idea, done.

search.go
jsoriano: Well, before it was sorting the packages received as values in a map of maps, in that case sorting the packages as is done here would imply to copy the packages first to a list and then copying them again when sorting, or creating a list of pointers to the packages and sorting it. In any of these cases sorting only the keys and then creating the list of packages was probably better.

Now this method receives already a list of pointers to the packages, what can be sorted right away without the need of building the list of keys to sort.

mtojek: > Now this method receives already a list of pointers to the packages, what can be sorted right away without the need of building the list of keys to sort.

I was concerned that it operates on a list of relatively complex package structures. If this is not the thing, then feel free to resolve this comment.

util/packages.go
```diff
@@ -19,84 +20,326 @@ import (
...
-	for _, packagesPath := range allPaths {
-		log.Printf("Packages in %s:", packagesPath)
-		err := filepath.Walk(packagesPath, func(path string, info os.FileInfo, err error) error {
-			relativePath, err := filepath.Rel(packagesPath, path)
-			if err != nil {
-				return err
-			}
+	err := filepath.Walk(packagesPath, func(path string, info os.FileInfo, err error) error {
+		err = i.walkerFn(packagesPath, path, info, err)
+		if err != walkerIndexFile {
```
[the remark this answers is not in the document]
jsoriano: Better, yes :+1:

ruflin:
I really like the idea of introducing the Indexer especially as it also gets rid of all the passing around of paths. It also means assuming at one point we need to have an Indexer as a service (ES?)) it can be added to the list.

I only skimmed through the changes and it contains a lot of refactoring too which is great. As you know, I'm not too big of a fan of large PRs and if possible it would be nice to take out some parts of the refactoring / cleanup into separate PRs as it will make quicker to review and later in case of issues figuring out when it happened. But as I'm not the main reviewer leave it to @mtojek and you to decide if you want to get it all in at once.

PR #704, opened 2021-08-04
Prepare version 0.21.0

jsoriano:


mtojek (commented):
Please check if some unit tests don't require regenerating: `go test . -generate`.

mtojek (approved):

Issue #714, opened 2021-07-26
Incorrect count for Datastore and Network integrations is displayed on Browse tab under Integrations.


vipulsrivastava-qasource:
**Kibana version:** 7.14 BC-4 Kibana cloud environment

**Host OS and Browser version:** All, All

**Preconditions:**

1. 7.14.0 BC-4 Kibana cloud environment should be available.

```
Build Details:
Build: 42656
Commit: 82a4f6a7fa23946667599787cdb1b2d82c1d0dfb
```

**Steps to reproduce:**

1. Login to Kibana.
2. Navigate to Integrations-> Browse tab.
3. Click on Datastore integration under Browse by category.
4. Observe that incorrect count is displayed for datastore integration.
5. Now, again click on Network integration under Browse by category
6. Observe that incorrect count is displayed for Network integration.

**Actual Result**
Incorrect count for Datastore and Network integrations is displayed on Browse tab under Integrations.

**Expected Result**
Correct count for Datastore and Network integrations is displayed on Browse tab under Integrations.

**What's working**

- This issue is not occurring for other integration under Browse by category.

**What's not working**
N/A

**Screen Recording**

- **Datastore Integration**
![Datastore](https://user-images.githubusercontent.com/69579402/126942122-d6592556-e6e4-4cfb-99fa-73c3a11f536f.png)

- **Network Integration**
![Network](https://user-images.githubusercontent.com/69579402/126942135-21bf7a04-c929-402f-91da-baf4be840af2.png)

kaiyan-sheng:
Thanks @jen-huang . Does anyone know why when we click a package tile in different locations, the version changes? 
![integrations-version](https://user-images.githubusercontent.com/14081635/130831134-ef79a469-2158-4668-af43-bd6fd89f012c.gif)

mtojek:
> This would show the parent package in more cases, what may change the UX, and it may break things if there are consumers of /search that expect to have always all the policies included in the packages.

I vote for not changing the API behavior for customers. I guess that most people ignores categories and counters in the left pane and simply focus on the search result. We shouldn't filter out results.

> Other option could be to include the categories in the /search response, and Kibana must do the filtering on the client side.

I would prefer this option if we weren't GA, we could simply remove `/categories` endpoint and just include it as "meta" in the search results.

[...]

PR #716, opened 2021-09-07
Adjust counting logic for categories/policy templates

mtojek:
Issue: https://github.com/elastic/package-registry/issues/714

This PR corrects following issues:

- [x] `/categories` counts policies and packages separatedly as now, but packages are counted too if they don't match but any of their policies does (for the AWS case this would mean that the count for datastore would be 3: the package and the rds and dynamodb policies, it is 2 now).
- [x] `/search` returns packages with their policies filtered out (for AWS it would return a package with just 2 policies, what would match the count of 3).

### DIFF
--- a/util/strings.go
+++ b/util/strings.go
@@ -1,0 +1,15 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package util
+
+// StringsContains function checks if needle is present in the stack.
+func StringsContains(haystack []string, needle string) bool {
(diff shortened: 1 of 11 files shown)


util/strings.go
```diff
@@ -0,0 +1,15 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package util
+
+// StringsContains function checks if needle is present in the stack.
+func StringsContains(haystack []string, needle string) bool {
```
jsoriano, line 8: :smile: love the variable names

jsoriano (approved):
Great, thanks!

PR #719, opened 2021-09-09
Fix: don't list old packages with categories incompatible with latest revisions 

mtojek:
Issue: https://github.com/elastic/package-registry/issues/714

This PR addresses [this](https://github.com/elastic/package-registry/issues/714#issuecomment-914340280) and [this](https://github.com/elastic/package-registry/issues/714#issuecomment-904519559) issue, so that old packages are hidden from search results.

In tests:

`example-0.0.2` has `web` category, but it's hidden from `/search?category=web`, because `example-1.0.0` doesn't have this category. The implementation without `all=true` will focus only on latest packages. This is inline with numbers returned by category endpoint.

Personal comment:

I think that this search/category logic must be refactored as it's too much tangled at the moment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Fix: don't list old packages with categories incompatible with latest revisions. [#719](https://github.com/elastic/package-registry/pull/719)
+
 ### Added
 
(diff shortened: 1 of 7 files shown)


jsoriano:
> I think that this search/category logic must be refactored as it's too much tangled at the moment.

+1, I should continue with https://github.com/elastic/package-registry/pull/703 :sweat_smile: where the search code is refactored an unified so all endpoints use the same.

jsoriano (approved):

PR #732, opened 2021-09-22
Fix the documentation for kibana.version query param

andrewkroh:
The docs said the name of the parameter was `kibana`, but the code uses `kibana.version`.
So this fixes the docs.

https://github.com/elastic/package-registry/blob/95673bb86e422c64ef2db16675b90660c28a0f27/search.go#L71

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -72,7 +72,7 @@ paths:
         - schema:
             type: string
           in: query
-          name: kibana
+          name: kibana.version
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -72,7 +72,7 @@ paths:
         - schema:
             type: string
           in: query
-          name: kibana
+          name: kibana.version
```
mtojek, line 75: Ok, so we have proof that nobody uses this API with OpenAPI :)

mtojek (approved):

ruflin (approved):

PR #733, opened 2021-09-22
De-dot 'kibana.version' in /package JSON responses

### LINKED ISSUE #730: JSON responses contain key with dot in 'kibana.version'
The response from the `/package` endpoint uses a key name that contains a dot. It's not common to use field names that contain dots and this might lead to confusion when processing the response data. For example the `elastic-package` tool would not [unmarshal this field](https://github.com/elastic/elastic-package/blob/a0a7defd0db574d36924b52bf6fe5dce899d9f74/internal/packages/packages.go#L78-L81) if it were returned.

Example: https://epr.elastic.co/package/gcp/1.1.0/

```
  "conditions": {
    "kibana.version": "^7.15.0"
  },
```

My preferred representation would be

```
  "conditions": {
    "kibana": {
      "version": "^7.15.0"
    }
  },
```

It should be easy to fix the package-registry, but changing it could break clients. So are there any clients using this field? I can check Kibana, but is that the only user?

andrewkroh:
This changes the response JSON from

```
  "conditions": {
    "kibana.version": "^7.15.0"
  },
```

to

```
  "conditions": {
    "kibana": {
      "version": "^7.15.0"
    }
  },
```

Closes #730

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* Change format of responses to `/package` to make `{"constraint": {"kibana.version": "7.16.0"}}` be `{"constraint": {"kibana": {"version": "7.16.0"}}}`. [#733](https://github.com/elastic/package-registry/pull/733)
+
 ### Bugfixes
 
 ### Added
--- a/testdata/generated/package.json
+++ b/testdata/generated/package.json
@@ -15,7 +15,9 @@
     "azure"
   ],
   "conditions": {
-    "kibana.version": "~7.x.x"
+    "kibana": {
+      "version": "~7.x.x"
+    }
   },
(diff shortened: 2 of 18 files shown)


mtojek (commented):
I'm not sure if it doesn't mean that we break the API or just ... unwrap keys :) @ruflin WDYT ?

ruflin:
This might break Fleet. @joshdover @jen-huang Lets make sure the team reviews / tests this before it gets in.

jen-huang (approved):
I tested this locally and did not see any breaking changes in Fleet, in fact we already had the de-dotted version typed:
https://github.com/elastic/kibana/blob/3eb93a84f52e1c884de06ce089fcd6f1dbbf23fa/x-pack/plugins/fleet/common/types/models/package_spec.ts#L54

I also tested the search endpoint that Fleet uses to query for packages which are compatible, i.e. `/search?experimental=true&kibana.version=7.14.1` and that worked as expected too.

PR #734, opened 2021-09-22
Add 'constraints' and 'owner' to /search responses

### LINKED ISSUE #731: [Enhancement] Include 'constraints' and 'owner' in /search responses
I'm interested in having the `/search` endpoint return the `constraints` (Kibana requirements) and `owner` (maintaining Github team). This would be useful information to have when filtering on the search results.

For example with `elastic-package status`, which uses `/search` to collect information on the status of packages in the registry, this data could be included in the output (or used to filter). It might be used to determine what packages are exclusively available in 7.N, but are not in `production`.

The fields would need moved to `BasePackage` (or a separate `SearchResponse` type could be created). https://github.com/elastic/package-registry/blob/89b3dd2ee3d4f3a5cf67cbe035e17737b83a3103/util/package.go#L73-L74

andrewkroh:
This adds the constraints.kibana.version and owner fields to /search responses.

It also removes HTML escaping from the JSON responses to make them more readable.

Closes #731


mtojek (approved):
I guess that this PR is safe to be added to the mainline.

ruflin (approved):
LGTM. I like the new json marshalling.

@andrewkroh Any chance you could share some background on where we will make use of the fields in the search output?

Update: Found the answer to my question here: https://github.com/elastic/package-registry/issues/731

mtojek:
Let merge this as I have the release on my today TODO list.

Issue #722, opened 2021-09-14
Add category-names to packages in search response / Add package-ids to category list in categories response


thomasneirynck:
EPR service does not expose an API to determine which categories a package belongs to, or vice versa, which packages belong to a list of categories.

e.g.  `https://epr-snapshot.elastic.co/search` is missing a `categories`-property for each package
```
[
  {
      "name":"activemq",
      "title":"ActiveMQ",
      "version":"0.0.1",
      "release":"beta",
      "description":"ActiveMQ Integration",
      "type":"integration",
      "download":"/epr/activemq/activemq-0.0.1.zip",
      "path":"/package/activemq/0.0.1",
      "icons":[
         {
            "src":"/img/activemq.svg",
            "path":"/package/activemq/0.0.1/img/activemq.svg",
            "title":"activemq",
            "size":"32x32",
            "type":"image/svg+xml"
         }
      ],
      "policy_templates":[
         {
            "name":"activemq",
            "title":"ActiveMQ logs and metrics",
            "description":"Collect logs and metrics from ActiveMQ instances"
         }
      ]
   },
   {
      "name":"auditd",
      "title":"Auditd",
      "version":"1.1.3",
      "release":"ga",
      "description":"This Elastic integration collects and parses logs from the Audit daemon (auditd)",
      "type":"integration",
      "download":"/epr/auditd/auditd-1.1.3.zip",
      "path":"/package/auditd/1.1.3",
      "icons":[
         {
            "src":"/img/linux.svg",
            "path":"/package/auditd/1.1.3/img/linux.svg",
            "title":"linux",
            "size":"299x354",
            "type":"image/svg+xml"
         }
      ],
      "policy_templates":[
         {
            "name":"auditd",
            "title":"Auditd logs",
            "description":"Collect logs from Auditd instances"
         }
      ]
   },
...
]
```

e.g. `https://epr-snapshot.elastic.co/categories` is missing package-ids for each category
```
[
   {
      "id":"aws",
      "title":"AWS",
      "count":1
   },
   {
      "id":"azure",
      "title":"Azure",
      "count":3
   },
   {
      "id":"cloud",
      "title":"Cloud",
      "count":2
   },
   {
      "id":"config_management",
      "title":"Config management",
      "count":1
   },
...
]
```

[...]

PR #735, opened 2021-09-28
[Fleet] Add categories to search response output

thomasneirynck:
Closes https://github.com/elastic/package-registry/issues/722

This adds the categories to the search-response output.
-  Inclusion of categories is required so clients can build a facetted browser, merging multiple sources into overlapping category taxonomy. (this is the goal of https://github.com/elastic/kibana/issues/93084)
-  Adding this will also avoid having to run repeated requests to EPR. Now Kibana needs to issue a search each time a category is selected. It also needs to run a single-search up-front for all packages just so it can provide a total count. This PR will remove the need for this redundant network traffic.

cc @joshdover

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add categories to /search output. Categories are added to the package and policy-templates. [#722](https://github.com/elastic/package-registry/issues/722)
+
 ### Deprecated
 
 ### Known Issues
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Add categories to /search output. Categories are added to the package and policy-templates. [#722](https://github.com/elastic/package-registry/issues/722)
+* Add categories to /search output. Categories are added to the package and policy-templates. [#735](https://github.com/elastic/package-registry/pull/735)
 
 ### Deprecated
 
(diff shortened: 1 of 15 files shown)


util/package.go
```diff
@@ -83,14 +83,16 @@ type BasePackage struct {
 	Icons               []Image              `config:"icons,omitempty" json:"icons,omitempty" yaml:"icons,omitempty"`
 	Internal            bool                 `config:"internal,omitempty" json:"internal,omitempty" yaml:"internal,omitempty"`
 	BasePolicyTemplates []BasePolicyTemplate `json:"policy_templates,omitempty"`
+	Categories          []string             `config:"categories" json:"categories"`
```
mtojek, line 86: Please include yaml annotation and "omitempty" for consistency.

mtojek (changes requested):
Thanks for the PR. Let's make sure if it works correctly for all our use cases. 

1. Please run ~`make format`~`mage format` to clean this code.
2. Please add a changelog entry to the `CHANGELOG.md`.
3. Please run `go test .` to see if something has changed (`-generate` to regenerate golden files).

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add categories to /search output. Categories are added to the package and policy-templates. [#722](https://github.com/elastic/package-registry/issues/722)
```
mtojek, line 15: nit: that should be link to the pull request (see other bullets)

mtojek (approved):
LGTM!

I double-checked it against the policy template case (url: `/search?category=datastore&experimental=true`) and it seems to be correct: https://github.com/elastic/package-registry/blob/d2bbfb6efbca8c93d0c9195c871d40450817fae0/testdata/generated/search-category-datastore.json

I think you're good to go, Thomas. You can proceed with [releasing procedure](https://github.com/elastic/package-registry#release). If you encounter any blockers, we can help you with them. FYI release cycles of EPR are detached from Elastic release trains, so feel free to do this anytime.

PR #742, opened 2021-10-07
Fix issue with relative paths when loading data streams

jsoriano:
Relative paths calculation for data streams depends on the paths
having trailing separators or not. Use `filepath.Rel` to obtain these
relative paths.

### DIFF
--- a/packages/fs.go
+++ b/packages/fs.go
@@ -51,7 +51,7 @@ func (fs *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, er
 	for i := range matches {
 		match, err := filepath.Rel(fs.path, matches[i])
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to obtain path under package root path (%s): %w", fs.path, err)
 		}
 		matches[i] = match
 	}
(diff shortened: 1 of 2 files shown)


packages/fs.go
```diff
@@ -49,7 +49,11 @@ func (fs *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, er
 		return
 	}
 	for i := range matches {
-		matches[i] = matches[i][len(fs.path+string(filepath.Separator)):]
+		match, err := filepath.Rel(fs.path, matches[i])
+		if err != nil {
+			return nil, err
```
mtojek, line 54: nit: not sure what's the general rule here, but maybe it's worth wrapping it with some context.

packages/package.go
```diff
@@ -511,7 +511,11 @@ func (p *Package) GetDataStreamPaths() ([]string, error) {
 	}
 
 	for i, _ := range paths {
-		paths[i] = paths[i][len(dataStreamBasePath)+1:]
+		relPath, err := filepath.Rel(dataStreamBasePath, paths[i])
+		if err != nil {
+			return nil, err
```
mtojek, line 516: nit: same here (what's the relPath)

mtojek (approved):

Issue #711, opened 2021-08-17
Add TLS support


jsoriano:
Currently it is only possible to run the registry with TLS in combination with a proxy.

Add support for TLS in the registry itself with the usual settings to support custom certificates. 

Support custom certificates also in the docker image.

PR #745, opened 2021-10-12
Add config from environment and config path flag

jsoriano:
Thinking about adding the TLS flags for #711, I found that it could be handy, specially for the containerized registry, to allow to include these settings using environment variables.

So for example to add certificates to a containerized registry, one needs to mount the certificates as volumes, and add the environment variables, what is a common practice. Using flags only, the command line needs to be modified, what can override other defaults as the default address or `-disable-package-validation`, [included in current distribution images](https://github.com/elastic/package-storage/blob/989ded50e8455777e2556dfd3d72735b284d5751/Dockerfile#L16).

This change includes:
* Some documentation notes about how the registry can be configured.
* A new flag `-config` to select the configuration file.
* Flags can be provided using environment variables. Flags have precedence over environment variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Configuration file path can be selected with the `-config` flag.
+* Configuration flags can be provided using environment variables.
+
 ### Deprecated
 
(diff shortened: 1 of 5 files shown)


mtojek (approved):
I like this idea, but let's document in the repository's README.

mtojek (approved):

PR #746, opened 2021-10-12
Listen with TLS if configured

jsoriano:
Add flags to enable listening on TLS.

_Pending testing and documentation, waiting for #745_


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Configuration file path can be selected with the `-config` flag. [#745](https://github.com/elastic/package-registry/pull/745)
 * Configuration flags can be provided using environment variables. [#745](https://github.com/elastic/package-registry/pull/745)
+* Add -tls-cert and -tls-key flags to configure HTTPS. [#711](https://github.com/elastic/package-registry/issues/711) [#746](https://github.com/elastic/package-registry/issues/746)
```
mtojek, line 17: nit: code tags (`-tls-cert`, `-tls-key`)

main.go
```diff
@@ -130,6 +135,13 @@ func initServer() *http.Server {
 	return &http.Server{Addr: address, Handler: router}
 }
 
+func runServer(server *http.Server) error {
+	if tlsCertFile != "" && tlsKeyFile != "" {
```
mtojek, line 139: nit: I'm wondering whether we need to introduce some basic validation here (to check if the file exist or similar). Otherwise we may end up with a generic HTTP server error.

PR #756, opened 2021-10-15
Handle Last-Modified/If-Modified-Since headers

### LINKED ISSUE #755: If-Modified-Since headers are ignored
Since the refactor in handlers in 1.4.0 (#703), the static handler is ignoring the `If-Modified-Since` header.

Before this version, a 304 status code was returned when static files were requested with this header, since the refactor the file is always returned with a 200 status code, independently of the cache headers.

jsoriano:
Handle headers used to indicate if the server should send the required resource based on its modification time.

Add tests to ensure that this continues to be supported.

Fixes #755.

### DIFF
--- a/packages/http.go
+++ b/packages/http.go
@@ -64,7 +64,7 @@ func ServeFile(w http.ResponseWriter, r *http.Request, p *Package, name string) 
 		return
 	}
 	if err != nil {
-		log.Printf("failed to check file modification time (%s) in package: %v", name, err)
+		log.Printf("stat failed for %s: %v", name, err)
 		http.Error(w, "internal server error", http.StatusInternalServerError)
 		return
 	}


packages/http.go
```diff
@@ -59,17 +58,24 @@ func ServeFile(w http.ResponseWriter, r *http.Request, p *Package, name string)
...
+
+	f, err := fs.Open(name)
 	if err != nil {
 		log.Printf("failed to open file (%s) in package: %v", name, err)
 		http.Error(w, "internal server error", http.StatusInternalServerError)
 		return
 	}
 	defer f.Close()
 
-	http.ServeContent(w, r, name, time.Time{}, f)
```
mtojek, line 74: :)

packages/http.go
```diff
@@ -59,17 +58,24 @@ func ServeFile(w http.ResponseWriter, r *http.Request, p *Package, name string)
...
 	}
 
-	f, err := fs.Open(name)
+	stat, err := fs.Stat(name)
 	if os.IsNotExist(err) {
 		http.Error(w, "resource not found", http.StatusNotFound)
 		return
 	}
+	if err != nil {
+		log.Printf("failed to check file modification time (%s) in package: %v", name, err)
```
mtojek, line 67: This log line might be misleading. I would just say that `stat failed: <err>`.

packages/http.go
jsoriano: No idea what I was thinking about :sweat_smile:

packages/http.go
jsoriano: Thanks, changed, keeping the requested path in the error message.

PR #757, opened 2021-10-18
Prepare 1.5.1 release

jsoriano:


jsoriano:
> I think you may need to bump up the app version in the .go file.

For patch versions it is already set to the next one: https://github.com/elastic/package-registry/blob/26fa143a5c5e0f37a7d1cea33164281c9806896c/main.go#L33

I will open another PR after this one with the change in the changelog and in the version for the next one.

PR #765, opened 2021-11-08
Remove internal parameter from packages and search queries

jsoriano:
The `internal` flag is not used in any package currently served by the production registry, and it is not part of the spec.

It was used for the `base` package during development phases, but this package is not needed in final versions, and is not served by the registry since some months ago.

Removing support for this flag in the registry.


mtojek (approved):
LGTM

Maybe consider also sending an email to the developers' group. I know that this property isn't used, but it's more about dropping the idea of internal packages.

andrewkroh (approved):

PR #766, opened 2021-11-09
Search multiple kibana versions

jsoriano:
Search and category endpoints accept multiple `kibana.version`, and results are combined.


jsoriano:
Closing as this is something that we don't need by now.

PR #784, opened 2021-12-17
[docs] Fix parameter in README example

simitt:
Fix parameter in example to `kibana.version`.


mtojek (approved):
I hope it didn't cause much trouble for you and THAT is the way this bug was discovered :) Thank you for fixing docs, Silvia!

PR #785, opened 2022-01-13
Add prerelease search parameter and deprecate experimental

jsoriano:
Implements support for prerelease semver tags in the package registry, in the context of https://github.com/elastic/package-spec/issues/225.

After this change, the package registry will return versions of packages in development only if `prerelease=true` or `experimental=true` parameters are included in the search requests. This shouldn't be breaking in current versions of Kibana because Kibana is currently always including `experimental=true`. A follow-up task will be created in Kibana to remove the uses of `experimental`, and add support for optionally enabling `prerelease` (https://github.com/elastic/kibana/issues/122973).

Summarizing changes here:
* In-development versions of packages are not returned by default on search queries. A version is considered an in-development prerelease if it includes a prerelease tag, or if its major is 0.
* `prerelease` parameter is added to include versions of packages in development in search requests.
* `experimental` parameter is deprecated.
* For compatibility with current versions of Kibana, when `experimental` is set to true, `prerelease` is also set to true.
* Packages without a release tag and a stable version according to semver are considered GA for legacy API purposes, or beta if they are prerelease versions. Datastreams without a release tag have the same release level as its parent package.


categories.go
```diff
@@ -84,11 +84,25 @@ func newCategoriesFilterFromQuery(query url.Values) (*packages.Filter, error) {
...
 
+	// Deprecated: release tags to be removed.
 	if v := query.Get("experimental"); v != "" {
 		filter.Experimental, err = strconv.ParseBool(v)
 		if err != nil {
 			return nil, fmt.Errorf("invalid 'experimental' query param: '%s'", v)
 		}
+
+		// For compatibility with older versions of Kibana.
+		if filter.Experimental {
```
mtojek, line 95: Would you mind challenging this change a bit?

It looks like we're renaming `experimental=true` to `prerelease=true`. Meaning of both parameters is not so far from each other, so maybe we should deprecate it? It would not affect the API in terms of legacy support.

packages/package_test.go
```diff
@@ -250,6 +250,27 @@ func TestNewPackageFromPath(t *testing.T) {
...
+func TestIsPrerelease(t *testing.T) {
+	cases := []struct {
+		version    string
+		prerelease bool
+	}{
+		{"0.1.0-rc1", true},
+		{"0.1.0", true}, // Major version 0 shouldn't be considered stable.
+		{"1.0.0-beta1", true},
+		{"1.0.0-rc.1", true},
+		{"1.0.0-SNAPSHOT", true},
```
mtojek, line 262: By looking at this set of use-case, I'm wondering if we should introduce a flag/property to spec for Elastic internal-only packages. Before we could drop such packages in the "snapshot" stage and even remove them if we decide not to experiment, it might be impossible in the setup. If the end-user enables "prerelease mode", they will see ALL unstable packages, even ones planned for a release.

Let me know what do you think. I don't mean that it should be part of this package.

CHANGELOG.md
```diff
@@ -9,20 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking changes
 
 * Ignore the `internal` parameter in packages and `/search` requests. [#765](https://github.com/elastic/package-registry/pull/765)
+* Packages with major version 0 or with prerelease labels are only returned by search requests when they include `prerelease=true` or `experimental=true`. [#785](https://github.com/elastic/package-registry/pull/785)
```
mtojek, line 12: Out of curiosity, do you think we should perform an exercise and check how many packages are affected because of this change?

mtojek (commented):
The PR looks good to me, but I left a few points which may lead to a longer discussion. Let me know what do you think.

categories.go
jsoriano: Do you mean to remove `Experimental` from the filtering rules and use only `Prerelease`? I considered that, but decided to explicitly leave it because their implementations differ. I can reconsider it in any case, I think the resulting responses will be the same.

It will probably change the behaviour when no `experimental` flag is set, but this behaviour is also changed now.

(Btw, I am deprecating `experimental`, but it could still be used)

packages/package_test.go
jsoriano: Umm, not sure to understand, could you give an example?

We already had an internal flag, but we removed it as it was not being used anywhere, nor seemed required https://github.com/elastic/package-registry/pull/765.

> If the end-user enables "prerelease mode", they will see ALL unstable packages, even ones planned for a release.

I think it is ok if users can see any kind of prerelease _published_ packages. If a developer team doesn't want a package to be available to users in any case because it is in hard development or something like this, I think they shouldn't publish it in a public registry. They can locally build and install it for development and testing purpouses.

categories.go
mtojek: Hm... what about staying with `experimental` instead of introducing a new flag? We could slightly extend its meaning to include more packages.

CHANGELOG.md
jsoriano: Packages with prerelease tags: 1, apm-server 8.0.0-rc1
Packages with major version 0: [48](https://gist.github.com/jsoriano/e0750af2d3f25d037cb3fa2a87ae0b1d), they are all marked as beta or experimental, except a couple of them. We will have to think what to do about them. The couple of ga ones should be probably versioned as 1.x.

Btw, I have also found that kibana, logstash and microsoft packages are marked as experimental, but have 1.x versions, we will have to check this too.

I can delay the merge of this till we clarify the status of these packages.

categories.go
jsoriano: Not a strong opinion, but in the future people will wonder why this flag is called `experimental` :slightly_smiling_face: 

Also, even when they have similar purpouses, they have slightly different behaviours. I prefer to include a new flag that future client implementations can use, and leave the old one as compatibility layer for old ones, that we can in some years directly remove when/if not used.

If the problem is about adding prerelease packages to the experimental response, and given that the only reason to do this is to maintain current behaviour for current implementations of kibana, that always use `experimental=true`. We can consider different alternatives:
* Version the API, so we add some prefix to endpoints, and we introduce the new behaviour only in the ones under `/v2`. This adds complexity, but we may end up needing something like this at some point in any case.
* Base the behaviour on the user agent. Future versions of Kibana are going to include their versions in the user agent (https://github.com/elastic/kibana/pull/121297), we could give different responses to different versions of Kibana. Though this is hacky and error prone, proxies may alter these headers.

mtojek: Got it 👍  Thanks for the discussion, let's try to preserve the KISS rule. I suppose that introducing another flag is simpler than preparing `/v2` endpoint.

mtojek (approved):
Implementation LGTM, let's just clarify affected packages.

packages/package_test.go
jsoriano: > It might be a problem if we want to show the world a package on a specific day due to a marketing campaign. By release, I mean publishing it to the registry.

This could still be done, by publishing the package on the specific day.

> In another use case, let's say, there is a team that experiments with packages and the registry. They publish a "hello-world" or PoC. The package will become visible for Kibana instances with "prerelease" enabled, even though the team wouldn't like to make it public.
> 
> I'm afraid that we'll end up with many PoC packages in the prerelease phase, usually ones that were parked in the snapshot stage.

Perhaps we still need the snapshot registry, or a "playground" one for these cases of hello-world learning packages. Maybe at some point we can have some kind of package registry SaaS, so developer teams can start their own registries, where they can publish whatever they want. Similar to what [PPAs](https://help.launchpad.net/Packaging/PPA) provide for debian repositories... 

I see these more as questions about what to do regarding the multiple registries, what I considered separatedly in https://github.com/elastic/package-spec/issues/225, as something that could be eventually done, independently of removing the release tag or not.

Also take into account that these cases are also not possible now with current implementation if we had a single registry, Kibana searches packages with `experimental=true`, so all published packages in a registry are always visible when Kibana is configured to use this registry. Even if we made Kibana to optionally use `experimental=true`, nothing prevents a developer publishing hello-worlds with `release: ga`.

mtojek: > Also take into account that these cases are also not possible now with current implementation if we had a single registry, Kibana searches packages with experimental=true, so all published packages in a registry are always visible when Kibana is configured to use this registry.

Yeah, but it's possible to mitigate it by using different stages (for example, snapshot).

Anyway, we can cut the discussion here if we think that's not the problem anytime soon. With new storage, we could think about adopting the concept of namespaces or accounts (similar to Dockerhub), and limit package visibility this way (for example, a private namespace for Elastic cloud-related packages).

jsoriano:
Thinking about the (not few) packages with version 0.x, I am wondering if we would need three modes of operations for search:

* No prerelease package is returned (as with this PR with `prerelease=false`).
* All prerelease packages are returned (as with this PR with `prerelease=true`).
* A third mode, that for packages with stable versions only returns the stable versions (as with `prerelease=false`), but it also returns packages that only have prerelease versions. This mode would help to give visibility to integrations that are in development, as the ones with 0.x versions only, while keeping a stable channel for packages that have GA versions.

@mtojek @akshay-saraswat wdyt about this third mode?

mtojek:
The third option might be hard to explain to users why some 0.x packages are considered GAs. It would be for keeping the rule as simple as possible `0.x = prerelease`. We don't have many of them, so we can release the missing ones as `1.0`. I'm happy to discuss it though!

BTW having packages marked as 0.2.0 or 0.6.0 and not maintaining them gives also a bad experience :) Users may consider them as unstable to use and discouraged to use the platform.

Maybe @ruflin can see some other benefits/drawbacks of this change that will help us take the best decision.

jsoriano:
> The third option might be hard to explain to users why some 0.x packages are considered GAs.

Well, they should still appear as non-GA somehow in the UIs, but yes, this mode would be a bit confusing, as prereleases would be available for some packages but not for others.

I would also prefer to keep rules simple for users, so they can clearly decide if they want prerelease packages or not in their deployments.

Probably the best is to keep pushing for GA versions of maintained packages, and keep this option simple.

jsoriano:
> It would be good to sync up on this with the Fleet team before you roll out the change.

Yep, I created https://github.com/elastic/kibana/issues/122973 for the changes in Fleet.

> A pure technical thought: Every query param is a filter. That means by default ALL packages should be returned and from there the user can filter down.

This is not exactly true with the current implementation. By default all non-experimental packages are returned, `experimental=true` is needed to include all packages (and `all=true` for all versions). I have followed this principle when implementing `prerelease`. By default the safer set of packages and versions are returned, and flags can be used to enable additional sets of packages.

Looking it from a pure data perspective, it makes more sense to return everything by default and filter down with parameters, but from the kind of operational data exposed here, I think it makes sense to have defaults that return more safe or relevant results, and use parameters to extend them in particular use cases.

ruflin:
> This is not exactly true with the current implementation.

I had exactly the same "challenge" when I implemented it the first time. 

I can see both approaches work. The reason I think of the "filter down" part is there might become a point where the search index is served by Elasticsearch. In this scenario it would mean "adding"  param to the request means under the hood the query is likely adjusted to remove a filter.

But as said before, both approaches will work, go with your preference.

testdata/generated/categories-prerelease-error.json
```diff
@@ -0,0 +1 @@
+invalid 'prerelease' query param: 'foo'
```
mtojek, line 1: nit: not a JSON file, right?

mtojek (approved):
LGTM, I assume you will take the decision when to push this PR :)

testdata/generated/categories-prerelease-error.json
jsoriano: Non-json files renamed in https://github.com/elastic/package-registry/pull/785/commits/b43b38aa2d6ed6bad66825b0a88d5dd967ced1ec.

andrewkroh:
@jsoriano Can you please update the README to reflect these changes. It's missing `prerelease` param and should now show `experimental` as deprecated.

jsoriano:
> @jsoriano Can you please update the README to reflect these changes. It's missing `prerelease` param and should now show `experimental` as deprecated.

Good one, I'll do it, thanks!

PR #787, opened 2022-01-20
Use Ubuntu LTS as base image instead of CentOS

jsoriano:
Move to Ubuntu LTS as other Elastic projects have done after the discontinuation of CentOS fixed releases in favour of CentOS Stream.

Image size goes after this change from 242MB to 143MB.


jlind23 (approved):

mtojek (commented):
Post-merge LGTM :)

Issue #761, opened 2021-10-27
Generate OpenAPI definitions automatically from Go code


jsoriano:
There is an openAPI definition available in the repository, but there is nothing checking that it is maintained up to date with the actual implementation.

Consider strategies to maintain this definition updated.

mtojek:
QQ: Do we know if anyone uses the OpenAPI posted in the [package-registry](https://github.com/elastic/package-registry/blob/main/openapi.yml)? If we know that it's outdated/incorrect/not usable, maybe it's better to remove it and:

1. Leave another issue to support any kind of generation.
2. Forget about it.

jsoriano:
+1 to remove it by now if not used (or not usable).

I would keep an open issue about providing an OpenAPI definition or something similar. I think this would be a nice to have, and can be also helpful for us to generate documentation.

jsoriano:
FTR, it seems there was a discussion about using registry's open API in Kibana/Fleet, but it was discarded: https://github.com/elastic/kibana/issues/70711

PR #794, opened 2022-02-23
Update README and OpenAPI to include prerelease parameter

jsoriano:
Update README and OpenAPI definition accordingly to the changes in #785.

OpenAPI definition is still most likely outdated in any case, there is another open issue about this (https://github.com/elastic/package-registry/issues/761).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,16 +19,18 @@ The `/search` API endpoint has few additional query parameters. More might be ad
   By default this endpoint always returns only the newest compatible package.
 * `category`: Filters the package by the given category. Available categories can be seend when going to `/categories` endpoint.
 * `package`: Filters by a specific package name, for example `mysql`. Returns the most recent version.
-* `all`: This can be set to true to list all package versions. This is set to `false` by default.
-* `experimental`: This can be set to true to list packages considered to be experimental. This is set to `false` by default.
+* `all`: This can be set to `true` to list all package versions. This is set to `false` by default.
+* `prerelease`: This can be set to `true` to list prerelease versions of packages. This is set to `false` by default.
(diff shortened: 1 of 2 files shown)


mtojek (approved):

README.md
```diff
@@ -19,16 +19,18 @@ The `/search` API endpoint has few additional query parameters. More might be ad
   By default this endpoint always returns only the newest compatible package.
 * `category`: Filters the package by the given category. Available categories can be seend when going to `/categories` endpoint.
 * `package`: Filters by a specific package name, for example `mysql`. Returns the most recent version.
-* `all`: This can be set to true to list all package versions. This is set to `false` by default.
-* `experimental`: This can be set to true to list packages considered to be experimental. This is set to `false` by default.
+* `all`: This can be set to `true` to list all package versions. This is set to `false` by default.
+* `prerelease`: This can be set to `true` to list prerelease versions of packages. This is set to `false` by default.
```
andrewkroh, line 23: Could you define what "prerelease versions" means. Like say that it mean versions <1.0 and packages using the deprecated `beta` and `experimental` release tags (IIRC).

jsoriano: Added description. The registry fully relies on semantic versioning to decide if the version of a package is a prerelease, release tags are not considered. This is why it is important to have consistency between release tags and versions till we fully sunset versions that rely on release tags.

mtojek (approved):
nit: we may consider introducing any logic to validate the `openapi.yml`. I'm not sure if there is any project using it now.

jsoriano:
> 

Yep, we have this issue about validating the definition https://github.com/elastic/package-registry/issues/761

PR #795, opened 2022-03-01
Use Common Log Format for logging output

jsoriano:
Use [Combined Log Format](http://httpd.apache.org/docs/2.2/logs.html#combined), so package-registry logs can be parsed with the
same tools that can be used to parse logs of common web servers such as
Apache or nginx.

This log format is not much more verbose and includes additional information:
* Timezone in the timestamp.
* HTTP method, HTTP version, status code and response size.
* Referer (when available).
* User agent (when available).

This changes the log format from:
```
2022/03/01 10:12:31 source.ip: 127.0.0.1:42334, url.original: /
2022/03/01 10:12:38 source.ip: 127.0.0.1:42342, url.original: /search
```
To:
```
127.0.0.1 - - [01/Mar/2022:10:28:28 +0100] "GET / HTTP/1.1" 200 71 "" "curl/7.58.0"
127.0.0.1 - - [01/Mar/2022:10:28:57 +0100] "GET /search HTTP/1.1" 200 7912 "" "curl/7.58.0"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Use Combined Log Format [#795](https://github.com/elastic/package-registry/pull/785).
+
 ### Deprecated
 
 ### Known Issues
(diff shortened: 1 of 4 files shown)


jsoriano:
Discussed offline with @ruflin, lets move to JSON ECS logs so no ingest pipeline is needed for ingestion in ES.

PR #796, opened 2022-03-01
Add structured logging

jsoriano:
Add structured logging using JSON ECS format. This includes much more information in logs for http requests. As it uses JSON ECS, it doesn't need special parsing.

`-log-level` can be used to select log level.

Logger is a global singleton in the `util` package. Not ideal, but was not trivial to pass the logger to all the components. We can iterate on this though.

Why am I using `zap` for structured logging? It is one of the two Go loggers currently supported by Elastic for [ECS logging](https://www.elastic.co/guide/en/ecs-logging/overview/current/intro.html). The other one is Logrus. I chose `zap` because it is supposed to be faster, but I don't have a strong opinion.

Example lines now:
```
{"log.level":"info","@timestamp":"2022-03-02T20:46:29.747+0100","log.logger":"http","message":"GET /search HTTP/1.1","source.address":"127.0.0.1","http.request.method":"GET","http.response.code":200,"http.response.body.bytes":7912,"source.ip":"127.0.0.1","user_agent.original":"curl/7.58.0","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-03-02T20:46:37.282+0100","log.logger":"http","message":"GET / HTTP/1.1","source.address":"127.0.0.1","http.request.method":"GET","http.response.code":200,"http.response.body.bytes":71,"source.ip":"127.0.0.1","user_agent.original":"curl/7.58.0","ecs.version":"1.6.0"}
```
With `-log-type dev` a "developer mode" can be enabled, that logs like this:
```
2022-03-02T20:30:19.402+0100	INFO	http	GET / HTTP/1.1	{"source.address": "127.0.0.1", "http.request.method": "GET", "http.response.code": 200, "http.response.body.bytes": 71, "source.ip": "127.0.0.1", "user_agent.original": "curl/7.58.0"}
2022-03-02T20:30:21.323+0100	INFO	http	GET /search HTTP/1.1	{"source.address": "127.0.0.1", "http.request.method": "GET", "http.response.code": 200, "http.response.body.bytes": 7912, "source.ip": "127.0.0.1", "user_agent.original": "curl/7.58.0"}
```

Follow up of #795.


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Structured logging following JSON ECS format [#796](https://github.com/elastic/package-registry/pull/786).
```
ruflin, line 15: Changing the log structure shloud be considered a breaking change.

util/logging.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+
+			LogRequest(logger, next, w, r)
+		})
+	}
+}
+
+// LogRequest captures information from a handler handling a request, and generates logs
+// using this information.
+func LogRequest(logger *zap.Logger, handler http.Handler, w http.ResponseWriter, req *http.Request) {
```
mtojek, line 99: nit: does it need to be exposed?

util/logging.go
```diff
@@ -0,0 +1,132 @@
...
+
+var logLevel = zap.LevelFlag("log-level", zap.InfoLevel, "log level (default \"info\")")
+var logType = flag.String("log-type", defaultLoggerType, "log type (ecs, dev)")
+
+var logger *zap.Logger
+var loggerMutex sync.Mutex
+
+// InitLogger initializes the logger, this is ignored afer a logger has
+// been created for this process.
+func InitLogger(loggerType string) {
```
mtojek, line 37: If we know that there are two types of loggers and no more, then we can replace them with `UseDevLogger` and `UseProductionLogger`.

mtojek (approved):
LGTM

Is there any other benefit coming from this PR apart from cleanup and a nice console logger for development purposes? I mean, do we log more data or include error context?

jsoriano:
> Is there any other benefit coming from this PR apart from cleanup and a nice console logger for development purposes? I mean, do we log more data or include error context?

For most of requests it is the same data, plus timestamps with timezone and caller file and line. This is one of the lines printed on startup when reading the packages:
```
{"log.level":"info","@timestamp":"2022-03-03T11:02:35.070+0100","log.origin":{"file.name":"packages/packages.go","file.line":208},"message":"found package","package.name":"foo","package.version":"1.0.0","package.path":"testdata/package/foo/1.0.0","ecs.version":"1.6.0"}
```

For requests we log more data:

Before:
```
2022/03/01 10:12:38 source.ip: 127.0.0.1:42342, url.original: /search
```
After:
```
{"log.level":"info","@timestamp":"2022-03-03T11:02:38.059+0100","log.logger":"http","message":"GET /search HTTP/1.1","source.address":"127.0.0.1","http.request.method":"GET","http.response.code":200,"http.response.body.bytes":7912,"source.ip":"127.0.0.1","user_agent.original":"curl/7.58.0","ecs.version":"1.6.0"}
```

We also have log levels now.

util/logging.go
jsoriano: Not really, changed. I considered to expose it because it is generic enough to be used with any handler and a zap logger. But no need to expose it by now.

mtojek (approved):
> Btw, I wonder if we should log "access" and "error" logs to different streams, as is usual in web servers. We could log "access" logs to stdout, and everything else to stderr. @mtojek wdyt?

I'm fine with such a change if the `docker logs` command still works so that we will miss neither access nor error logs.

jsoriano:
@mtojek I found that I was not capturing valuable fields from the url, as the path and the query. I have added this and a couple of fields more. And a test for the captured fields. Could you please take another look?

jsoriano:
> > Btw, I wonder if we should log "access" and "error" logs to different streams, as is usual in web servers. We could log "access" logs to stdout, and everything else to stderr. @mtojek wdyt?
> 
> I'm fine with such a change if the `docker logs` command still works so that we will miss neither access nor error logs.

Umm, yes, let's leave this for a separate change, it can have unexpected implications.

util/logging.go
```diff
@@ -110,36 +111,66 @@ func LoggingMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
...
+		if intPort, err := strconv.Atoi(port); err == nil && intPort != 0 {
+			fields = append(fields, zap.Int("url.port", intPort))
+		}
+	}
 	if user := req.URL.User; user != nil {
 		if username := user.Username(); username != "" {
-			fields = append(fields, zap.String("user.name", username))
+			fields = append(fields, zap.String("url.username", username))
 		}
 	}
```
mtojek, line 172: Hm, we don't expect any users here, right?

mtojek (approved):

util/logging.go
jsoriano: Correct, let's remove this.

mtojek (approved):

PR #798, opened 2022-03-04
ci: move to fleet-ci

v1v:
## What is the problem this PR solves?

Use `fleet-ci`  to scale horizontally the CI.

NOTE: this PR will fail in `beats-ci` 

## Why

In order to distribute the load and reduce a single point of failure,

## Actions

- [ ] Agree when to merge
- [ ] Create webhook to point to `fleet-ci`

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -1,10 +1,10 @@
-  
+
 ---
 
 ##### GLOBAL METADATA
 
 - meta:
-    cluster: beats-ci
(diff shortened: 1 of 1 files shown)


mtojek (approved):
LGTM as long as it doesn't break anything and we have required CI permissions.

v1v:
> LGTM as long as it doesn't break anything and we have required CI permissions.

Nothing changes in terms of permissions, but the URL is different, both CI controllers are similar in terms of configuration and permissions

PR #809, opened 2022-05-05
Support input packages

mtojek:
Meta-issue: https://github.com/elastic/package-spec/issues/319

Blocker: https://github.com/elastic/package-spec/pull/328

This PR modifies the Package Registry to support "input packages".

Changes:
1. Use `sql_input` as test package.
2. Add missing fields to "package index" API method.
3. Filter by "type" with /search API method.
4. Regenerate expected test results.


testdata/generated/search-all.json
```diff
@@ -354,6 +354,37 @@
       "web"
     ]
   },
+  {
+    "name": "sql_input",
+    "title": "SQL Input",
+    "version": "1.0.1",
+    "release": "ga",
+    "description": "Execute custom queries against an SQL database and store the results in Elasticsearch.",
+    "type": "input",
```
jsoriano, line 363: What are old versions of Kibana do with packages that are not integrations?

Should we filter by `type: integration` when no `type` is passed? and maybe have a `type: all` to list all packages.

Or maybe this is a point where we should start versioning the API?

main_test.go
```diff
@@ -86,6 +86,7 @@ func TestEndpoints(t *testing.T) {
 		{"/search?prerelease=true", "/search", "search-package-prerelease.json", searchHandler(indexer, testCacheTime)},
 		{"/search?prerelease=foo", "/search", "search-package-prerelease-error.txt", searchHandler(indexer, testCacheTime)},
 		{"/search?category=datastore&prerelease=true", "/search", "search-category-datastore-prerelease.json", searchHandler(indexer, testCacheTime)},
+		{"/search?type=input", "/search", "search-input-packages.json", searchHandler(indexer, testCacheTime)},
```
jsoriano, line 89: Maybe we should add test cases for:
* An input package with multiple versions.
* A package that has a type on one version, and another type on another version.

testdata/generated/search-all.json
mtojek: > and maybe have a type: all to list all packages.

I wouldn't like to design an API that won't be used by Fleet at all (KISS).

> Should we filter by type: integration when no type is passed? 

That would be counter-intuitive, you want to search for packages, but for some historical reason, the /search filters some of them. I would only go with that option to prevent any problems on the Fleet side.

> Or maybe this is a point where we should start versioning the API?

It's always a good practice to version it, but not sure if this is a "dead end" to introduce versioning or if we just enable another GET parameter.

jsoriano: Well, then going back to the first question, what are current/old versions of Kibana do with packages that are not integrations?

mtojek: Passing the question to @joshdover.

main_test.go
mtojek: > An input package with multiple versions.

I added one more revision of `sql_input`.

> A package that has a type on one version, and another type on another version.

`integration_input` package, older revision is integration, newer is input.

I haven't noticed anything unusual, but feel free to review API responses in case I missed something.

jsoriano (commented):
Looks good, waiting for the discussion about backwards compatibility with Kibana/Fleet.

main_test.go
mtojek: Good idea, fixed

mtojek:
As the input-type spec has been [released](https://github.com/elastic/package-spec/pull/328), I suppose that this is the next task to be reviewed/merged.

testdata/generated/search-all.json
joshdover: This seems like something that could easily be tested manually without code-level knowledge of Fleet. That would be how I test this personally.

jlind23 (approved):

jlind23 (commented):
LGTM

testdata/generated/search-all.json
mtojek: Ok, I started the 7.16.3 stack and was able to list the `sql_input` package (`type: input`), which is not good. On the other hand if we use Kibana constraint to cut old Kibana versions that should do the trick.

Zrzut ekranu 2022-05-26 o 13 46 20

WDYT @joshdover? I can go any direction, either adjust API or adjust Kibana constraints.

jsoriano: > On the other hand if we use Kibana constraint to cut old Kibana versions that should do the trick.

Agree that this can be a good solution. We should probably add a check for this in the package-spec, and we would need to wait to know the version of Kibana that is going to support input packages for this.

mtojek: Sounds like a good idea, but, to be honest, I wouldn't like to block this PR until that happens.

jsoriano: Ah yeah, I didn't want to mean that :slightly_smiling_face: But this may delay declaring input package as GA in the spec.

jsoriano (approved):

testdata/generated/search-all.json
jlind23: @mtojek I would say that each new input package should start with a kibana version constraints higher or equals to 8.3 which should then do the trick isn't it?

mtojek: Yes, Julien, that's the idea. Just waiting for the confirmation/approval from @joshdover.

joshdover: Kibana version constraint enforced by package-spec SGTM, agreed it can be done as a follow up.

PR #812, opened 2022-05-16
ci: publish docker images

v1v:
### What

Tag the production docker image with the Unified Release version.

### Why

The docker images will be available as soon as the tag release is created in https://github.com/elastic/fleet-server

https://github.com/elastic/fleet-server/tags

Therefore, versions will be aligned with the releases.

### Issues

Notifies https://github.com/elastic/fleet-server/pull/1471

### DIFF
--- a/.ci/deploy-package-registry.groovy
+++ b/.ci/deploy-package-registry.groovy
@@ -1,0 +1,49 @@
+#!/usr/bin/env groovy
+
+@Library('apm@current') _
+
+pipeline {
+  agent { label 'ubuntu-20' }
+  environment {
+    BASE_DIR = "src/github.com/elastic/package-registry"
+    PIPELINE_LOG_LEVEL = 'INFO'
+    DOCKER_REGISTRY = 'docker.elastic.co'
+    DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
+    DOCKER_TAG = "${params.DOCKER_TAG}"
+    DOCKER_IMG_SOURCE = "${env.DOCKER_REGISTRY}/package-registry/distribution:production"
+    DOCKER_IMG_TARGET = "${env.DOCKER_REGISTRY}/package-registry/distribution:${env.DOCKER_TAG}"
+  }
+  options {
+    timeout(time: 1, unit: 'HOURS')
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20', daysToKeepStr: '30'))
+    timestamps()
+    ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
+    quietPeriod(10)
+  }
+  parameters {
+    string(name: 'DOCKER_TAG', defaultValue: 'latest', description: 'The docker tag to be published.')
+  }
+  stages {
+    stage('Publish Docker image'){
+      options { skipDefaultCheckout() }
+      steps {
+        dockerLogin(secret: "${env.DOCKER_REGISTRY_SECRET}", registry: "${env.DOCKER_REGISTRY}")
+        retryWithSleep(retries: 3, seconds: 5, backoff: true) {
+          sh(label: 'Docker pull', script: 'docker pull ${DOCKER_IMG_SOURCE}')
+        }
+        sh(label: 'Docker retag',  script: 'docker tag ${DOCKER_IMG_SOURCE} ${DOCKER_IMG_TARGET}')
+        retryWithSleep(retries: 3, seconds: 5, backoff: true) {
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):

.ci/deploy-package-registry.groovy
```diff
@@ -0,0 +1,49 @@
...
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20', daysToKeepStr: '30'))
+    timestamps()
+    ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
+    quietPeriod(10)
+  }
+  parameters {
+    string(name: 'DOCKER_TAG', defaultValue: 'latest', description: 'The docker tag to be published.')
```
mtojek, line 27: Maybe we should define an allowed list of values? Otherwise, we may end up with typos or incorrect tags.

v1v: This pipeline is intended to be consumed by the automation system. The automation system will be the one running the unified release process for staging and the git tag creation for the releases. 

Nevertheless, if someone runs this pipeline manually then there is a potential case of creating tags. Though, I'm not sure if this will be a case but a potential miss-use.

Otherwise, what values do you think should be verified? a regex with the major.minor.patch format?

kuisathaverat (approved):

.ci/deploy-package-registry.groovy
mtojek: Yes, I was thinking about the semver regexp.

v1v: Sure, let me add the regex validation. OTOH, there is one potential case by overriding an existing old release, for instance: `8.1.0`, it could be an accident if this job gets triggered, the automation is reliable since it's the one driving the official releases.

mtojek: > Sure, let me add the regex validation. OTOH, there is one potential case by overriding an existing old release

But the risk is present regardless of the presence of regexp, correct?

cachedout: Drive-by comment: in addition to providing guardrails against doing the wrong thing, this kind of validation also serves as a future reminder about the acceptable input, which may make maintenance easier. (Obviously, this can be accomplished via in-line documentation as well.)

v1v: > But the risk is present regardless of the presence of regexp, correct?

Yes

.ci/deploy-package-registry.groovy
```diff
@@ -24,9 +24,21 @@ pipeline {
     quietPeriod(10)
   }
   parameters {
-    string(name: 'DOCKER_TAG', defaultValue: 'latest', description: 'The docker tag to be published.')
+    string(name: 'DOCKER_TAG', defaultValue: 'latest', description: 'The docker tag to be published (format: major.minor.patch).')
```
mtojek, line 27: `major.minor.patch-prerelease`

.ci/deploy-package-registry.groovy
```diff
@@ -47,3 +59,8 @@ pipeline {
     }
   }
 }
+
+def isMajorMinorPatch(String version) {
+  def match = version =~ /\d+.\d+.\d+/
```
mtojek, line 64: same here

.ci/deploy-package-registry.groovy
v1v: what's `pre-release`? an example please

v1v: nevermind:

> * `qualifiedVersion` == '7.0.0-alpha1'

.ci/deploy-package-registry.groovy
```diff
@@ -0,0 +1,66 @@
...
+    quietPeriod(10)
+  }
+  parameters {
+    string(name: 'DOCKER_TAG', defaultValue: 'latest', description: 'The docker tag to be published (format: major.minor.patch(-prerelease)?).')
+  }
+  stages {
+    stage('Validate docker tag'){
+      options { skipDefaultCheckout() }
+      steps {
+        // Validate only semver are allowd for the docker tag.
```
jsoriano, line 33: Nit.
```suggestion
        // Validate only semver are allowed for the docker tag.
```

jsoriano (approved):
Thanks!

.ci/jobs/deploy-package-registry-pipeline.yml
```diff
@@ -0,0 +1,24 @@
+---
+- job:
+    name: deploy-package-registry
```
jsoriano, line 3: Nit. Nothing is actually deployed here, but released, or tagged.

v1v: Yep, what name do you think it could be the best one?

v1v: `release-package-registry-distribution`

Issue #767, opened 2021-11-16
Detach Package Registry from Package Storage


jlind23:
And give ability to pull packages from the Package Storage directly without the need of baking special distributions

While refactoring the Package Registry, please consider applying any Go project layout (original issue: https://github.com/elastic/package-registry/issues/475).

PR #813, opened 2022-05-18
Populate fake storage for testing purposes

mtojek:
Issue: https://github.com/elastic/package-registry/issues/767

This PR introduces fake storage populated with a snapshot of the search index from Package Storage v2.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Prepare stub for Storage Indexer. Disable fetching packages from Package Storage v1. [#811](https://github.com/elastic/package-registry/pull/811)
+* Populate fake storage for testing purposes. [#813](https://github.com/elastic/package-registry/pull/813)
```
jsoriano, line 16: Umm, I am wondering now if these changes need to be mentioned in the changelog, they are not user facing :thinking:

storage/fakestorage.go
```diff
@@ -0,0 +1,81 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
jsoriano, line 1: Rename this file to `something_test.go`, or move it to a test sub-package to be sure that we don't mix test-only things with production code?

storage/index.go
```diff
@@ -0,0 +1,105 @@
...
+			Type         string     `json:"type,omitempty"`
+			Description  string     `json:"description,omitempty"`
+			InputGroup   string     `json:"input_group,omitempty"`
+			TemplatePath string     `json:"template_path,omitempty"`
+			Vars         []variable `json:"vars,omitempty"`
+		} `json:"input,omitempty"`
+		Screenshots []image    `json:"screenshots,omitempty"`
+		Vars        []variable `json:"vars,omitempty"`
+	} `json:"policy_templates,omitempty"`
+}
```
jsoriano, line 57: Oh no, another place where we have to keep these structs :sweat_smile:  https://github.com/elastic/package-spec/issues/322

storage/fakestorage_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"cloud.google.com/go/storage"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPrepareFakeServer(t *testing.T) {
+	// given
+	indexFile := "testdata/search-index-all-1.json"
```
jsoriano, line 20: Could we add a comment here about how to regenerate this file? I guess this has been downloaded from an actual registry?

storage/index.go
```diff
@@ -0,0 +1,105 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package storage
```
jsoriano, line 5: Maybe `storage` is too generic? Could we rename this to `gcs` or something google-cloud specific?

go.mod
```diff
@@ -3,15 +3,17 @@ module github.com/elastic/package-registry
 go 1.18
 
 require (
+	cloud.google.com/go/storage v1.22.0
 	github.com/Masterminds/semver/v3 v3.1.0
 	github.com/elastic/go-licenser v0.3.1
 	github.com/elastic/go-ucfg v0.8.4-0.20200415140258-1232bd4774a6
 	github.com/felixge/httpsnoop v1.0.2
+	github.com/fsouza/fake-gcs-server v1.37.12
```
jsoriano, line 11: I am wondering if we need a full gcs implementation. Files required by the registry are available in public buckets, right? Could we just use `httptest` serving static files for testing? Is there any file only available in private buckets?

jsoriano: Well, the index itself is private, right? Could we read it with authenticated HTTP requests so we don't need a GCP client in the registry?

storage/index.go
mtojek: I was also thinking about `storage/v2`, as `v2` is strictly bound with GCP buckets. WDYT?

storage/fakestorage.go
mtojek: Fixed

CHANGELOG.md
mtojek: Removed.

storage/fakestorage_test.go
mtojek: Added

go.mod
jsoriano: > If we want to go with authenticated HTTP requests, then we need a GCP signer code in the Package Registry, right?

> Sample of URL signing in Go: https://gist.github.com/poindextrose/afa4d4bb5e93c4c3de77. It means that we will still need a dependency on `google.golang.org/cloud/storage`.

Umm, are signed urls the only option to download private files?

It seems to be possible to use normal urls and a bearer token (seen [here](https://cloud.google.com/storage/docs/authentication) and [here](https://stackoverflow.com/questions/64781844/google-cloud-storage-downloading-a-file-using-curl-request)). We would only need a dependency to get this token, but then everything else could be normal http.

storage/index.go
jsoriano: I would like to see something generated from the spec, like https://github.com/elastic/package-spec/issues/322.

go.mod
mtojek: Correct me if I'm wrong, but if this is OAuth, then it may happen that the token expires while the Package Registry is running and we will have to reauthorize again?

storage/index.go
mtojek: It is slightly different from the spec as it includes extra properties like assets and data streams hence I was thinking about a dedicated repo for the model.

jsoriano: Ah, yes, you are right, this is not the same.

PR #815, opened 2022-05-30
Add marshaler logic to Packages

mtojek:
Follow-up on suggestion in the Package Storage Infra repo.

This PR implements (un-)marshaler interface for `packages.Packages`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Prepare stub for Storage Indexer. Disable fetching packages from Package Storage v1. [#811](https://github.com/elastic/package-registry/pull/811)
-* Add marshaler/unmarshaler logic to Packages. [#815](https://github.com/elastic/package-registry/pull/815)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Prepare stub for Storage Indexer. Disable fetching packages from Package Storage v1. [#811](https://github.com/elastic/package-registry/pull/811)
+* Add marshaler/unmarshaler logic to Packages. [#815](https://github.com/elastic/package-registry/pull/815)
 
 ### Deprecated
 
--- a/packages/marshaler.go
+++ b/packages/marshaler.go
@@ -8,10 +8,23 @@ import (
 	"encoding/json"
 )
 
+type MarshallerOption func(packages *Packages) error
+
 func MarshalJSON(pkgs *Packages) ([]byte, error) {
 	return json.MarshalIndent(pkgs, " ", " ")
 }
 
-func UnmarshalJSON(content []byte, pkgs *Packages) error {
-	return json.Unmarshal(content, pkgs)
+func UnmarshalJSON(content []byte, pkgs *Packages, options ...MarshallerOption) error {
+	err := json.Unmarshal(content, pkgs)
+	if err != nil {
+		return err
+	}
+
+	for _, opt := range options {
+		err = opt(pkgs)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/packages/marshaler.go
+++ b/packages/marshaler.go
@@ -1,0 +1,17 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package packages
+
+import (
+	"encoding/json"
+)
+
+func MarshalJSON(pkgs *Packages) ([]byte, error) {
+	return json.MarshalIndent(pkgs, " ", " ")
+}
+
+func UnmarshalJSON(content []byte, pkgs *Packages) error {
+	return json.Unmarshal(content, pkgs)
+}
--- a/packages/marshaler.go
+++ b/packages/marshaler.go
@@ -6,6 +6,9 @@ package packages
 
 import (
 	"encoding/json"
+	"github.com/pkg/errors"
+	"os"
+	"path/filepath"
 )
 
 type MarshallerOption func(packages *Packages) error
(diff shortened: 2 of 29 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Prepare stub for Storage Indexer. Disable fetching packages from Package Storage v1. [#811](https://github.com/elastic/package-registry/pull/811)
+* Add marshaler/unmarshaler logic to Packages. [#815](https://github.com/elastic/package-registry/pull/815)
```
jsoriano, line 16: No user-facing feature, no need to add changelog entry.

packages/marshaler.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/pkg/errors"
+)
+
+type MarshallerOption func(packages *Packages) error
+
+func MarshalJSON(pkgs *Packages) ([]byte, error) {
+	return json.MarshalIndent(pkgs, " ", " ")
+}
+
+func UnmarshalJSON(content []byte, pkgs *Packages, options ...MarshallerOption) error {
```
jsoriano, line 21: Could `Marshal` and `Unmarshal` have been implemented as methods of the `Packages` type, so it implements `Un`/`marshaller`?

Well, I guess we need it this way because of the options?

packages/marshaler_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	expectedFile, err := os.ReadFile(testFile)
+	require.NoError(t, err)
+
+	var packages Packages
+
+	// when
+	err = UnmarshalJSON(expectedFile, &packages,
+		ResolveBasePaths(packagesBasePaths...),
+		UseFsBuilder(ExtractedFileSystemBuilder))
```
jsoriano, line 50: Is a FS builder needed here?

Could we have a test unmarshaling a set of packages read from a search index?

I could imagine something like this:
 ```suggestion
	err = UnmarshalJSON(indexSearchFile, &packages, URLPaths(eprBaseURL))
```

Something like `URLPaths` would be only needed if we set the paths in the packages to the URL where the resource is available in the remote storage.

This can perfectly fit in a follow up, but just in case you expect any problem by now.

File: packages/packages.go [hunk 4]
```diff
@@ -92,21 +92,22 @@ func NewFileSystemIndexer(paths ...string) *FileSystemIndexer {
...
 	}
-	fsBuilder := func(p *Package) (PackageFileSystem, error) {
-		return NewExtractedPackageFileSystem(p)
-	}
 	return &FileSystemIndexer{
 		paths:     paths,
 		label:     "FileSystemIndexer",
 		walkerFn:  walkerFn,
-		fsBuilder: fsBuilder,
+		fsBuilder: ExtractedFileSystemBuilder,
```
jsoriano, line 102: :+1:

packages/marshaler.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, opt := range options {
+		err = opt(pkgs)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ResolveBasePaths(packagesPath ...string) MarshallerOption {
```
jsoriano, line 43: I find this a bit hacky, and not sure if we are going to need this. For cases where the package is available locally, we use the current local indexers. For the packages in the remote index, we unmarshal the index JSON, and we don't have the packages locally.

I would remove this if not needed.

packages/marshaler.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+			if !manifestFound {
+				return errors.Errorf("manifest file is missing (package: %s, version: %s)", (*packages)[i].Name, (*packages)[i].Version)
+			}
+		}
+		return nil
+	}
+}
+
+func UseFsBuilder(builder FileSystemBuilder) MarshallerOption {
```
jsoriano, line 68: Also, not sure if this option would be required here for cases where we don't have the package available locally.

jsoriano (commented):
Added some comments about the needing of the included options.
To index local packages I think we can continue using `NewPackage`, as current indexers do.
When loading the JSON of an index in the storage, the registry doesn't have access to the files locally, so it should work without a FS builder, and without needing to update references to local paths.

packages/marshaler.go
mtojek: If we decide to drop all options, then yes, but it would be hackish too. Here is [an article](http://choly.ca/post/go-json-marshalling/) about the approach.

jsoriano: Sounds good, we can keep it this way.

CHANGELOG.md
mtojek: Removed.

packages/marshaler.go
mtojek: Removed all options.

packages/marshaler.go
mtojek: Removed all options.

testdata/generated/package/example/0.0.2/index.json
```diff
@@ -9,7 +9,7 @@
   "path": "/package/example/0.0.2",
   "conditions": {
     "kibana": {
-      "version": ">=6.0.0"
+      "version": "\u003e=6.0.0"
```
mtojek, line 12: Not sure why that kind of change happened.

packages/marshaler_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	var packages Packages
+
+	// when
+	err = json.Unmarshal(expectedFile, &packages)
+
+	// then
+	require.NoError(t, err, "packages should be loaded")
+	for i := range indexer.packageList {
+		require.Equal(t, packages[i].Name, indexer.packageList[i].Name)
```
jsoriano, line 54: Nit. Are packages sorted in both indexes? Or having the same package in the same index is circumstantial?

testdata/generated/package/example/0.0.2/index.json
mtojek: Fixed

jsoriano (approved):
:+1:

packages/marshaler_test.go
mtojek: Yes, they are sorted. The order shouldn't be impacted.

Issue #816, opened 2022-05-31
fs.Glob skips directories while using .zip file system


mtojek:
While I was working on using the Package Registry's model in the Package Storage Infra, I found a bug in handling `fs.Glob`. `GetDataStreamPaths` [checks if there are any data stream directories](https://github.com/elastic/package-registry/blob/9053aa31e3d310fcb57e5541be5b4047c7826782/packages/package.go#L547) in the .zip archive, but it may happen that there are no intermediate directories and the .zip is 100% correct.

```bash
$ unzip -l jobs/indexing/testdata/1password-0.1.1.zip
Archive:  jobs/indexing/testdata/1password-0.1.1.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
      351  03-04-2022 14:43   1password-0.1.1/changelog.yml
     1344  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/agent/stream/httpjson.yml.hbs
     3560  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/elasticsearch/ingest_pipeline/default.yml
      592  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/fields/base-fields.yml
     3605  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/fields/ecs.yml
     1085  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/fields/fields.yml
     1619  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/manifest.yml
     2242  03-04-2022 14:43   1password-0.1.1/data_stream/item_usages/sample_event.json
     1348  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
     4081  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/elasticsearch/ingest_pipeline/default.yml
      596  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/fields/base-fields.yml
     4865  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/fields/ecs.yml
     1282  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/fields/fields.yml
     1628  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/manifest.yml
     2291  03-04-2022 14:43   1password-0.1.1/data_stream/signin_attempts/sample_event.json
[...]

PR #817, opened 2022-05-31
Support zip files without directory entries

jsoriano:
Some filesystems, as some zip files, can lack of entries for directories.
When looking for files, don't rely on directories.

Fixes https://github.com/elastic/package-registry/issues/816.


mtojek (approved):
Thanks!

jsoriano:
> Does it need a changelog entry as it's a bugfix?

Good point, changelog added.

PR #821, opened 2022-06-07
Update image dependencies

mtojek:
This PR bumps up dependencies:
* Ubuntu version 20.04 -> 22.04
* Go version 1.18.0 -> 1.18.3


Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN go build .
 
 
 # Run binary
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
jsoriano, line 15: We use to include version updates in the changelog. Consider also mentioning changes in the Go runtime that may affect the registry, if any.

PR #823, opened 2022-06-09
Implement remote resolver for storage indexer

### LINKED ISSUE #767: Detach Package Registry from Package Storage

mtojek:
This PR implements hijack handlers for the storage indexer to redirect package zip, artifacts, static requests to https://package-storage.elastic.co/.

It uses HTTP 303 to pass the request. It was the easiest way to implement because I don't have to write content streaming at all. No need for Accept-Range tests, download tests, etc.

Breaking changes:
* Package Registry with storage indexer passes all requests to the Package Storage, it means that some resources won't be available to download except for statics (docs, imgs, etc.)

Spotted bugs:
* Package Storage v2 doesn't return correct content-types: [SVG served via legacy EPR](https://epr.elastic.co/package/aws/1.16.4/img/logo_ebs.svg), [SVG from Package Storage v2](https://package-storage.elastic.co/artifacts/static/aws-1.16.4/img/logo_ebs.svg)

Local testing:

```
go run . -feature-storage-indexer \
  -storage-indexer-bucket-internal gs://mtojek-temp/package-storage-v2 \
  -log-level debug
```

How to prepare the development bucket (`cursor.json` and `search-index-all.json`):

```
$ gsutil ls -r gs://mtojek-temp/package-storage-v2/v2
gs://mtojek-temp/package-storage-v2/v2/:
gs://mtojek-temp/package-storage-v2/v2/

gs://mtojek-temp/package-storage-v2/v2/metadata/:
gs://mtojek-temp/package-storage-v2/v2/metadata/
gs://mtojek-temp/package-storage-v2/v2/metadata/cursor.json

gs://mtojek-temp/package-storage-v2/v2/metadata/1/:
gs://mtojek-temp/package-storage-v2/v2/metadata/1/
gs://mtojek-temp/package-storage-v2/v2/metadata/1/search-index-all.json
```

Questions:
* Should we move indices to the public bucket and just keep the `queue-publishing` with package candidates in the private bucket? I don't remember why have we decided to put indices in the internal one...

Fixes #767


mtojek:
Status update:

I tried to use Google Storage client to reach out the public bucket, but failed as `http.ServeContent` needs `objectHandle` to implement `io.ReadSeekCloser`. I found the [community implementation](https://github.com/googleapis/google-cloud-go/issues/1124#issuecomment-1100755501), but I can't guarantee that it's stable, but worked. I decided to revert.

I tried with a generic `http.Client`, but failed to implement streaming as it's mandatory to serve `Range` requests. I wasn't able to figure out how to implement it nicely. Reverted.

Now, I'm working on enabling the L7 proxy mode when the response can be streamed directly from Package Storage. I don't know what I do with streaming resources from the zip file. _In the worst case, we will have to unpack and store them in the public bucket._

artifacts.go
```diff
@@ -58,7 +60,13 @@ func artifactsHandler(indexer Indexer, cacheTime time.Duration) func(w http.Resp
 			return
 		}
 
+		aPackage := packageList[0]
+		if storageIndexer, assert := aPackage.Indexer().(*storage.Indexer); assert {
+			storageIndexer.HijackArtifactsHandler(w, r, aPackage)
```
jsoriano, line 65: I wouldn't call it "Hijack", here we are doing a legit redirect :slightly_smiling_face:

Some suggestions:
```suggestion
			storageIndexer.RedirectArtifactsHandler(w, r, aPackage)
```
```suggestion
			storageIndexer.RemoteArtifactsHandler(w, r, aPackage)
```

artifacts.go
```diff
@@ -58,7 +60,13 @@ func artifactsHandler(indexer Indexer, cacheTime time.Duration) func(w http.Resp
...
+			return
+		}
```
jsoriano, line 67: Instead of relying on the type of the indexer, maybe we could add some value in the package that helps knowing if the resource is local or remote. Each indexer would be in charge of setting this value accordingly.
With this maybe less refactor is needed, and other future indexers could have redirects too.

```suggestion
		if p := packageList[0]; p.Remote; {
			packages.ServeRedirect(w, r, p)
			return
		}
```
Or it could be a pointer to a resolver that implements the `Hijack*Handler` methods and can be used to resolve the redirect.
```suggestion
		if p := packageList[0]; p.RemoteResolver != nil; {
			packages.ServeRedirect(w, r, p)
			return
		}
```
Perhaps this logic could also be in `packages.ServePackage`, so the handlers remain the same.

jsoriano (commented):
Added a suggestion about explicitly configuring the package to indicate that it requires redirects, instead of relying on the indexer, let me know what you think. I think this may require less refactors.

artifacts.go
mtojek: I can apply those renamings, not a problem :)

I found an analogy with [http.Hijacker](https://pkg.go.dev/net/http#Hijacker), which lets a handler take over the TCP connection. In this case, we let Storage Indexer handlers process requests, which include a redirection.

artifacts.go
mtojek: `RemoteResolver` would be a public field, which I wouldn't expose if it isn't necessary. It might be confusing for other projects using it. Today it's just package-storage-infra, but maybe in the future, we will use it also in elastic-package.

I'm not sure whether depending on Local/Remote boolean property will be flexible enough for future purposes. If we decide to support multiple Remote indexers, we will have to introduce another field.

WDYT?

jsoriano: > `RemoteResolver` would be a public field, which I wouldn't expose if it isn't necessary. It might be confusing for other projects using it. Today it's just package-storage-infra, but maybe in the future, we will use it also in elastic-package.

Well, with current implementation the indexer is being exposed with the new `Indexer()` method, what also includes the `Hijack` methods, apart of other things.

My proposal would only expose what is needed in a package to decide if it has to be served locally or as a redirect, and if it is with a redirect, how to do it. This is information that could make sense to have in the Package struct. This would also avoid the need of having indexer-specific logic in handlers.

If preferred, we could make `RemoteResolver` a method instead of a property.

packages/http.go
```diff
@@ -15,11 +15,11 @@ import (
 	"github.com/elastic/package-registry/util"
 )
 
-func ServePackage(w http.ResponseWriter, r *http.Request, p *Package) {
+// ServeLocalPackage is used by artifactsHandler to serve packages and signatures.
+func ServeLocalPackage(w http.ResponseWriter, r *http.Request, p *Package, packagePath string) {
 	span, _ := apm.StartSpan(r.Context(), "ServePackage", "app")
 	defer span.End()
 
-	packagePath := p.BasePath
```
jsoriano, line 22: Couldn't we continue using `p.BasePath` as the `packagePath`? Having to pass the package and its path can be error prone.

artifacts.go
```diff
@@ -58,7 +58,13 @@ func artifactsHandler(indexer Indexer, cacheTime time.Duration) func(w http.Resp
...
 		}
 
+		aPackage := packageList[0]
+		if aPackage.RemoteResolver() != nil {
+			aPackage.RemoteResolver().RedirectArtifactsHandler(w, r, aPackage)
+			return
+		}
+
 		cacheHeaders(w, cacheTime)
-		packages.ServePackage(w, r, packageList[0])
```
jsoriano, line 62: Nit. `ServePackage` could include the logic to serve local or remote packages, same thing with `ServeFile` and `ServeSignature`.
This way it is easier to remember that we need to take this into account if we add a new handler that uses these helpers, or a new `ServeSomething` helper.

packages/http.go
```diff
@@ -39,18 +38,27 @@ func ServePackage(w http.ResponseWriter, r *http.Request, p *Package) {
...
+
+	stream, err := os.Open(packagePath)
+	if err != nil {
+		logger.Error("failed to open file", zap.Error(err))
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+	defer stream.Close()
+
+	http.ServeContent(w, r, packagePath, f.ModTime(), stream)
```
jsoriano, line 53: Do we still need these changes now that we are doing redirects?

packages/resolver.go
```diff
@@ -0,0 +1,13 @@
...
+
+package packages
+
+import "net/http"
+
+type RemoteResolver interface {
+	RedirectArtifactsHandler(w http.ResponseWriter, r *http.Request, p *Package)
+	RedirectStaticHandler(w http.ResponseWriter, r *http.Request, p *Package, resourcePath string)
+	RedirectSignaturesHandler(w http.ResponseWriter, r *http.Request, p *Package)
+}
```
jsoriano, line 13: Nit. Actually, looking at this interface, this is also implemented by `ServePackage`, `ServeFile` and `ServeSignature`. Maybe all packages could have a "Handler", and each indexer provides an implementation, the local ones would use the `Serve*` methods, and the remote ones would use the specific remote resolver implementations.
If we do this, it could be in a future refactor.

storage/resolver.go
```diff
@@ -0,0 +1,40 @@
...
+	staticURL := resolver.artifactsStaticURL.
+		ResolveReference(&url.URL{Path: nameVersion}).
+		ResolveReference(&url.URL{Path: resourcePath})
+	http.Redirect(w, r, staticURL.String(), http.StatusSeeOther)
+}
+
+func (resolver storageResolver) RedirectSignaturesHandler(w http.ResponseWriter, r *http.Request, p *packages.Package) {
+	nameVersionSigZip := fmt.Sprintf("%s-%s.zip.sig", p.Name, p.Version)
+	signatureURL := resolver.artifactsPackagesURL.ResolveReference(&url.URL{Path: nameVersionSigZip})
+	http.Redirect(w, r, signatureURL.String(), http.StatusSeeOther)
```
jsoriano, line 37: Maybe we should use the plain 301 (`http.StatusMovedPermanently`) or 308 (`http.StatusPermanentRedirect`). 303 (`http.StatusSeeOther`) seems to be used after PUT or PUSH operations (https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#temporary_redirections) to change the method back to GET.

packages/http.go
mtojek: Keep in mind that the same function is used to serve signatures, so you can't depend on the base path.

In signatures.go:

```go
packages.ServeLocalPackage(w, r, packageList[0], packageList[0].BasePath+".sig")
```

packages/indexer.go
```diff
@@ -2,17 +2,15 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package main
+package packages
```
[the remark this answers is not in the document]
mtojek: Fixed.

packages/http.go
mtojek: Well, in my opinion, it's much clearer and readable now.

storage/resolver.go
mtojek: 301 might be permanently cached in browsers or intermittent proxy servers. If you decide to change endpoints, it might be hard to reroute traffic.

Maybe 302 (temporary)? WDYT?

packages/resolver.go
mtojek: I would leave it for further refactoring. I think that this idea is similar to the first approach I proposed (packages hijack request handling).

artifacts.go
mtojek: I tried that approach at first but decided to hide the logic of local/remote behind filesystem interfaces. I decided to revert when I had to implement the `ReadSeekCloser` interface.

I understand that you propose...

```go
		aPackage := packageList[0]
		if aPackage.RemoteResolver() != nil {
			aPackage.RemoteResolver().RedirectArtifactsHandler(w, r, aPackage)
			return
		}

		cacheHeaders(w, cacheTime)
		packages.ServeLocalPackage(w, r, packageList[0], packageList[0].BasePath)
```

... to move `if aPackage.RemoteResolver() != nil {` logic into `packages.ServeLocalPackage`, right? What do you recommend to do with `cacheHeaders`? Should this function call be also moved to `packages.ServeLocalPackage`?

jsoriano: The code here could stay the same.
```
cacheHeaders(w, cacheTime)
packages.ServePackage(w, r, packageList[0])
```
`ServePackage` would include the new logic to do the redirect if the package passed as argument includes a remote resolver.

jsoriano: Well, unless we need different cache times for redirects.

packages/http.go
jsoriano: Agree with the increased cleanness after avoiding the `else`. But this is changing from `ServeFile` to `Open` + `ServeContent`. Is this change needed? Are we sure this is not going to change any behaviour?

storage/resolver.go
mtojek: I suppose that this thread is not relevant anymore. We will reuse `cacheHeaders`.

jsoriano: Umm, are files cached beyond the cache headers we are including? I think it'd be fine if browsers do some caching of these files, even if we change endpoints, the urls should end up redirecting to the same files.
If we wouldn't use redirects, we'd have the same risk of having resources cached.

Thinking on Permanent vs. Temporary, I'd say that here we are doing a Permanent redirect, we are not serving the files anymore from the registry but from the object storage, this is a "permanent" relocation.

artifacts.go
mtojek: Fixed.

storage/resolver.go
mtojek: Fixed: `http.StatusMovedPermanently`

packages/http.go
```diff
@@ -15,11 +15,34 @@ import (
...
+func ServePackageSignature(w http.ResponseWriter, r *http.Request, p *Package, packagePath string) {
+	span, _ := apm.StartSpan(r.Context(), "ServePackageSignature", "app")
+	defer span.End()
+
+	if p.RemoteResolver() != nil {
+		p.RemoteResolver().RedirectSignaturesHandler(w, r, p)
+		return
+	}
+	serveLocalPackage(w, r, p, packagePath)
+}
```
jsoriano, line 40: Following with https://github.com/elastic/package-registry/pull/823#discussion_r907093896, good point about reusing the code for both things, but we could still get the paths from the package, do we don't need to pass both things.

```suggestion
func ServePackage(w http.ResponseWriter, r *http.Request, p *Package) {
	span, _ := apm.StartSpan(r.Context(), "ServePackage", "app")
	defer span.End()

	if p.RemoteResolver() != nil {
		p.RemoteResolver().RedirectArtifactsHandler(w, r, p)
		return
	}
	serveLocalPackage(w, r, p, p.BasePath)
}

// ServePackageSignature is used by signaturesHandler to serve signatures.
func ServePackageSignature(w http.ResponseWriter, r *http.Request, p *Package) {
	span, _ := apm.StartSpan(r.Context(), "ServePackageSignature", "app")
	defer span.End()

	if p.RemoteResolver() != nil {
		p.RemoteResolver().RedirectSignaturesHandler(w, r, p)
		return
	}
	serveLocalPackage(w, r, p, p.BasePath+".sig")
}
```

PR #825, opened 2022-06-30
Add filter to search by subscription

jsoriano:
Add a new filter `elastic.subscription`, that can be used to search for packages available in a given subscription.

New condition defined in the spec in https://github.com/elastic/package-spec/pull/355
Part of https://github.com/elastic/package-spec/issues/298


jsoriano:
Closing as this doesn't seem to be neede by now.

PR #826, opened 2022-06-30
Fill deprecated license field with the value of the new subscription condition

jsoriano:
Alternative to https://github.com/elastic/package-registry/pull/825, that only adds a compatibility layer from the old to the new field.


mtojek (commented):
It seems that you have to update golden files.

mtojek (commented):
As the next step, you can adjust the `elastic-package create` to handle it correctly.

BTW Does it need a changelog entry?

jsoriano:
Changelog added.

mtojek (approved):

Issue #797, opened 2022-03-02
package-registry observability integration


jsoriano:
Add integration to observe deployments of the package registry. This integration could collect metrics and logs, and provide dashboards. It could also link to APM data.

- [x] Add metrics endpoint https://github.com/elastic/package-registry/pull/827
- [x] Update elastic-package stack to expose metrics endpoint https://github.com/elastic/elastic-package/pull/885
- [x] Add a new integration for Elastic Package Registry 
    - Metrics and Dashboards https://github.com/elastic/integrations/pull/3671
    - Elastic Package Registry logo https://github.com/elastic/integrations/pull/3880
    - Update favicon in Package Registry service https://github.com/elastic/package-registry/pull/858

akshay-saraswat:
[Product Requirements Doc](https://docs.google.com/document/d/1C7_gs25Sb9lmv7YlC0Ff1vHAp6_VMqgOknhiWtb9Cj8)

jlind23:
Closing it as all issues are done.

PR #827, opened 2022-07-01
Add metrics to Elastic Package Registry

mrodm:
Relates #797
 
Add metrics to Elastic Package Registry. These metrics ~are being sent to Elastic Stack through the APM agent and also they~ are being exposed in the same application in the  "/metrics" endpoint.

These metrics are being instrumented using Prometheus client. The list of metrics exposed in this PR are:
- epr_service_info (Gauge)
- epr_in_flight_requests (Gauge)
- epr_number_indexed_packages (Gauge)
- epr_http_requests_total (Counter)
- epr_http_request_duration_seconds_{bucket,sum,count} (Histogram)
- epr_http_request_size_bytes_{bucket,sum,count} (Histogram)
- epr_http_response_size_bytes_{bucket,sum,count} (Histogram)
- epr_storage_indexer_updates_index_success_total (Counter)
- epr_storage_indexer_updates_index_error_total (Counter)
- epr_storage_requests_total (Counter)
- epr_storage_indexer_update_index_duration_seconds_{bucket,sum,count) (Histogram)
- epr_storage_indexer_get_duration_seconds_{bucket,sum,count} (Histogram)

Some example of metrics that are going to be exposed (to be reviewed buckets yet):

```
# HELP epr_service_info Version information about this binary
# TYPE epr_service_info gauge
epr_service_info{instance="ec4bd25798d7",version="1.9.1"} 1

# HELP epr_in_flight_requests A gauge of requests currently being served by the http server.
# TYPE epr_in_flight_requests gauge
epr_in_flight_requests 0

# HELP epr_http_requests_total A counter for requests to the http server.
# TYPE epr_http_requests_total counter
epr_http_requests_total{code="200",method="get",path="/categories"} 10
epr_http_requests_total{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip"} 30
epr_http_requests_total{code="200",method="get",path="/package/{packageName:[a-z0-9_]+}/{packageVersion}/"} 30
epr_http_requests_total{code="200",method="get",path="/search"} 40
epr_http_requests_total{code="301",method="get",path="/package/{packageName:[a-z0-9_]+}/{packageVersion}/"} 30
epr_http_requests_total{code="400",method="get",path="/package/{packageName:[a-z0-9_]+}/{packageVersion}/"} 1
epr_http_requests_total{code="404",method="get",path="/package/{packageName:[a-z0-9_]+}/{packageVersion}/"} 2


# HELP epr_http_response_size_bytes A histogram of response sizes for requests to the http server.
# TYPE epr_http_response_size_bytes histogram
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="16"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="32"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="64"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="128"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="256"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="512"} 0
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="1024"} 21
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="65536"} 31
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="262144"} 63
epr_http_response_size_bytes_bucket{code="200",method="get",path="/search",le="+Inf"} 63
epr_http_response_size_bytes_sum{code="200",method="get",path="/search"} 4.192611e+06
epr_http_response_size_bytes_count{code="200",method="get",path="/search"} 63

# HELP epr_http_request_duration_seconds A histogram of latencies for requests to the http server.
# TYPE epr_http_request_duration_seconds histogram
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.005"} 57
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.01"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.025"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.05"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.1"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.25"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.5"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="1"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="2.5"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="5"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="10"} 63
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="+Inf"} 63
epr_http_request_duration_seconds_sum{code="200",method="get",path="/search"} 0.13667412999999992
epr_http_request_duration_seconds_count{code="200",method="get",path="/search"} 63

# HELP epr_http_request_size_bytes A histogram of sizes of requests to the http server.
# TYPE epr_http_request_size_bytes histogram
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="16"} 0
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="32"} 0
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="64"} 31
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="128"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="256"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="512"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="1024"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="65536"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="262144"} 63
epr_http_request_size_bytes_bucket{code="200",method="get",path="/search",le="+Inf"} 63
epr_http_request_size_bytes_sum{code="200",method="get",path="/search"} 4348
epr_http_request_size_bytes_count{code="200",method="get",path="/search"} 63

```


mrodm:
A new middleware has been added to get all the metrics per request, and they have been exposed in /metrics following the prometheus standard.

These first metrics are:
- Total Requests.
- In flight requests.
- Histogram Request size (bytes).
- Histogram Response size (bytes).
- Histogram Duration of the request (seconds).

The prometheus client offers also a set of default metrics. From these metrics, at least "uptime" values can be obtained from "process_start_time_seconds" metric:
```
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.65694381591e+09
```
 
For all the metrics except "in flight requests", it has been added as labels: code, method and path. That would allow us to check in more detail the values, e.g. check specific endpoints.

As an example of these:
```
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.005"} 0
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.01"} 0
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.025"} 0
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.05"} 0
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.1"} 4
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.25"} 19
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="0.5"} 30
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="1"} 30
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="2.5"} 30
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="5"} 30
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="10"} 30
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip",le="+Inf"} 30
epr_http_request_duration_seconds_sum{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip"} 6.300512742
epr_http_request_duration_seconds_count{code="200",method="get",path="/epr/{packageName}/{packageName:[a-z0-9_]+}-{packageVersion}.zip"} 30
```

main.go
```diff
@@ -186,6 +189,7 @@ func initAPMTracer(logger *zap.Logger) *apm.Tracer {
 	if err != nil {
 		logger.Fatal("Failed to initialize APM agent", zap.Error(err))
 	}
+	tracer.RegisterMetricsGatherer(apmprometheus.Wrap(prometheus.DefaultGatherer))
```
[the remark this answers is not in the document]
jsoriano: I think that we can focus on the prometheus endpoint by now. It is more clear how to build an integration around that. We can add the APM metrics gatherer later, specially seeing that it is basically a one-liner.

For users of APM, some of the information related to requests can be obtained in any case with more detail using the APM data.

main.go
```diff
@@ -281,7 +285,9 @@ func getRouter(logger *zap.Logger, config *Config, indexer Indexer) (*mux.Router
 	router.HandleFunc(signaturesRouterPath, signaturesHandler)
 	router.HandleFunc(packageIndexRouterPath, packageIndexHandler)
 	router.HandleFunc(staticRouterPath, staticHandler)
+	router.Handle("/metrics", promhttp.Handler())
```
jsoriano, line 288: For production environments we probably want to expose this endpoint in other host/port, so metrics are not publicly available. We may need to add a flag for this.

util/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+				64 * 1024 * 1024,  /* 64MiB */
+				512 * 1024 * 1024, /* 512MiB */
+			},
+		},
+		[]string{"code", "method", "path"},
+	)
+)
+
+// MetricsMiddleware is a middleware used to metric every request received
+func MeticsMiddleware() mux.MiddlewareFunc {
```
jsoriano, line 88:
```suggestion
func MetricsMiddleware() mux.MiddlewareFunc {
```

util/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			route := mux.CurrentRoute(req)
+			path, _ := route.GetPathTemplate()
+			if req.RequestURI == "/metrics" {
+				next.ServeHTTP(w, req)
+				return
+			}
+			labels := prometheus.Labels{"path": path}
+			handler = promhttp.InstrumentHandlerCounter(httpRequestsTotal.MustCurryWith(labels), handler)
+			handler = promhttp.InstrumentHandlerDuration(httpRequestDurationSeconds.MustCurryWith(labels), handler)
```
jsoriano, line 108: APM also collects this kind of duration information. Though I think it is ok to have this also as a metric.

util/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+				64,
+				256,
+				512,
+				1024,              /* 1KiB */
+				64 * 1024,         /* 64KiB */
+				256 * 1024,        /* 256KiB */
+				512 * 1024,        /* 512KiB */
+				1024 * 1024,       /* 1MiB */
+				64 * 1024 * 1024,  /* 64MiB */
+				512 * 1024 * 1024, /* 512MiB */
```
jsoriano, line 80: Is there any reasoning behind the selection of these buckets? Maybe we can use the current APM data we collect now to select values according to our use case.

jsoriano (commented):
This is looking good, added some comments by now.

util/metrics.go
mrodm: No reason for those buckets, I've set them just randomly and I was checking how they fit executing some curls.

If APM exposes that data so it can be checked the actual values, I'll check it and update this list.

util/logging.go
```diff
@@ -97,13 +97,13 @@ func LoggingMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
...
+			switch r.RequestURI {
+			case "/health":
+				// Do not log requests to these endpoints
 				next.ServeHTTP(w, r)
-				return
+			default:
+				logRequest(logger, next, w, r)
 			}
-
-			logRequest(logger, next, w, r)
```
mtojek, line 106: Frankly speaking, I would log all requests. At least we know that some script/bot/logic/ hits the healthcheck. WDYT?

util/metrics.go
mtojek: With v2 changes, we don't expect any responses great than a few KBs. In case the Fleet wants to download a ZIP file, it will be redirected (HTTP 301) to the https://package-storage.elastic.co/.

util/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+			},
+		},
+		[]string{"code", "method", "path"},
+	)
+)
+
+// MetricsMiddleware is a middleware used to metric every request received
```
mtojek, line 87: nit: used to measure?

mtojek (commented):
It would be great if you post a sample `/metrics` response in the issue description.

mtojek:
My fault, I haven't read the whole thread. I can see metrics [here](https://github.com/elastic/package-registry/pull/827#issuecomment-1173886795).

Few ideas to consider:
1. Is it possible to collect also EPR version (dimension)? It would be nice to correlate metrics with rollouts.
2. Do you plan to cover also indexers (`GetPackages`, number of indexed packages, errors, search time, etc.)?

go.mod
```diff
@@ -13,13 +13,15 @@ require (
 	github.com/joeshaw/multierror v0.0.0-20140124173710-69b34d4ec901
 	github.com/magefile/mage v1.9.0
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.12.2
 	github.com/stretchr/testify v1.7.1
-	go.elastic.co/apm v1.14.0
+	go.elastic.co/apm v1.15.0
```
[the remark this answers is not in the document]
jsoriano: > The last time we bumped up APM, it resulted in the CPU 100% load :) Maybe it's worth separating the library upgrade from implementation?

I remember some issue, but wasn't it a false alarm at the end? In any case I agree on updating the library in a separate PR if possible.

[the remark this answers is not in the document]
jsoriano: > @jsoriano do you know if package-registry is under dependabot's umbrella? I mean, does it receive updates?

Not yet. Enabling it in https://github.com/elastic/package-registry/pull/829.

util/logging.go
mrodm: For context, I just changed the code here because I wanted to filter outlogs from /metrics too. As part of one of the comments now /metrics is under another host:port (http.ListenAndServe()), so at the end it was not needed.

Not a strong opinion here, I see that this filter in case of /health endpoint could be advantageous to inspect logs later, since it could flood elastic with too many GET requests to /health (they could be filtered in a KQL query too). However, this is hiding from us the actual requests (in logs) to all the endpoints.

I could remove it and log all requests, and check whether or not it causes a problem when we try to diagnose, check logs... Ans if that is the case, we could remove it.
WDYT ?

jsoriano:
> Is it possible to collect also EPR version (dimension)? It would be nice to correlate metrics with rollouts.

This kind of "global" metadata uses to be added in Prometheus as labels of a dummy metric to avoid increasing a lot the size of responses. For example in Kubernetes you find "metrics" like the following ones, instead of adding this metadata to all the metrics of a service or a node. But assigning this info to the proper events in Metricbeat is a bit tricky.
```
kube_service_info{namespace="kube-system",service="kube-dns",cluster_ip="10.96.0.10",external_name="",load_balancer_ip=""} 1
```
```
kube_node_info{node="kind-worker2",kernel_version="5.10.25-linuxkit",os_image="Ubuntu 21.04",container_runtime_version="containerd://1.5.2",kubelet_version="v1.16.15",kubeproxy_version="v1.16.15",provider_id="kind://docker/kind/kind-worker2",pod_cidr="10.244.1.0/24",internal_ip="172.20.0.3"} 1
```

In istio you can also find some versions this way: https://github.com/elastic/beats/blob/949d7ccd1f866ddbcd73edb8ba0cd7b8af7f918d/x-pack/metricbeat/module/istio/istiod/_meta/testdata/istiod.v1.7.1.plain#L357

The problem of this trick for Metricbeat is that Metricbeat groups prometheus metrics by labels, so it may be difficult to put the version in all documents, unless we modify the Metricbeat collectors.
Kubernetes metricsets need Go code exactly for that, like here: https://github.com/elastic/beats/blob/949d7ccd1f866ddbcd73edb8ba0cd7b8af7f918d/metricbeat/module/kubernetes/state_service/state_service.go#L77-L87
But this doesn't seem to be available in the generic collector used in integrations.

main.go
```diff
@@ -126,6 +133,24 @@ func initHttpProf(logger *zap.Logger) {
 	}()
 }
 
+func initMetricsServer(logger *zap.Logger) {
+	// If -dry-run=true is set, service stops here after validation
+	if dryRun {
+		os.Exit(0)
+	}
```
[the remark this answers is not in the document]
mrodm: This function was based on the `initServer` one, and I must review the usage of dryRun here.

I think the call to os.Exit(0) in initServer could also be moved to main(). I'll check that too.

util/logging.go
mrodm: > I got tempted by this when migrating to the ECS logger, but it flooded the logs quite a bit. Let's leave it for a separate PR if we do it.

Ok, I'll keep the same logic and let's try that in a separate PR.

main.go
```diff
@@ -154,11 +181,6 @@ func initServer(logger *zap.Logger) *http.Server {
 	combinedIndexer := NewCombinedIndexer(indexers...)
 	ensurePackagesAvailable(ctx, logger, combinedIndexer)
 
-	// If -dry-run=true is set, service stops here after validation
-	if dryRun {
-		os.Exit(0)
-	}
-
```
mrodm, line 161: Moved to `main()`.
It looks safe to be executed the following lines:
```go
	router := mustLoadRouter(logger, config, combinedIndexer)
	apmgorilla.Instrument(router, apmgorilla.WithTracer(apmTracer))

	return &http.Server{Addr: address, Handler: router}
```
and in main.go just do the exit call.
Do you think there could be any issue?

mtojek:
> I could try to create a metric with that info at least to be exposed in /metrics. I wanted also to add a node_name or instance label in metrics.

My intention was to fix the following use case: during the rollout of a new EPR deployment, we can easily determine whether errors are coming from the freshly deployed app.

storage/indexer.go
```diff
@@ -160,6 +160,7 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
 	defer i.m.Unlock()
 	i.cursor = storageCursor.Current
```
mtojek, line 161: Could we also save the `cursor.Current`? It's a string value. It would be great to observe how it changes.

main.go
mtojek: The idea for dry-run is to run the package validation, fail in case of an error,  or `os.Exit(0)` when it's safe. To be 100% sure, you can try running it against a path with packages and see if it doesn't skip validation.

storage/indexer.go
mrodm: If it is a string, not sure how to do that because it's not a counter neither a gauge.
What is the format of that cursor? Could it be casted to float64 that string ? If so, I could set a gauge metric and set that value on each update.

main.go
mrodm: Tested in a container, and it seems that works as expected. Here it is the output:

```
root@7f7596816c73:/package-registry# cat config.yml
package_paths:
  - /packages/development
  - /packages/production
  - /packages/staging
  - /packages/snapshot
root@7f7596816c73:/package-registry# ./package-registry --dry-run
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.958Z","log.origin":{"file.name":"package-registry/main.go","file.line":95},"message":"Package registry started","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.961Z","log.origin":{"file.name":"package-registry/main.go","file.line":247},"message":"Packages paths: /packages/development, /packages/production, /packages/staging, /packages/snapshot","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.961Z","log.origin":{"file.name":"package-registry/main.go","file.line":249},"message":"Cache time for /search: 10m0s","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.961Z","log.origin":{"file.name":"package-registry/main.go","file.line":250},"message":"Cache time for /categories: 10m0s","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.961Z","log.origin":{"file.name":"package-registry/main.go","file.line":251},"message":"Cache time for all others: 10m0s","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.962Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/development","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:54:55.966Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/production","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:01.087Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/staging","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:01.309Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/snapshot","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:02.318Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/development","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:02.807Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/production","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:02.828Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/staging","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:02.932Z","log.origin":{"file.name":"packages/packages.go","file.line":191},"message":"Searching packages in /packages/snapshot","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-07-05T11:55:02.932Z","log.origin":{"file.name":"package-registry/main.go","file.line":269},"message":"1584 package manifests loaded","ecs.version":"1.6.0"}
root@7f7596816c73:/package-registry#
```

main.go
```diff
@@ -223,6 +245,8 @@ func getPackagesBasePaths(config *Config) []string {
 
 func printConfig(logger *zap.Logger, config *Config) {
 	logger.Info("Packages paths: " + strings.Join(config.PackagePaths, ", "))
+	logger.Info("Cache time for /: " + config.CacheTimeIndex.String())
+	logger.Info("Cache time for /index.json: " + config.CacheTimeIndex.String())
```
mrodm, line 249: Just add these info messages for completeness

storage/indexer.go
mtojek: No, it might be dangerous if we forget about its format :) In the worst case, we can collect "cursor" as logs.

mrodm:
Added a new metric `epr_service_info` that contains as a labels the instance name and the EPR version running.

```
# HELP epr_service_info Version information about this binary
# TYPE epr_service_info gauge
epr_service_info{instance="ec4bd25798d7",version="1.9.1"} 1
```

search.go
```diff
@@ -32,11 +32,13 @@ func searchHandler(indexer Indexer, cacheTime time.Duration) func(w http.Respons
 			Filter: filter,
 		}
 
+		start := time.Now()
 		packages, err := indexer.Get(r.Context(), &opts)
 		if err != nil {
 			notFoundError(w, errors.Wrapf(err, "fetching package failed"))
 			return
 		}
+		util.SearchProcessDurationSeconds.Observe(time.Since(start).Seconds())
```
mrodm, line 41: Example of this metric:

```
# HELP epr_search_process_duration_seconds A histogram of package search processes.
# TYPE epr_search_process_duration_seconds histogram
epr_search_process_duration_seconds_bucket{le="0.005"} 23
epr_search_process_duration_seconds_bucket{le="0.01"} 23
epr_search_process_duration_seconds_bucket{le="0.025"} 23
epr_search_process_duration_seconds_bucket{le="0.05"} 23
epr_search_process_duration_seconds_bucket{le="0.1"} 23
epr_search_process_duration_seconds_bucket{le="0.25"} 23
epr_search_process_duration_seconds_bucket{le="0.5"} 23
epr_search_process_duration_seconds_bucket{le="1"} 23
epr_search_process_duration_seconds_bucket{le="2.5"} 23
epr_search_process_duration_seconds_bucket{le="5"} 23
epr_search_process_duration_seconds_bucket{le="10"} 23
epr_search_process_duration_seconds_bucket{le="+Inf"} 23
epr_search_process_duration_seconds_sum 0.012933142
epr_search_process_duration_seconds_count 23
```

mrodm:
> > I could try to create a metric with that info at least to be exposed in /metrics. I wanted also to add a node_name or instance label in metrics.
> 
> My intention was to fix the following use case: during the rollout of a new EPR deployment, we can easily determine whether errors are coming from the freshly deployed app.

There is a new metric named `epr_service_info` with that information. I've been checking and usually the version running is exposed in a metric like this, and not added as labels into all the metrics (as @jsoriano mentioned in this comment https://github.com/elastic/package-registry/pull/827#issuecomment-1174795119).

Considering the use case you described, this metric could be added into the plots as another metric with a constant line. Each version could have a different legend (using the labels). Something like described [here](https://jorgelbg.me/2019/02/display-the-applications-version-in-your-grafana-dashboards/)

Another option to be able to analyze that use case, it is adding the EPR version field into every request log here https://github.com/elastic/package-registry/blob/main/util/logging.go#L119

main.go
```diff
@@ -134,10 +148,15 @@ func initHttpProf(logger *zap.Logger) {
...
 func initMetricsServer(logger *zap.Logger) {
-	// If -dry-run=true is set, service stops here after validation
 	if dryRun {
-		os.Exit(0)
+		return
+	}
+
+	hostname, found := os.LookupEnv("HOSTNAME")
+	if !found {
+		hostname = defaultInstanceName
```
jsoriano, line 157: You can also try with [`os.Hostname()`](https://pkg.go.dev/os#Hostname).

main.go
```diff
@@ -143,6 +152,11 @@ func initMetricsServer(logger *zap.Logger) {
 		return
 	}
 	logger.Info("Starting http metrics in " + metricsAddress)
+	hostname, found := os.LookupEnv("HOSTNAME")
```
mtojek, line 155: I think that there is a specific function called `os.Hostname`.

util/metrics.go
```diff
@@ -104,6 +110,8 @@ var (
 // MetricsMiddleware is a middleware used to measure every request received
 func MetricsMiddleware() mux.MiddlewareFunc {
 	// Rergister all metrics
+	prometheus.MustRegister(ServiceInfo)
```
mtojek, line 113: Should `ServiceInfo` be exposed or can be private?

util/metrics.go
```diff
@@ -0,0 +1,140 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
mtojek, line 1: I wouldn't put more stuff in `util` as it will become a trash can eventually. What about another package called `metrics`?

util/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+
+			handler = promhttp.InstrumentHandlerCounter(httpRequestsTotal.MustCurryWith(labels), handler)
+			handler = promhttp.InstrumentHandlerDuration(httpRequestDurationSeconds.MustCurryWith(labels), handler)
+			handler = promhttp.InstrumentHandlerInFlight(httpInFlightRequests, handler)
+			handler = promhttp.InstrumentHandlerRequestSize(httpRequestSizeBytes.MustCurryWith(labels), handler)
+			handler = promhttp.InstrumentHandlerResponseSize(httpResponseSizeBytes.MustCurryWith(labels), handler)
+			handler.ServeHTTP(w, req)
+		})
+	}
+}
```
jsoriano, line 140: I think it'd be good to move this to its own `metrics` package. For example when using `util.ServiceInfo` I wouldn't think this is anything related to monitoring.

Alternatively, use some `Metric` prefix/suffix for metrics-related variables, or "namespace" them by putting them in some kind of `MetricsManager` that contains everything related to metrics.

mtojek (commented):
I left a few minor comments, but in general, it's close to getting merged. We can add more metrics in follow-ups.

One thing that is missing is a CHANGELOG entry.

search.go
mtojek: Let me challenge this.

Keep in mind that there are customers who need to deploy EPR and Prometheus `/metrics` is a standard endpoint that is easy to integrate with. APM isn't and we shouldn't require to install it additionally.

main.go
mrodm: Thanks! Updated to use that function

main.go
mrodm: Sure, changed

util/metrics.go
mrodm: Sure! I will put this code in a new metrics package.
I was thinking in the last changes that this kind of naming `utils.ServiceInfo` was not meaningful. So, better in its own package.

search.go
mtojek: Ok, I won't insist on adding the metric if it doesn't provide much value at the moment and there are other means to measure `/search` duration.

Regarding, the integration with APM - personally, I consider it as a heavy application to run and another component to maintaining, so I wouldn't install it. In the case of Elastic-hosted EPR, the APM is already present, so I wouldn't complain about it loud :)

main.go
```diff
@@ -126,6 +147,28 @@ func initHttpProf(logger *zap.Logger) {
...
+func initMetricsServer(logger *zap.Logger) {
+	if dryRun {
+		return
+	}
+
+	hostname, found := os.LookupEnv("HOSTNAME")
+	if !found {
+		hostname = defaultInstanceName
+	}
+	util.ServiceInfo.WithLabelValues(version, hostname).Set(1)
```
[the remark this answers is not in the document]
mrodm: Agreed, it's not very intuitive.
I'll go for the second approach, adding a specific method to set that metric for the current version information.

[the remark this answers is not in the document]
mrodm: Agreed, it's not very intuitive.
As there have been added new metrics using labels (e.g. StorageRequestsTotal), I would keep the first approach you proposed (using With() ), and set/increment all the metrics in the same way.

util/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const metricsNamespace = "epr"
+
+// info metric
```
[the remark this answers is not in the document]
mrodm: I've added those comments just to some kind of grouping. As every metric variable has a help message, I think it's not needed to add a comment too.

search.go
mrodm: As @jsoriano points out, metric `epr_http_request_duration_seconds` is quite similar. The difference was mainly that this metric (`epr_search_process_duration_seconds`) was focused on the time to perform just the `indexer.Get()` call.

From that http request duration time, there are some processes that are not considered:
- create the filter from the query
- build the final array of packages (getPackageOutput)
- write headers
- print/write the response (data from the indexer)

I see that there is FilterPackages span, that is calculated inside the Apply() method of Filters. That would be something similar to what it is considered here.

Do you think that considering that this metric could lead to a worthy measurement?

Probably the metric name could be rephrased to be more meaningful, something like:
- epr_storage_get_duration_seconds
- epr_storage_indexer_get_duration_seconds

Following that, I could also add also a new metric to get an histogram about the update index process: `epr_storage_indexer_update_index_duration_seconds`.

mrodm: I've updated the name and narrow the scope of this metric.
The change can be seen here: https://github.com/elastic/package-registry/pull/827/commits/00315215f6f90bcd159d82fcd22baa53f0b6f7f5

As a proposal I've changed the name from "epr_search_process_duration_seconds" to "epr_storage_indexer_get_duration_seconds"

And this is the difference between these two metrics after running just this curl:
- curl https://localhost:8080/search

In this way, we can track all the get calls performed. There is one call more in epr_storage_indexer_get_duration_seconds because of the ensurePackageAvailable method.

```
# HELP epr_storage_indexer_get_duration_seconds A histogram of package search processes.
# TYPE epr_storage_indexer_get_duration_seconds histogram
epr_storage_indexer_get_duration_seconds_bucket{le="0.005"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.01"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.025"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.05"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.1"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.25"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="0.5"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="1"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="2.5"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="5"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="10"} 2
epr_storage_indexer_get_duration_seconds_bucket{le="+Inf"} 2
epr_storage_indexer_get_duration_seconds_sum 4.85e-07
epr_storage_indexer_get_duration_seconds_count 2

# HELP epr_http_request_duration_seconds A histogram of latencies for requests to the http server.
# TYPE epr_http_request_duration_seconds histogram
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.005"} 0
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.01"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.025"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.05"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.1"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.25"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="0.5"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="1"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="2.5"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="5"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="10"} 1
epr_http_request_duration_seconds_bucket{code="200",method="get",path="/search",le="+Inf"} 1
epr_http_request_duration_seconds_sum{code="200",method="get",path="/search"} 0.008124299
epr_http_request_duration_seconds_count{code="200",method="get",path="/search"} 1
```

mtojek: An extra challenge/follow-up: if you can cover with metrics failed attempts of updating index/cursor, then the entire flow would be covered.

README.md
```diff
@@ -236,6 +236,19 @@ It will be listening in the given address.
 
 You can read more about this profiler and the available endpoints in the [pprof documentation](https://pkg.go.dev/net/http/pprof).
 
+## Metrics
+Package registry is instrumented to expose Prometheus metrics. These metrics are exposed under the `/metrics` endpoint.
+By default this endpoint runs `localhost:9000`. These metrics can be scraped like:
```
jsoriano, line 241: So far APM and pprof are disabled by default, should we do the same with metrics? This would also help to avoid backwards compatibility issues.

README.md
```diff
@@ -236,6 +236,19 @@ It will be listening in the given address.
 
 You can read more about this profiler and the available endpoints in the [pprof documentation](https://pkg.go.dev/net/http/pprof).
 
+## Metrics
```
jsoriano, line 239: A new line is needed here.
```suggestion
## Metrics

```

main.go
jsoriano: I meant that you could use both methods :slightly_smiling_face: the `HOSTNAME` environment variable, and `os.Hostname()`.
Consider creating a function to obtain the hostname from one of these two methods, plus the fallback of using `localhost`.

metrics/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+	prometheus.MustRegister(StorageIndexerUpdateIndexErrorsTotal)
+	prometheus.MustRegister(StorageRequestsTotal)
+
+	return func(next http.Handler) http.Handler {
+		handler := next
+
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			route := mux.CurrentRoute(req)
+			path, _ := route.GetPathTemplate()
+			labels := prometheus.Labels{"path": path}
```
jsoriano, line 39: Should we handle the error of `route.GetPathTemplate()`, and set the `path` label only when available?
```suggestion
			path, err := route.GetPathTemplate()
			if err == nil {
				labels := prometheus.Labels{"path": path}
			}
```

README.md
mrodm: Ok, I'll keep the same approach to avoid any possible backwards compatibility issue.

metrics/middleware.go
mrodm: Good catch, I missed handling this error.

As the prometheus.Labels var must have all the values defined in the metrics (path should have some value), I think I'll go with this:
 
```
 			path, err := route.GetPathTemplate()
			if err != nil {
				path := "unknown"
			}
			labels := prometheus.Labels{"path": path}
```

When I try to query urls that are not defined like `https://localhost:8080/notfound`, the `router.NotFoundHandler` is executed and there is no metric about that in metrics endpoint. That's why I think that error did not raise in my tests.

mtojek (approved):
It seems to be complete, all my comments are addressed. Well done!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Update Go version and base Ubuntu image. [#821](https://github.com/elastic/package-registry/pull/821)
 * Add support for "threat_intel" category. [#841](https://github.com/elastic/package-registry/pull/841)
+* Instrument package registry with Prometheus metrics. [#827](https://github.com/elastic/package-registry/pull/827)
```
mrodm, line 17: Should I mark this as experimental here and in the Readme, in case some metrics (e.g. namings or buckets) are updated later ?

mtojek: Nah, I would say that we can go without special labels like experimental or beta. It's just yet another feature. If it starts failing, we will bugfix it.

jsoriano (approved):
:+1:

PR #845, opened 2022-07-12
dockerfile: remove GO111MODULE env variable

kruskall:
As of Go 1.16 module-aware mode is enabled by default so
GO111MODULE=on is no longer needed.

See https://go.dev/doc/go1.16#go-command

This should also help with forward compatibility as
GOPATH and GO111MODULE will be removed in the future,

See https://github.com/golang/go/wiki/GOPATH#deprecating-and-removing-gopath-development-mode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@
 ARG GO_VERSION=1.18.3
 FROM golang:${GO_VERSION} AS builder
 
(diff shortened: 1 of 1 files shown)


jsoriano (approved):

jsoriano:
Thanks!

PR #851, opened 2022-07-26
Collect storage indexer metrics from its implementation

jsoriano:
Move collection of storage indexer metrics to its implementation, so these metrics are collected also when the combined indexer is not used, and only if the storage indexer is.

### DIFF
--- a/storage/indexer.go
+++ b/storage/indexer.go
@@ -173,6 +173,9 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
 }
 
 func (i *Indexer) Get(ctx context.Context, opts *packages.GetOptions) (packages.Packages, error) {
+	start := time.Now()
+	defer metrics.StorageIndexerGetDurationSeconds.Observe(time.Since(start).Seconds())
+
 	i.m.RLock()
 	defer i.m.RUnlock()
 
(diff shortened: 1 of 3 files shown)


storage/indexer.go
```diff
@@ -173,6 +173,9 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
 }
 
 func (i *Indexer) Get(ctx context.Context, opts *packages.GetOptions) (packages.Packages, error) {
+	start := time.Now()
+	defer metrics.StorageIndexerGetDurationSeconds.Observe(time.Since(start).Seconds())
+
```
mrodm, line 178: Moving this to each implementation, I guess we would need to add labels to be able to break down or split the duration.

This metric was intended to get the duration values for all indexers (file system, zip files and storage remote indexers). Should new metrics be added for those indexers ?

jsoriano: Well, it happens that now we are always using the `CombinedIndexer` to initialize indexers, but this wouldn't be needed in cases where we use an only indexer, in a future refactor we may forget about this and lose this metric.

The other point I see is that this metric seems to be for the storage indexer, but if used in the combined indexer, this is also going to be reported for any other indexer.

Maybe we should have an only metric, with a label for the type of indexer.

jsoriano: (this is the kind of refactor I am refering to https://github.com/elastic/package-registry/pull/852)

jsoriano:
Closing this PR as there may more changes needed, Mario will take a deeper look, thanks!

PR #852, opened 2022-07-26
Refactor index initialization to use combined indexer only where needed

jsoriano:
While looking to recent issues around the combined indexer (https://github.com/elastic/package-registry/pull/851, https://github.com/elastic/package-registry/pull/849), I have found that we are also using it when using a single indexer.

Refactor this code to avoid making the assumption that the combined indexer is always used.


mrodm (approved):
👍

PR #853, opened 2022-07-26
Rename indexer metrics related to get operations

mrodm:
Rename indexer get metrics (histogram) to be specific to each type of indexer.
According to the current indexers, this refers to be able to split Get operations depending on the file system, zip and storage indexers.

Calculations of these metrics have been moved to their specific implementations (Get method).

Relates https://github.com/elastic/package-registry/issues/797

Before this PR, the metrics exposed for Get operations were:
```
# HELP epr_storage_indexer_get_duration_seconds A histogram of latencies for get processes run by the indexer.
# TYPE epr_storage_indexer_get_duration_seconds histogram
epr_storage_indexer_get_duration_seconds_bucket{le="0.005"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.01"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.025"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.05"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.1"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.25"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="0.5"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="1"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="2.5"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="5"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="10"} 90
epr_storage_indexer_get_duration_seconds_bucket{le="+Inf"} 90
epr_storage_indexer_get_duration_seconds_sum 2.2363000000000006e-05
epr_storage_indexer_get_duration_seconds_count 90
```

With this patch proposed, the metrics exposed will be like this:
```
# HELP epr_indexer_get_duration_seconds A histogram of latencies for get processes run by the indexer.
# TYPE epr_indexer_get_duration_seconds histogram
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.005"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.01"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.025"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.05"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.1"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.25"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="0.5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="1"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="2.5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="10"} 161
epr_indexer_get_duration_seconds_bucket{indexer="FileSystemIndexer",le="+Inf"} 161
epr_indexer_get_duration_seconds_sum{indexer="FileSystemIndexer"} 0.001622019999999999
epr_indexer_get_duration_seconds_count{indexer="FileSystemIndexer"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.005"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.01"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.025"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.05"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.1"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.25"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="0.5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="1"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="2.5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="5"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="10"} 161
epr_indexer_get_duration_seconds_bucket{indexer="ZipFileSystemIndexer",le="+Inf"} 161
epr_indexer_get_duration_seconds_sum{indexer="ZipFileSystemIndexer"} 0.0005851810000000002
epr_indexer_get_duration_seconds_count{indexer="ZipFileSystemIndexer"} 161
```


metrics/metrics.go
```diff
@@ -39,32 +39,33 @@ var (
...
 		prometheus.HistogramOpts{
 			Namespace: metricsNamespace,
 			Name:      "storage_indexer_update_index_duration_seconds",
-			Help:      "A histogram of latencies for update index processes run by the indexer.",
+			Help:      "A histogram of latencies for update index processes run by the storage indexer.",
 		},
 	)
 
-	StorageIndexerGetDurationSeconds = prometheus.NewHistogram(
+	IndexerGetDurationSeconds = prometheus.NewHistogramVec(
```
jsoriano, line 62: Umm, I am wondering how useful is this metric. Looking at the sample metrics in the description, they seem to be all quite fast as these operations happen mostly in memory, and they end up all in the lower bucket.
We also have metrics in the middleware for `/search`, that may be more useful for final users, as it includes the time of the whole response.

So I wonder if maybe we can remove this metric. Or if we leave it, lower the boundaries of each bucket so they don't end up all in the lower bucket.

jsoriano (commented):
Thanks!

main.go
```diff
@@ -68,7 +68,7 @@ var (
 func init() {
 	flag.BoolVar(&printVersionInfo, "version", false, "Print Elastic Package Registry version")
 	flag.StringVar(&address, "address", "localhost:8080", "Address of the package-registry service.")
-	flag.StringVar(&metricsAddress, "metrics-address", "", "Address to expose the Prometheus metrics.")
+	flag.StringVar(&metricsAddress, "metrics-address", "", "Address to expose the Prometheus metrics (experimental). ")
```
mrodm, line 71: Just a minor change here to add this Prometheus metric endpoint is subject to changes.
Do you think it is worthy to add it ?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Rename indexer metrics related to get operations. [#853](https://github.com/elastic/package-registry/pull/853)
+
```
mrodm, line 14: Should it be moved to another section in the changelog ?

main.go
jsoriano: Sounds good to me, we may redefine something else while we start using it.

CHANGELOG.md
jsoriano: I think it can fit as a bugfix :slightly_smiling_face: Though it could be also seen as a breaking change.

metrics/metrics.go
mrodm: Currently I don't have data about these get operations in the storage indexer. 
I was expecting that this metric would be more significant in that storage indexer than in the file system indexers. I was thinking that maybe those get ops against storage indexer would be slower (just hypothesis, I don't have data about this).

CHANGELOG.md
mtojek: A bugfix is fine :)

storage/indexer.go
```diff
@@ -43,6 +46,7 @@ func NewIndexer(storageClient *storage.Client, options IndexerOptions) *Indexer
 	return &Indexer{
 		storageClient: storageClient,
 		options:       options,
+		label:         "StorageIndexer",
```
mtojek, line 49: Hm.. could it be a const instead of a structure field?

mtojek (approved):

storage/indexer.go
mrodm: I wanted to follow the same structure that it is used for FileSystemIndexers ([link](https://github.com/elastic/package-registry/blob/main/packages/packages.go#L57)) where the label is part of the struct.

metrics/metrics.go
jsoriano: Well, current implementations are only filtering a list in memory. But it is true that the operation in the storage indexer is in a mutex whose lock can have some wait time while the index is updated. We may revisit the bucket thresholds when we have more data.

jsoriano (approved):

mtojek (approved):

indexer.go
```diff
@@ -48,7 +44,7 @@ func (c CombinedIndexer) Get(ctx context.Context, opts *packages.GetOptions) (pa
 		packages = packages.Join(p)
 	}
 
-	if !opts.Filter.AllVersions {
+	if opts != nil && opts.Filter != nil && !opts.Filter.AllVersions {
```
mrodm, line 47: This change was needed after updating this branch with main. 
This Get method is called with opts as `nil` from https://github.com/elastic/package-registry/blob/main/main.go#L290

jsoriano (approved):

PR #854, opened 2022-07-27
Expose source.license through the API

jsoriano:
Add label indicating the distribution license of the package, as defined in https://github.com/elastic/package-spec/pull/355.

Required by Fleet for https://github.com/elastic/package-spec/issues/298 (https://github.com/elastic/package-spec/issues/298#issuecomment-1196023007).

### DIFF
--- a/packages/package.go
+++ b/packages/package.go
@@ -82,7 +82,7 @@ type BasePackage struct {
 	Title               *string              `config:"title,omitempty" json:"title,omitempty" yaml:"title,omitempty"`
 	Version             string               `config:"version" json:"version"`
 	Release             string               `config:"release,omitempty" json:"release,omitempty"`
-	Source              *Source              `config:"source,omitempty" json:"source,omitempty"`
+	Source              *Source              `config:"source,omitempty" json:"source,omitempty" yaml:"source,omitempty"`
 	Description         string               `config:"description" json:"description"`
 	Type                string               `config:"type" json:"type"`
 	Download            string               `json:"download" yaml:"download,omitempty"`
--- a/packages/package.go
+++ b/packages/package.go
@@ -124,7 +124,7 @@ type PolicyTemplate struct {
 
 // Source contains metadata about the source of the package and its distribution.
 type Source struct {
-	License string `config:"license" json:"license"`
+	License string `config:"license,omitempty" json:"license,omitempty" yaml:"license,omitempty"`
 }
 
 type Conditions struct {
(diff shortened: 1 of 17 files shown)


packages/package.go
```diff
@@ -121,6 +122,11 @@ type PolicyTemplate struct {
 	TemplatePath string `config:"template_path,omitempty" json:"template_path,omitempty" yaml:"template_path,omitempty"`
 }
 
+// Source contains metadata about the source of the package and its distribution.
+type Source struct {
+	License string `config:"license" json:"license"`
```
mtojek, line 127: nit: omit empty and yaml?

mtojek (approved):

packages/package.go
jsoriano: Added

packages/package.go
```diff
@@ -82,6 +82,7 @@ type BasePackage struct {
 	Title               *string              `config:"title,omitempty" json:"title,omitempty" yaml:"title,omitempty"`
 	Version             string               `config:"version" json:"version"`
 	Release             string               `config:"release,omitempty" json:"release,omitempty"`
+	Source              *Source              `config:"source,omitempty" json:"source,omitempty"`
```
andrewkroh, line 85: Do you want a `yaml:"source,omitempty"` here too? This struct inconsistently uses `yaml` tags.

jen-huang:
Thanks @jsoriano! I'll test the changes next week when I am back from PTO.

packages/package.go
jsoriano: Yeah, good point, we should take a look to these differences. @mtojek do you know why there are fields with different sets of annotations?

andrewkroh (approved):

jsoriano:
Merging this so this is available soon.

@jen-huang let me know if this is not enough after your tests, but I think we cannot do something very different to expose this value :slightly_smiling_face:

mrodm (approved):
LGTM

jen-huang:
> @jen-huang let me know if this is not enough after your tests, but I think we cannot do something very different to expose this value 🙂

I tested this for https://github.com/elastic/kibana/pull/137799 and things work fine 👍

Could I trouble you to also include a `LICENSE.txt` in one of these test packages? I need to test from the UI that we display a modal containing the license text.

jsoriano:
> Could I trouble you to also include a `LICENSE.txt` in one of these test packages? I need to test from the UI that we display a modal containing the license text.

Do you need a real package with a `LICENSE.txt`? Or it is enough with a test package? For test package you can use the [`good` package](https://github.com/elastic/package-spec/tree/cf00245987ddfa9a9688e37294d23c3c268f5cc4/test/packages/good) in the package-spec test suite.
For real packages, we have to open the discussion about how to include the license text file in built packages, in the current integrations monorepo the license is in the root of the repository, and adding headers in files may be needed to properly apply the license.

jen-huang:
No, I don't need a real package, just something to test my local EPR with. I'll try loading up the `good` package. Thanks!

jsoriano:
Packages will eventually contain the license of the repository after this change https://github.com/elastic/elastic-package/pull/920

Issue #770, opened 2021-11-16
Implement Proxy Indexer


jlind23:
The goal of this issue is to implement a new indexing mode - Proxy Indexer.

Proxy Indexer combines results from the https://epr.elastic.co/ together with other indexers. This way users don't have to download all packages to expose it, but just "link it" with the remote endpoint.

The feature can be used together with elastic-package to decrease the size of the EPR Docker image.

mtojek:
@jsoriano Please let me know what do you think about this idea.

jsoriano:
I like this idea. In combination with a local indexer it can be also a way for real users to combine custom packages with the oficial registry. Or this proxy can include more features like a local cache, that could maybe help in some use cases.

jsoriano:
We may need to revisit how we configure the registry now. How would we configure a registry that uses a proxy indexers plus a local one?

mtojek:
> In combination with a local indexer it can be also a way for real users to combine custom packages with the oficial registry. Or this proxy can include more features like a local cache, that could maybe help in some use cases.

Yes, I was thinking about extensions and extra features, but maybe after MVP :)

> We may need to revisit how we configure the registry now. How would we configure a registry that uses a proxy indexer plus a local one?

The original plan is to keep the local indexer enabled by default and make the proxy indexer switchable on a feature flag (similar to the storage indexer). I'm also fine with adjusting the configuration (YAML file).

jsoriano:
Yeah, a `--proxy-to https://epr.elastic.co` flag may work by now :slightly_smiling_face:

mtojek:
After spending some time on implementation, I realized that it won't be a true indexer, as an indexer needs access to unexposed package properties (BasePath, semVer). Also, an indexer is not aware of the call context - all properties vs. base search ones.

I will give try a different approach - more like a proxy mode than an indexer.

PR #860, opened 2022-08-04
Implement proxy mode

mtojek:
Issue: https://github.com/elastic/package-registry/issues/770

This PR implements the proxy mode to connect to EPR and combine remote resources with existing local packages.

Testing:

```bash
go run . --feature-proxy-mode --log-level debug
```


mtojek:
@dependabot merge upstream

proxymode/proxymode.go
```diff
@@ -0,0 +1,150 @@
...
+
+type ProxyOptions struct {
+	Enabled bool
+	ProxyTo string
+}
+
+func NoProxy() *ProxyMode {
+	proxyMode, err := NewProxyMode(ProxyOptions{})
+	if err != nil {
+		log.Fatalf("no proxy mode should not return an error: %v", err)
```
jsoriano, line 39: Nit. panic here?

proxymode/proxymode.go
```diff
@@ -0,0 +1,150 @@
...
+	resolver       *proxyResolver
+}
+
+type ProxyOptions struct {
+	Enabled bool
+	ProxyTo string
+}
+
+func NoProxy() *ProxyMode {
+	proxyMode, err := NewProxyMode(ProxyOptions{})
```
jsoriano, line 37: Nit. Explicitly set enabled to false here, just in case `ProxyOptions` change.
```suggestion
	proxyMode, err := NewProxyMode(ProxyOptions{Enabled: false})
```

proxymode/proxymode.go
```diff
@@ -0,0 +1,150 @@
...
+		return packages.Packages{}, nil
+	}
+	logger := util.Logger()
+
+	proxyURL := *r.URL
+	proxyURL.Host = pm.destinationURL.Host
+	proxyURL.Scheme = pm.destinationURL.Scheme
+	proxyURL.User = pm.destinationURL.User
+
+	proxyRequest, err := http.NewRequest("GET", proxyURL.String(), nil)
```
jsoriano, line 86: Nit. Use the same method in the proxy call?
```suggestion
	proxyRequest, err := http.NewRequest(r.Method, proxyURL.String(), nil)
```

proxymode/proxymode.go
```diff
@@ -0,0 +1,150 @@
...
+	if !ok {
+		return nil, errors.New("missing package name")
+	}
+
+	packageVersion, ok := vars["packageVersion"]
+	if !ok {
+		return nil, errors.New("missing package version")
+	}
+
+	urlPath := fmt.Sprintf("/package/%s/%s/", packageName, packageVersion)
```
jsoriano, line 130: Cannot we reuse the path from the original request?

search.go
```diff
@@ -13,6 +13,10 @@ import (
 	"strconv"
 	"time"
 
+	"go.uber.org/zap"
+
+	"github.com/elastic/package-registry/proxymode"
```
jsoriano, line 18: Nit. Move import to the bottom.

search.go
```diff
@@ -116,6 +135,7 @@ func getPackageOutput(ctx context.Context, packageList packages.Packages) ([]byt
 
 	// Packages need to be sorted to be always outputted in the same order
 	sort.Sort(packageList)
+	// FIXME ensure unique packageName:packageVersion
```
jsoriano, line 138: This should be  already ensured by `packages.Join()`.

main.go
```diff
@@ -310,8 +317,16 @@ func mustLoadRouter(logger *zap.Logger, config *Config, indexer Indexer) *mux.Ro
 }
 
 func getRouter(logger *zap.Logger, config *Config, indexer Indexer) (*mux.Router, error) {
-	artifactsHandler := artifactsHandler(indexer, config.CacheTimeCatchAll)
-	signaturesHandler := signaturesHandler(indexer, config.CacheTimeCatchAll)
+	proxyMode, err := proxymode.NewProxyMode(proxymode.ProxyOptions{
+		Enabled: featureProxyMode,
```
jsoriano, line 321: As this is on technical preview, should we add a warning here when enabled?

package_index.go
```diff
@@ -47,24 +52,33 @@ func packageIndexHandler(indexer Indexer, cacheTime time.Duration) func(w http.R
...
 		opts := packages.NameVersionFilter(packageName, packageVersion)
-		packages, err := indexer.Get(r.Context(), &opts)
+		pkgs, err := indexer.Get(r.Context(), &opts)
 		if err != nil {
 			logger.Error("getting package path failed", zap.Error(err))
 			http.Error(w, "internal server error", http.StatusInternalServerError)
 			return
 		}
-		if len(packages) == 0 {
+		if len(pkgs) == 0 && proxyMode.Enabled() {
```
jsoriano, line 61: I don't like so much that we need to add logic to all handlers for the proxy mode.

I guess that another option would be to implement it as an indexer, but as indexers are now we would be making unneeded calls to the remote registry if there is already a local package that matches. Is this the reason to don't use indexers?

Maybe a way to overcome this could be to add to the indexer interface a `GetFirst() (*packages.Package, error)` method, that could be used by indexers to optimize the [several] cases where we only want the first package. Most indexers would implement `GetFirst()` as returning the first package in the result of `Get()`, but the `CombinedIndexer` could implement it as breaking the loop once it finds a package in one indexer.
Or a similar approach would be to add a new `MaxResults` option to `GetOptions`, to return only the first n packages found. But an option can be ignored/forgotten, while a method in the interface needs to be implemented.

proxymode/proxymode.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return pkgs, nil
+}
+
+func (pm *ProxyMode) Categories(r *http.Request) (map[string]*packages.Category, error) {
+	if !pm.options.Enabled {
+		return map[string]*packages.Category{}, nil
+	}
+
+	panic("categories: not implemented yet")
```
jsoriano, line 113: For a follow up?

jsoriano (commented):
Looks good, added some comments.

My only concern is about needing to add logic to all handlers for this. I wonder if we can implement it as an indexer, added a comment about this. In any case I would be also ok if you prefer to don't use indexers for this.

proxymode/proxymode.go
mtojek: The `Package` method is called from different places (artifacts, signatures, etc.) From handler's perspective, we need to pick the package-index JSON instead of referring to package resources.

package_index.go
mtojek: > I guess that another option would be to implement it as an indexer, but as indexers are now we would be making unneeded calls to the remote registry if there is already a local package that matches. Is this the reason to don't use indexers?

I analyzed this case. If we want to implement an indexer then we will have to call the search endpoint (with `all=true`) on every API call. Responses would be heavy and heavier in the future. I wouldn't like to implement an artificial cache that syncs periodically with the remote Package Registry, but I'm happy to discuss other options.

proxymode/proxymode.go
mtojek: No, I forgot to git-push :)

proxymode/proxymode.go
mtojek: Fixed

proxymode/proxymode.go
mtojek: Fixed

proxymode/proxymode.go
mtojek: Not sure about this. I don't know if we want to pass the HEAD or POST requests.

search.go
mtojek: Fixed

search.go
mtojek: Fixed

main.go
mtojek: Fixed

main.go
```diff
@@ -317,14 +318,16 @@ func mustLoadRouter(logger *zap.Logger, config *Config, indexer Indexer) *mux.Ro
 }
 
 func getRouter(logger *zap.Logger, config *Config, indexer Indexer) (*mux.Router, error) {
+	if !featureProxyMode {
```
jsoriano, line 321: Should it be the other way around? :thinking: 
```suggestion
	if featureProxyMode {
```

proxymode/proxymode.go
```diff
@@ -34,9 +33,9 @@ type ProxyOptions struct {
 }
 
 func NoProxy() *ProxyMode {
-	proxyMode, err := NewProxyMode(ProxyOptions{})
+	proxyMode, err := NewProxyMode(ProxyOptions{Enabled: false})
 	if err != nil {
-		log.Fatalf("no proxy mode should not return an error: %v", err)
+		panic(errors.Wrapf(err, "no proxy mode should not return an error"))
```
jsoriano, line 38: Nit. Confusing double negation.
```suggestion
		panic(errors.Wrapf(err, "proxy mode should not return an error when disabled"))
```

package_index.go
jsoriano: > I analyzed this case. If we want to implement an indexer then we will have to call the search endpoint (with `all=true`) on every API call.

Would this still be true with the `GetFirst()` approach? The logic would be similar to what is done here in each handler, but it would be centralized in the `CombinedIndexer`: it would return the first package found without needing to look for all indexers, so it wouldn't call the proxy for packages available locally. And when it calls the proxy, it would be with the parameters of the request, not with `all=true`.

Another advantage of using an indexer is that we can add it only when the feature is enabled, without needing to check on every call if the proxy is enabled.

proxymode/proxymode.go
mtojek: Good idea, fixed.

main.go
mtojek: Right, fixed

proxymode/proxymode.go
mtojek: Rephrased

package_index.go
mtojek: Don't get me wrong, I'd love to use a dedicated indexer here, but I'm not sure if it's possible to apply the concept here.

The problem with an indexer is that it isn't aware of the calling context and it can't return correct/full package metadata. Consider following flows:

Search packages:
1. Call `indexer.Get(options)`:
1.1 Mashal the filter to query.
1.2 Make a call `/search?all=true` (heavy HTTP response)
1.3 Apply the filter.
1.4 Fill private (`fsBuilder`, `resolver`, `semver`) or skipped fields (`BasePath`)

_Problem: Unfortunately, returned packages don't contain all properties of policy templates (Package vs BasePackage), so the result returned by indexer is always incorrect._

Package resources/signature:
1. Call `indexer.Get(options)`:
1.1 Mashal the filter to query - _based on the filter we should figure out to which endpoint we should proxy request?_
1.2 Make a call, for example: [/package/apm/8.4.0/](https://epr.elastic.co/package/apm/8.4.0/)
1.3 Apply the filter.
1.4 Fill private (`fsBuilder`, `resolver`, `semver`) or skipped fields (`BasePath`)

Categories:

1. Call `indexer.Get(options)`:
1.1 Mashal the filter to query.
1.2 Make a call `/search?all=true` (heavy HTTP response) 
...

_Problem: Now, we need to pull all packages to map them into categories. We need to visit also categories in policy templates, BUT the `/search` endpoint don't expose them (all categories are group together in `categories: [ ... ]`. For the sake of `/categories`, we could use those buggy packages, but they would work accidentally (we don't analyze policy templates)._  

I might have missed something while I tried to implement this approach. Feel free to share your thoughts!

jsoriano: > I'd love to use a dedicated indexer here, but I'm not sure if it's possible to apply the concept here.

Yeah, it is a bit beyond of their concept to use indexers for this. More than specifically using them, I was wondering about a way to separate more the proxy mode without needing to modify all the handlers, so when disabled no additional code is executed.

To search packages I don't think that we would need to search all and filter, we could translate the `options.Filter` into an equivalent search query, and wouldn't be needed to filter afterwards. From the result the private fields would be filled as they are now in this PR. https://github.com/elastic/package-registry/pull/860/files#diff-d3d83ca5e1802bcf5d721013c2c366abb0d682c3e100e4307630c34f85cf6f89R93-R100

Though for categories it is true that we would need to search for all packages, and even this could be inaccurate as you mention.

Two more ideas to separate the proxy mode:
* Use a middleware in the router. This middleware would need to be able to handle every API endpoint, capture the response from the actual handlers in the response writer and interpret it, decide if needed to make a request to the proxified registry, and merge the results. May be cumbersome and not sure if always possible, but would allow to inject the proxy only when enabled.
* Follow an approach similar to the storage indexer, each registry has a new API to expose an "index" of the packages it serves. Proxy registries can periodically query this API and update their local index.

But we can continue with the current approach, every option seems to have its cons. Thanks for the discussion!

jsoriano (approved):
Ok, let's continue with current approach, unless my last comment rings some bell on how to do it without modifying the handlers :slightly_smiling_face: Thanks for the discussion!

PR #865, opened 2022-08-17
ci: transform git tag with `v` format

v1v:
### What

Pipeline helper should be the one validating and transforming the given docker tag to the right format.

### Why

This pipeline is called when a git tag release happens, and unfortunately those git tags follow the syntax `v<major>.<minor>.<patch>` where `<major>`, `<minor>` and `<patch>` are digits, i.e: `v8.3.0` but the docker tag is required to be `<major>.<minor>.<patch>` based.

Rather than delegating to the consumers to do the manipulation, let's do the transformation if starting with `v` within the pipeline itself.

### Fixes

image

### DIFF
--- a/.ci/release-package-registry-distribution.groovy
+++ b/.ci/release-package-registry-distribution.groovy
@@ -11,7 +11,6 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_TAG = "${params.DOCKER_TAG}"
     DOCKER_IMG_SOURCE = "${env.DOCKER_REGISTRY}/package-registry/distribution:production"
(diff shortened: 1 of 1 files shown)


mrodm (approved):
:+1:

.ci/release-package-registry-distribution.groovy
```diff
@@ -11,7 +11,6 @@ pipeline {
     DOCKER_REGISTRY_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_TAG = "${params.DOCKER_TAG}"
     DOCKER_IMG_SOURCE = "${env.DOCKER_REGISTRY}/package-registry/distribution:production"
```
jsoriano, line 13: `DOCKER_IMG_SOURCE` is also only used in the publishing step, would it make sense to move this env variable also there?

jsoriano (approved):
Thanks!

jsoriano (approved):

amannocci (approved):
LGTM

PR #878, opened 2022-09-20
Don't use io.ReadAll while recycling indices

mtojek:
This PR replaces `io.ReadAll` with `json.Decode` as the routine doesn't need the index content. I'm not convinced about this [structure](https://github.com/elastic/package-registry/blob/main/packages/marshaler.go#L32) as it uses `json.Unmarshaller`, but I don't have an idea how to improve it.

Right now:

Zrzut ekranu 2022-09-20 o 12 19 19

If you can observe some other places to refactor, feel free to comment.

Testing (if you want to play):

```
go run . -feature-storage-indexer -storage-indexer-bucket-internal gs://mtojek-temp/package-storage-v2 -log-level debug -httpprof ":6060"
```

`mtojek-temp` is in the `elastic-observability` namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Don't use io.ReadAll while recycling indices. [#878](https://github.com/elastic/package-registry/pull/878)
+
 ### Added
 
 ### Deprecated
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/fsouza/fake-gcs-server v1.40.1
 	github.com/gorilla/mux v1.8.0
 	github.com/joeshaw/multierror v0.0.0-20140124173710-69b34d4ec901
+	github.com/json-iterator/go v1.1.12
 	github.com/magefile/mage v1.13.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.13.0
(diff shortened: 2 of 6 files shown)


jsoriano:
> I'm not convinced about this [structure](https://github.com/elastic/package-registry/blob/main/packages/marshaler.go#L32) as it uses `json.Unmarshaller`, but I don't have an idea how to improve it.

These calls to `json.Unmarshaller` will be done for each package JSON one after the other, this should be fine memory-wise. Have you found performance problems there?

> If you can observe some other places to refactor, feel free to comment.

Other place that may be copying too many objects is `transformSearchIndexAllToPackages`, that seems to be copying all package manifests.
Maybe this transformation could be done directly by `loadSearchIndexAll`. Or `searchIndexAll` struct could contain pointers to manifests that can be reused in the transform step.

jsoriano (approved):
:+1: only by not keeping the whole index response in memory there should be improvements in memory usage.

mtojek:
> These calls to json.Unmarshaller will be done for each package JSON one after the other, this should be fine memory-wise. Have you found performance problems there?

It depends on how json manages the decode cache inside, but I'm not master of that area, so can't answer if we can spot fields to improve there.

> Other place that may be copying too many objects is transformSearchIndexAllToPackages, that seems to be copying all package manifests.

Agree, but I don't see this part "hot" in the calltree graph. Only 500 KB is used there.

I don't know if there is any json marshaller that deallocates used memory. I tried to play with [jsoniter](https://github.com/json-iterator/go), but it gave similar results (as the logic doesn't jump over fields).

Still 👍  to merge it, @jsoriano? :)

mtojek:
> Do you have a profile with the new code?

Do you mean `http://localhost:6060/debug/pprof/heap`? Yes, I dumped it, but it's easy to generate it just in case.

mtojek:
Just in case, before introducing this change:

Zrzut ekranu 2022-09-20 o 13 54 27

PR #881, opened 2022-09-21
Use JSON decoder tokenizer to parse packages from storage indexer

jsoriano:
Use JSON decoder tokenizer to parse packages from storage indexer, and add a benchmark for the "Init" process.

The benchmark allows to compare results: slightly more allocations are needed now, but about 30% less memory is used, and it is slightly faster.

This can probably be further improved by directly applying the transformations on the parsed packages. It could be also nice to benchmark just this function and not the whole Init process, though this is the heaviest part of this process.

Before:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/storage
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkInit-8                3         398302468 ns/op        235217858 B/op    648922 allocs/op
```

After:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/storage
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkInit-8                3         379386685 ns/op        168450578 B/op    650022 allocs/op
```

### DIFF
--- a/storage/index.go
+++ b/storage/index.go
@@ -38,9 +38,46 @@ func loadSearchIndexAll(ctx context.Context, storageClient *storage.Client, buck
 	defer objectReader.Close()
 
 	var sia searchIndexAll
-	err = json.NewDecoder(objectReader).Decode(&sia)
-	if err != nil {
-		return nil, errors.Wrapf(err, "can't decode the index file (path: %s)", rootedIndexStoragePath)
+	dec := json.NewDecoder(objectReader)
+
+	for dec.More() {
+		// Read everything till the "packages" key in the map.
+		token, err := dec.Token()
+		if err != nil {
+			return nil, errors.Wrapf(err, "unexpected error while reading index file")
+		}
+		if key, ok := token.(string); !ok || key != "packages" {
+			continue
+		}
+
+		// Read the opening array now.
+		token, err = dec.Token()
+		if err != nil {
+			return nil, errors.Wrapf(err, "unexpected error while reading index file")
+		}
+		if delim, ok := token.(json.Delim); !ok || delim != '[' {
+			return nil, errors.Errorf("expected opening array, found %v", token)
+		}
+
+		// Read the array of packages one by one.
+		for dec.More() {
+			var p packageIndex
+			err = dec.Decode(&p)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unexpected error parsing package from index file (token: %v)", token)
+			}
+			sia.Packages = append(sia.Packages, p)
+		}
+
+		// Read the closing array delimiter.
+		token, err = dec.Token()
+		if err != nil {
+			return nil, errors.Wrapf(err, "unexpected error while reading index file")
+		}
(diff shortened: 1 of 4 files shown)


storage/index.go
```diff
@@ -38,9 +38,46 @@ func loadSearchIndexAll(ctx context.Context, storageClient *storage.Client, buck
 	defer objectReader.Close()
 
 	var sia searchIndexAll
-	err = json.NewDecoder(objectReader).Decode(&sia)
-	if err != nil {
-		return nil, errors.Wrapf(err, "can't decode the index file (path: %s)", rootedIndexStoragePath)
+	dec := json.NewDecoder(objectReader)
+
+	for dec.More() {
```
mtojek, line 43: nit: Maybe leave a comment about why are you proposing this as it won't be super clear. We could simplify the same logic with `json.Unmarshal`.

mtojek (approved):
> This can probably be further improved by directly applying the transformations on the parsed packages. It could be also nice to benchmark just this function and not the whole Init process, though this is the heaviest part of this process.

I wouldn't preoptimize it. This is already a decent result. Let's deploy it in the cloud and see how it goes.

storage/index.go
jsoriano: Added.

jsoriano:
> I wouldn't preoptimize it. This is already a decent result. Let's deploy it in the cloud and see how it goes.

Ok, but I am leaving the comment, I think this is something that we should try, this will also save some memory.

mtojek:
> Ok, but I am leaving the comment, I think this is something that we should try, this will also save some memory.

Frankly speaking, I'm wondering if that isn't supported by [jsoniter](https://github.com/json-iterator/go). What you're doing here is iterating over next objects.

jsoriano:
Ok, I have tried to apply the transforms in place and surprisingly (to me :slightly_smiling_face:) the results don't vary a lot, so let's keep it like this.
```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/storage
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkInit-8                3         394299581 ns/op        166942168 B/op    650017 allocs/op
```

> Frankly speaking, I'm wondering if that isn't supported by [jsoniter](https://github.com/json-iterator/go). What you're doing here is iterating over next objects.

We could also give it a try, I will wait before merging this.

jsoriano:
Ok, did a quick test with jsoniter, decoding the whole document (jsoniter doesn't expose `Token()`), and it is slightly faster for our use case, but uses more memory and allocations:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/storage
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkInit-8                3         339263116 ns/op        275712688 B/op   1165652 allocs/op
```

mtojek (approved):

PR #893, opened 2022-10-06
Remove prerelease packages when experimental=true is used

jsoriano:
Remove prerelease packages from search responses when `experimental=true` is used, for packages that have non-prerelease versions available.

That is for example:
* If only `1.0.0-beta1` and `1.0.0-beta2` versions exist for a package, both are returned.
* If `1.0.0-beta1`, `1.0.0-beta2` and `1.0.0` versions exist for a package, only `1.0.0` version is returned.

This way packages published as prerelease according to semver won't be available for versions of kibana that don't implement https://github.com/elastic/kibana/issues/122973.
This is done to avoid unexpectedly publishing prerelease packages to versions of Kibana that use `experimental=true` when the single registry v2 is used.

A side effect of this change is that prereleases of new packages for older versions of Kibana will "disappear" once a GA version is published.

This implements the suggestion from https://github.com/elastic/ingest-dev/issues/1285#issuecomment-1259271623.

There is some code duplication, I am intentionally keeping it to clearly separate the code path for old and new versions of Kibana.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* Search results for requests including `experimental=true` don't return
+  prerelease versions of packages that have been released at least once as GA. [#893](https://github.com/elastic/package-registry/pull/893)
+
 ### Bugfixes
 
 ### Added
--- a/packages/packages.go
+++ b/packages/packages.go
@@ -309,6 +309,10 @@ func (f *Filter) Apply(ctx context.Context, packages Packages) Packages {
 	span, ctx := apm.StartSpan(ctx, "FilterPackages", "app")
 	defer span.End()
 
+	if f.Experimental {
+		return f.legacyApply(ctx, packages)
+	}
+
 	// Checks that only the most recent version of an integration is added to the list
 	var packagesList Packages
 	for _, p := range packages {
@@ -318,8 +322,7 @@ func (f *Filter) Apply(ctx context.Context, packages Packages) Packages {
 		}
 
 		// Skip prerelease packages by default.
-		// Ignore condition if experimental flag is used (older Kibanas).
-		if !f.Experimental && p.IsPrerelease() && !f.Prerelease {
+		if p.IsPrerelease() && !f.Prerelease {
 			continue
 		}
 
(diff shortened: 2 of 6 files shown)


jen-huang:
> A side effect of this change is that prereleases of new packages for older versions of Kibana will "disappear" once a GA version is published.

Can you give an example of this being true? I am under the assumption that with these changes, older Kibana won't get the prerelease versions.

Can you also clarify what will happen with packages such as:
```
"name": "redisenterprise",
"title": "Redis Enterprise",
"version": "0.1.1",
"release": "beta",
```
(this is an example of a beta package currently returned today using `kibanaVersion=7.14.0&experimental=true`)

jsoriano:
@jen-huang with these changes, older Kibanas will get prerelease versions only if there is no GA version of these packages.

For the example you mention, this 0.1.1 version of redisenterprise will be available for older kibanas only if there is no other GA version available for them.

If a 1.0.0 version is released, then the 0.1.1 version will stop being available for older Kibanas, but the 1.0.0 will continue being available. This is what I mean with disappearing package versions, non-GA versions stop appearing in search results, but GA versions of the package continue appearing.

So, imagine that we have:
* `redisenterprise-0.1.1`, beta, available for all versions >= 7.14.
* `redisenterprise-1.0.0`, GA, available for all versions >= 8.0.

This will happen:
* For 7.14.0 (`experimental=true&kibana.version=7.14.0`), only 0.1.1 will be available.
* For 8.5.0 (`experimental=true&kibana.version=8.5.0`), only 1.0.0 will be available. 0.1.1 won't, because a GA version has been published. This is the case I refer with versions disappearing.
* For 8.7.0 with the prerelease toggle enabled (`prerelease=true&kibana.version=8.7.0`), both 0.1.1 and 1.0.0 will be available.
* For 8.7.0 with the prerelease toggle disabled (`kibana.version=8.7.0`), only the GA versions, 1.0.0 in this case, will be available.

For all versions there will be some version of the package.

jsoriano:
I have added some [test cases](https://github.com/elastic/package-registry/pull/893/commits/b4d34eb51338a3c1a4145bcdd49d336c2b8e2ecb) that illustrate my last comment.

kpollich (approved):
LGTM based on the parent issue conversation. Thank you for the clarifying comments here as well as the additional test cases.

PR #894, opened 2022-10-14
Return experimental packages on searches with prerelease=true

jsoriano:
Return experimental packages even if `experimental=true` is not set, when `prerelease=true` is set.

Solves cases found during https://github.com/elastic/package-registry/pull/891, discussed in https://github.com/elastic/package-registry/pull/892#discussion_r988222034.


mtojek (approved):

mrodm (approved):
:+1:

PR #899, opened 2022-10-25
Update default proxy URL

mrodm:
This PR updates the default value for the parameter `proxy-to` to use the Elastic Package Registry production DNS as a proxy.


jsoriano (approved):

jsoriano:
To be merged after the switch to storage v2.

PR #900, opened 2022-10-26
Add lite distribution to be versioned along with Elastic stacks

mrodm:
As part of the available distributions, in Package Storage V2 there is a distribution named "lite" with a subset of the packages available in the production package registry.

This PR adds a new stage to version this docker image along with the production for every Elastic stack release.

Proposed docker image and tag: `docker.elastic.co/package-registry/distribution:lite-8.x.y`


jsoriano (approved):
:+1:

PR #914, opened 2022-11-24
Add subcategories definitions

jsoriano:
First stab to subcategories support.

By now it only models the categories with the subcategories, loading them from a yaml file. We can consider moving this yaml file to the spec so it can be also consumed from other places.

This PR doesn't include the changes in the API, this will come in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* New categories. []()
 * Update default value for `proxy-url` parameter to be Elastic Package Registry production. [#899](https://github.com/elastic/package-registry/pull/899)
 
 ### Deprecated
--- a/categories/categories.go
+++ b/categories/categories.go
@@ -1,0 +1,72 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package categories
+
+import (
+	_ "embed"
+	"fmt"
+	"io"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Category is a common structure for all kinds of categories.
+type Category struct {
+	Title string `yaml:"title"`
+}
+
+// MainCategory is a main category, that can have subcategories.
+type MainCategory struct {
+	Category `yaml:",inline"`
+
+	SubCategories map[string]SubCategory `yaml:"subcategories"`
+}
+
+// SubCategory is a sub-category, should be contained in a Category.
+type SubCategory struct {
+	Category `yaml:",inline"`
+}
+
+// Categories is a list of categories.
+type Categories map[string]MainCategory
+
+func (categories Categories) TitlesMap() map[string]string {
+	if len(categories) == 0 {
+		return nil
+	}
+	titles := make(map[string]string)
+	for name, category := range categories {
+		titles[name] = category.Title
+
+		for name, category := range category.SubCategories {
+			titles[name] = category.Title
+		}
+	}
+	return titles
+}
+
+// ReadCategories reads the categories from a reader.
+func ReadCategories(r io.Reader) (Categories, error) {
+	var categoriesFile struct {
+		Categories Categories `yaml:"categories"`
+	}
+	dec := yaml.NewDecoder(r)
+	err := dec.Decode(&categoriesFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode categories: %w", err)
+	}
+	// TODO: Check for duplicated categories.
+	return categoriesFile.Categories, nil
+}
+
+// MustReadCategories reads the categories from a reader and panics if there is any error.
+func MustReadCategories(r io.Reader) Categories {
+	categories, err := ReadCategories(r)
+	if err != nil {
+		panic(err)
+	}
+
+	return categories
+}
--- a/categories/categories.yml
+++ b/categories/categories.yml
@@ -1,0 +1,143 @@
+categories:
+  aws:
+    title: "AWS"
+  azure:
+    title: "Azure"
+  cloud:
+    title: "Cloud"
+  config_management:
+    title: "Config management"
+  containers:
+    title: "Containers"
+  crm:
+    title: "CRM"
+  custom:
+    title: "Custom"
+  datastore:
+    title: "Datastore"
+  elastic_stack:
+    title: "Elastic Stack"
+  enterprise_serch:
+    title: "Enterprise Search"
+    subcategories:
+      app_search:
+        title: "Application Search"
+      connector: # TODO: Do we want three different connector categories?
+        title: "Connector"
+      connector_client:
+        title: "Connector Client"
+      connector_package:
+        title: "Connector Package"
+      crawler:
+        title: "Crawler"
+      language_client:
+        title: "Language Client"
+      native_search:
+        title: "Native Search"
+      workplace_search:
+        title: "Workplace Search"
+      content_source:
+        title: "Content Source"
+      sdk_search:
+        title: "SDK"
+  google_cloud:
+    title: "Google Cloud"
+  infrastructure:
+    title: "Infrastructure"
+    subcategories:
+      analytics_engine:
+        title: "Analytics Engine"
+      big_data: # TODO: Unify "big_data" and "analytics_engine"?
+        title: "Big Data"
+      kubernetes: # TODO: Duplicate of "kubernetes"?
+        title: "Kubernetes"
+      load_balancer:
+        title: "Load Balancer"
+      message_broker: # TODO: Duplicate of "message_queue"?
+        title: "Message Broker"
+      monitoring: # TODO: Duplicate of "monitoring".
+        title: "Monitoring"
+      notification: # TODO: Duplicate of "notification".
+        title: "Notification"
+      process_manager:
+        title: "Process Manager"
+      stream_processing:
+        title: "Stream Processing"
+      virtualization:
+        title: "Virtualization Platform"
+      webserver: # TODO: Duplicate of "web"?
+        title: "Web Server"
+      websphere: # TODO: Too specific.
+        title: "WebSphere Application Server"
+  kubernetes:
+    title: "Kubernetes"
+  languages:
+    title: "Languages"
+  message_queue:
+    title: "Message Queue"
+  monitoring:
+    title: "Monitoring"
+  network:
+    title: "Network"
+  notification:
+    title: "Notification"
+  observability:
+    title: "Observability"
+    subcategories:
+      application_observability:
+        title: "Application"
+      java_observability:
+        title: "Java"
+      monitoring: # TODO: Duplicate of "monitoring" and "infrastructure/monitoring"?
+        title: "Monitoring"
+  os_system:
+    title: "OS & System"
+  productivity:
+    title: "Productivity"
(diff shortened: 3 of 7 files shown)


categories/categories.go
```diff
@@ -0,0 +1,72 @@
...
+	var categoriesFile struct {
+		Categories Categories `yaml:"categories"`
+	}
+	dec := yaml.NewDecoder(r)
+	err := dec.Decode(&categoriesFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode categories: %w", err)
+	}
+	// TODO: Check for duplicated categories.
+	return categoriesFile.Categories, nil
```
mrodm, line 61: I was thinking just about those possible duplicated categories. Could a category be defined as a subcategory of more than category?

categories/categories.yml
```diff
@@ -0,0 +1,143 @@
...
+  datastore:
+    title: "Datastore"
+  elastic_stack:
+    title: "Elastic Stack"
+  enterprise_serch:
+    title: "Enterprise Search"
+    subcategories:
+      app_search:
+        title: "Application Search"
+      connector: # TODO: Do we want three different connector categories?
```
jlind23, line 25: @mukeshelastic @nimarezainia would you be able to weight in here? Do we want three categories?

jlind23 (commented):
Should we state somewhere in the code that categories shouldn't have more than 1 nested level for now?

jsoriano:
> Should we state somewhere in the code that categories shouldn't have more than 1 nested level for now?

Current code doesn't support nesting categories. I have added also a comment to the header of the yaml file.

categories/categories.go
```diff
@@ -0,0 +1,72 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package categories
+
+import (
+	_ "embed"
```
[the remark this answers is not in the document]
jsoriano: Fixed. Thanks.

categories/categories.go
jsoriano: This won't be possible with the current approach. We will need to use conventions (like `network_security` vs. `network_monitoring`) if we want something to be a subcategory in two different areas.

mrodm (approved):

jlind23 (approved):

categories/categories.yml
mukeshelastic: yes, let's limit the subcategories to only one level depth.. category -> subcategory. And author can specify only one subcategory

jsoriano: > And author can specify only one subcategory

Does it mean that a package can have only a subcategory? We don't have this limitation with categories, and this may complicate things.

jsoriano:
Updated list of categories adding the ones I missed.

There are still some open questions, they are as TODOs in the yml, but grouping them:
1. New subcategories that were main categories before. Options:
    1. Remove main category, leave the subcategory. I'd prefer this option. To avoid breaking changes this would require the API design with the [`subcategory_of`](https://github.com/elastic/package-spec/issues/451#issuecomment-1324172123) field, what is also fine for me.
    2. Keep main category, add a new subcategory with a different name.
2. New categories that overlap with old categories (such as `datastore` vs. `database`). Options:
    1. Keep old categories, dismiss new ones.
    2. Keep old categories, but change the title to the new one.
    3. Keep both categories. I'd avoid this option as can be confusing.
3. Duplicated subcategories (such as `observability/monitoring` vs. `infrastructure/monitoring`. Options:
    1. Keep subcategories with different suffixes. I have already added suffixes to some categories.
    2. Use a single category. We would need to decide which one to keep.
    3. Decide case by case between the previous options.
    4. Allow subcategories to be in multiple main categories. This can be ambiguous as described in https://github.com/elastic/package-registry/pull/914#discussion_r1031388151
4. Too specific product categories. Such as the one for WebSphere, but Kubernetes could be another one. Options:
    1. Keep them. This is probably fine, but where to draw the line before adding more.
    2. Keep them, but only as subcategories.
    3. Decide case by case.

@jlind23 @mukeshelastic @dborodyansky please take a look to these open questions. I can also take the decision, but I may be overlooking something.

amitkanfer:
Maybe we should ask the relevant peers from the solution teams to review and signoff?
Also - i guess it's not a problem to change it later on...? @jsoriano

jlind23:
@jsoriano my answers below

> 1. New subcategories that were main categories before. Options:
    1. Remove main category, leave the subcategory. I'd prefer this option. To avoid breaking changes this would require the API design with the [`subcategory_of`](https://github.com/elastic/package-spec/issues/451#issuecomment-1324172123) field, what is also fine for me.

This one will provide more clarity and avoid duplication problems in the near future.

> 2. New categories that overlap with old categories (such as `datastore` vs. `database`). Options:
    1. Keep old categories, dismiss new ones.
    2. Keep old categories, but change the title to the new one.
    3. Keep both categories. I'd avoid this option as can be confusing.

Looking at the work done by @dborodyansky I would rather remove the old categories and display the ones. @kpollich could this be a breaking change on the UI end? Don't we have some hard coded featured integration that relies on some categories?

> 3. Duplicated subcategories (such as `observability/monitoring` vs. `infrastructure/monitoring`. Options:
    1. Keep subcategories with different suffixes. I have already added suffixes to some categories.
  
Keeping subcategories with different suffixes is the best path forward, we might have observability users eager to find monitoring related integrations under the observability umbrella and who won't take a look under the infrastructure umbrella for example.

> 4. Too specific product categories. Such as the one for WebSphere, but Kubernetes could be another one. Options:
    1. Keep them. This is probably fine, but where to draw the line before adding more.
   
We should keep them for now. The line about adding more categories was already explained by Akshay in this document:
https://docs.google.com/document/d/1pMsELVC4XbGXeFsR2gvELk5HLrLaNCujztogRzKjbOU/edit

kpollich:
> Looking at the work done by @dborodyansky I would rather remove the old categories and display the ones. @kpollich could this be a breaking change on the UI end? Don't we have some hard coded featured integration that relies on some categories?

There is some hardcoding related to categories in the custom integrations plugin here: https://github.com/elastic/kibana/blob/2c774f536e2cd0b01b8829d842684e67df2c3d57/src/plugins/custom_integrations/common/index.ts

We'll need to align this file to the new categories/subcategories concept in general, but if there are any renames we should tackle those first to avoid breakage.

jsoriano:
Thanks @jlind23 for your clarifications, a couple of follow ups:

> > 1. New subcategories that were main categories before. Options:
> >    1. Remove main category, leave the subcategory. I'd prefer this option. To avoid breaking changes this would require the API design with the [`subcategory_of`](https://github.com/elastic/package-spec/issues/451#issuecomment-1324172123) field, what is also fine for me.
> 
> This one will provide more clarity and avoid duplication problems in the near future.
> 

:+1: 

> > 2. New categories that overlap with old categories (such as `datastore` vs. `database`). Options:
> >    1. Keep old categories, dismiss new ones.
> >    2. Keep old categories, but change the title to the new one.
> >    3. Keep both categories. I'd avoid this option as can be confusing.
> 
> Looking at the work done by @dborodyansky I would rather remove the old categories and display the ones. @kpollich could this be a breaking change on the UI end? Don't we have some hard coded featured integration that relies on some categories?
>

Removing categories is not an option, at least without breaking changes. The closer option would be option two, keeping old categories, but changing the titles.
There could be a similar fourth option, that would be to keep both old and new categories, but both with the new titles.
 
> > 3. Duplicated subcategories (such as `observability/monitoring` vs. `infrastructure/monitoring`. Options:
> >    1. Keep subcategories with different suffixes. I have already added suffixes to some categories.
> 
> Keeping subcategories with different suffixes is the best path forward, we might have observability users eager to find monitoring related integrations under the observability umbrella and who won't take a look under the infrastructure umbrella for example.
> 

:+1: 

Regarding the current main `monitoring` category, should we move it under `observability`, so we have:
* `monitoring` as subcategory of `observability`.
* `monitoring_infrastructure` as subcategory of `infrastructure`.

Or keep the three categories:
* `monitoring` main category as the current one.
* `monitoring_observability` as subcategory of `observability`.
* `monitoring_infrastructure` as subcategory of `infrastructure`.

> > 4. Too specific product categories. Such as the one for WebSphere, but Kubernetes could be another one. Options:
> >    1. Keep them. This is probably fine, but where to draw the line before adding more.
> 
> We should keep them for now. The line about adding more categories was already explained by Akshay in this document: https://docs.google.com/document/d/1pMsELVC4XbGXeFsR2gvELk5HLrLaNCujztogRzKjbOU/edit

:+1:

jsoriano:
TODOs resolved, and API changed, please review.

CHANGELOG.md
```diff
@@ -8,11 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* Updated titles of some categories. [#914](https://github.com/elastic/package-registry/pull/914)
```
jsoriano, line 11: This shouldn't be a breaking change, but not sure if clients are using the titles for something apart of showing them. 

@kpollich do you know if the integrations UI uses the titles for something apart of to show them?

testdata/generated/storage-indexer/categories.json
```diff
@@ -81,7 +84,8 @@
   },
   {
     "id": "web",
-    "title": "Web",
-    "count": 14
+    "title": "Web Server",
+    "count": 14,
+    "subcategory_of": "infrastructure"
   }
```
jsoriano, line 90: These will be the changes in the API:
* Titles updated for some categories.
* `subcategory_of` added in subcategories.

CHANGELOG.md
kpollich: I don't think so - seems like the custom integrations plugin relies on the category ID's for its logic, and Integrations UI just uses the title for disply purposes https://github.com/elastic/kibana/blob/2c774f536e2cd0b01b8829d842684e67df2c3d57/src/plugins/custom_integrations/common/index.ts

jlind23 (approved):

testdata/generated/categories-experimental.json
```diff
@@ -31,22 +31,25 @@
...
+    "count": 3,
+    "subcategory_of": "observability"
   },
   {
     "id": "web",
-    "title": "Web",
-    "count": 3
+    "title": "Web Server",
+    "count": 3,
+    "subcategory_of": "infrastructure"
```
jsoriano, line 53: @kpollich with this API response, Fleet is not going to be able to get the title of the parent category, should I change this to something like this?
```suggestion
    "parent_id": "infrastructure",
    "parent_name": "Infrastructure"
```

An alternative could be to include the parent category in the response, but this could affect current versions of kibana that would receive categories with count 0.

This also makes me wonder if we want to modify the search APIs, to include packages of subcategories, when searching for a main category. That would for example be to include packages of the `web` subcategory, when looking for the `integration` category.

kpollich: > with this API response, Fleet is not going to be able to get the title of the parent category, should I change this to something like this?

It seems like the issue is that integrations can be placed in a subcategory even if there are no integrations placed in the parent category, resulting in a case where the parent category is not included in the response. 

So we have a state like this

image

I think returning the `parent_name` would be most helpful here, as it allows the UI to render the `Infrastructure` category in this case even if it does not directly contain any integrations.

jsoriano: PR updated to include the `parent_name` in the API response, @kpollich please take another look.

kpollich (approved):

PR #915, opened 2022-11-24
Package registry as proxy of package-storage

mrodm:
Fixes https://github.com/elastic/kibana/issues/145582

This PR changes the behaviour of the requests performed against package-storage.elastic.co. Instead of being a http redirect (CODE 301), it has been changed to perform the actual request against package-storage and forward the contents/headers to the user.

Example of the request when this was applied to proxy mode:
```shell
 $ curl -o file.zip -v "http://localhost:8888/epr/elastic_package_registry/elastic_package_registry-0.0.1.zip"
*   Trying 127.0.0.1:8888...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to localhost (127.0.0.1) port 8888 (#0)
> GET /epr/elastic_package_registry/elastic_package_registry-0.0.1.zip HTTP/1.1
> Host: localhost:8888
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Cache-Control: max-age=600
< Cache-Control: public
< Content-Length: 655945
< Content-Type: application/zip
< Last-Modified: Tue, 26 Jul 2022 17:36:46 GMT
< Date: Thu, 24 Nov 2022 14:59:42 GMT
< 
{ [16384 bytes data]
100  640k  100  640k    0     0  1602k      0 --:--:-- --:--:-- --:--:-- 1605k
* Connection #0 to host localhost left intact
```


proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
...
+		if len(w.Header().Values(header)) > 0 {
+			continue
+		}
+		if _, ok := acceptedHeaders[header]; !ok {
+			continue
+		}
+		for _, value := range values {
+			w.Header().Add(header, value)
+		}
+	}
```
mrodm, line 45: In this loop, just the accepted headers (defined in `acceptedHeaders` map) are added to the response headers.

Here it can be checked all the headers that are received from package-storage.elastic.co:

```
Filtered header X-Goog-Generation:[1658857006542520]
Filtered header X-Goog-Storage-Class:[MULTI_REGIONAL]
Filtered header Access-Control-Allow-Origin:[*]
Filtered header X-Goog-Metageneration:[1]
Filtered header X-Goog-Stored-Content-Encoding:[identity]
Filtered header Age:[360]
Filtered header Date:[Thu, 24 Nov 2022 14:53:42 GMT]
Adding header Content-Type:application/zip
Filtered header (already exists) Cache-Control:[public,max-age=3600]
Filtered header Alt-Svc:[h3=":443"; ma=2592000,h3-29=":443"; ma=2592000]
Filtered header X-Guploader-Uploadid:[ADPycdsyIjbm....]
Filtered header X-Goog-Stored-Content-Length:[655945]
Filtered header X-Goog-Hash:[crc32c=Zx4KuQ== md5=Iqb3+RJove6sG6lrKgaPMw==]
Filtered header Accept-Ranges:[bytes]
Adding header Content-Length:655945
Filtered header Server:[UploadServer]
Adding header Last-Modified:Tue, 26 Jul 2022 17:36:46 GMT
Filtered header Etag:["22a6f7f91268bdeeac1ba96b2a068f33"]
```

proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
...
+	"Content-Type":   "",
+	"Last-Modified":  "",
+}
+
+func (pr proxyResolver) pipeRequestProxy(w http.ResponseWriter, remotePath string) error {
+	remoteURL := pr.destinationURL.ResolveReference(&url.URL{Path: remotePath})
+
+	resp, err := http.Get(remoteURL.String())
+	if err != nil {
+		return err
```
jsoriano, line 31: If we return here without setting a status code we may be returning a 200, but this should probably be a 500.

In general in this method we should be probably producing response errors instead of returning them. The callers neither return errors.

proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
...
+		}
+		for _, value := range values {
+			w.Header().Add(header, value)
+		}
+	}
+	w.WriteHeader(resp.StatusCode)
+
+	_, err = io.Copy(w, resp.Body)
+	if err != nil {
+		return err
```
jsoriano, line 50: This should probably produce a 500 error, maybe unless the error is caused because the client has closed the connection.

proxymode/resolver.go
mrodm: Added a response with `http.StatusInternalServerError` status.

proxymode/resolver.go
mrodm: Added a response with `http.StatusInternalServerError` status in case of error.

proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
...
 func (pr proxyResolver) RedirectStaticHandler(w http.ResponseWriter, r *http.Request, p *packages.Package, resourcePath string) {
 	remotePath := fmt.Sprintf("/package/%s/%s/%s", p.Name, p.Version, resourcePath)
-	staticURL := pr.destinationURL.
-		ResolveReference(&url.URL{Path: remotePath})
-	http.Redirect(w, r, staticURL.String(), http.StatusMovedPermanently)
+
+	pr.pipeRequestProxy(w, remotePath)
 }
 
 func (pr proxyResolver) RedirectSignaturesHandler(w http.ResponseWriter, r *http.Request, p *packages.Package) {
```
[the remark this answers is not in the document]
mrodm: I was thinking in options to rename them, I could rename them to be "ForwardArtifactsHandler, ForwardStaticHandler,..." for instance.

We should keep a generic name because in proxy mode it would be interesting to keep the `http.Redirect` instead, wouldn't it ? In following changes, I was about to move this implementation just to `storage/resolver.go`

WDYT ?

proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
...
+var acceptedHeaders = map[string]string{
+	"Content-Length": "",
+	"Content-Type":   "",
+	"Last-Modified":  "",
+}
+
+func (pr proxyResolver) pipeRequestProxy(w http.ResponseWriter, remotePath string) error {
+	remoteURL := pr.destinationURL.ResolveReference(&url.URL{Path: remotePath})
+
+	resp, err := http.Get(remoteURL.String())
```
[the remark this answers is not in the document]
mrodm: Thanks for these suggestions!

> * We may want to run this get with the context of the request, so if the connection with the client is aborted, we also interrupt this download.

True, adding the same context to the forwarded request.

> * I am thinking now that we may actually want to use the method of the request. For example if the method is `HEAD`, the response shouldn't be the whole file, and we shouldn't try to download the file.

I was focused on GET requests and I missed those scenarios. :+1:

proxymode/resolver.go
```diff
@@ -16,25 +17,57 @@ type proxyResolver struct {
 	destinationURL url.URL
 }
 
+var acceptedHeaders = map[string]string{
+	"Content-Length": "",
+	"Content-Type":   "",
+	"Last-Modified":  "",
```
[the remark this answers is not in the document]
mrodm: Added

proxymode/resolver.go
mrodm: I will add `date` for completeness.

proxymode/resolver.go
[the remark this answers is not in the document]
jsoriano: Ok, sounds good to me.

jsoriano (approved):

packages/resolver.go
```diff
@@ -7,7 +7,7 @@ package packages
 import "net/http"
 
 type RemoteResolver interface {
-	RedirectArtifactsHandler(w http.ResponseWriter, r *http.Request, p *Package)
-	RedirectStaticHandler(w http.ResponseWriter, r *http.Request, p *Package, resourcePath string)
-	RedirectSignaturesHandler(w http.ResponseWriter, r *http.Request, p *Package)
+	ForwardArtifactsHandler(w http.ResponseWriter, r *http.Request, p *Package)
+	ForwardStaticHandler(w http.ResponseWriter, r *http.Request, p *Package, resourcePath string)
+	ForwardSignaturesHandler(w http.ResponseWriter, r *http.Request, p *Package)
```
jsoriano, line 12: Perhaps we can remove the verb from here, and we could have different implementations, one with redirect, another one with proxy forwarding...
```suggestion
	ArtifactsHandler(w http.ResponseWriter, r *http.Request, p *Package)
	StaticHandler(w http.ResponseWriter, r *http.Request, p *Package, resourcePath string)
	SignaturesHandler(w http.ResponseWriter, r *http.Request, p *Package)
```

jsoriano (approved):

packages/resolver.go
mrodm: That makes totally sense, let's remove the verb

jsoriano (approved):

jsoriano (approved):

PR #926, opened 2022-12-07
update docker image URL in README

endorama:
I needed to use the docker images for the package registry and found the Docker URL to be wrong.

From [Elastic air-gapped doc](https://www.elastic.co/guide/en/fleet/8.5/air-gapped.html), I got this URL: `docker.elastic.co/package-registry/distribution`

This PR replaces references to Docker URL with that one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ docker run -p 8080:8080 {image id from prior step}
 
 **Commands ready to cut-and-paste**
 ```
-docker build --rm -t docker.elastic.co/package-registry/package-registry:main .
-docker run -it -p 8080:8080 $(docker images -q docker.elastic.co/package-registry/package-registry:main)
+docker build --rm -t docker.elastic.co/package-registry/distribution:main .
+docker run -it -p 8080:8080 $(docker images -q docker.elastic.co/package-registry/distribution:main)
 ```
 
 **Listening on HTTPS**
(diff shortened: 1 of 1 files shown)


jsoriano:
> I needed to use the docker images for the package registry and found the Docker URL to be wrong.

What is wrong with the current URL?
```
$ docker run --pull always docker.elastic.co/package-registry/package-registry:main --version
main: Pulling from package-registry/package-registry
e96e057aae67: Already exists 
898dbcfeb150: Pull complete 
0d13dd9a0d7c: Pull complete 
622f75915b4d: Pull complete 
Digest: sha256:9bec6a43973f076cea7bf9ceaf08887e438823c90ccf8523974fbc7517af1de8
Status: Downloaded newer image for docker.elastic.co/package-registry/package-registry:main
Elastic Package Registry version 1.16.1
```

> From [Elastic air-gapped doc](https://www.elastic.co/guide/en/fleet/8.5/air-gapped.html), I got this URL: `docker.elastic.co/package-registry/distribution`
> 
> This PR replaces references to Docker URL with that one.

These images contain the package registry plus a distribution of packages. Here we want to use an image including only the package registry, not a distribution of packages. In some places in the readme it is mentioned that an empty image is provided.

If the problem is that the image without packages doesn't start and is not clear what to do, maybe we need to specify that the registry needs a directory with packages, or to enable the proxy mode.

endorama:
Hi @jsoriano, thanks for the answer.

I think I'm mixing up URLs here. Let me try to clarify the issue.

`docker.elastic.co/package-registry/package-registry:main` this works!

I face issue when pulling version tagged images:
```
❯ docker pull docker.elastic.co/package-registry/package-registry:v8.5.0
Error response from daemon: manifest for docker.elastic.co/package-registry/package-registry:v8.5.0 not found: manifest unknown: manifest unknown
❯ docker pull docker.elastic.co/package-registry/package-registry:8.5.0
Error response from daemon: manifest for docker.elastic.co/package-registry/package-registry:8.5.0 not found: manifest unknown: manifest unknown
```

What would be the correct way to do that?

endorama:
Thanks @jsoriano for the explanation and sorry for this PR. I'll close this as not applicable.

PR #928, opened 2022-12-07
Fix headers forwarding when forwarding artifacts requests to the package storage

jsoriano:
We were doing the headers forwarding after writing the body, but [`ResponseWriter.Write()`](https://pkg.go.dev/net/http#ResponseWriter) can set its own headers when others are not found.
Set the headers before doing any write in the response.

### DIFF
--- a/package_storage_test.go
+++ b/package_storage_test.go
@@ -214,6 +214,7 @@ func TestPackageStorage_ResolverHeadersResponse(t *testing.T) {
 	webServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Foo", "bar")
 		w.Header().Set("Last-Modified", "time")
+		w.Header().Set("Content-Type", "image/svg+xml")
 		fmt.Fprintf(w, "%s\n%s\n%+v\n", r.Method, r.RequestURI, r.Header)
 	}))
 	defer webServer.Close()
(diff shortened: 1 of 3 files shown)


package_storage_test.go
```diff
@@ -236,11 +237,14 @@ func TestPackageStorage_ResolverHeadersResponse(t *testing.T) {
...
-			path:            staticRouterPath,
-			file:            "1password-logo-light-bg.svg.response",
-			responseHeaders: map[string]string{"Last-Modified": "time"},
-			handler:         staticHandler,
+			endpoint: "/package/1password/0.1.1/img/1password-logo-light-bg.svg",
+			path:     staticRouterPath,
+			file:     "1password-logo-light-bg.svg.response",
+			responseHeaders: map[string]string{
+				"Last-Modified": "time",
+				"Content-Type":  "image/svg+xml",
```
kpollich, line 245: Thanks for adding this to the tests 👍

storage/resolver.go
```diff
@@ -36,17 +36,23 @@ func (resolver storageResolver) pipeRequestProxy(w http.ResponseWriter, r *http.
 	resp, err := client.Do(forwardRequest)
 	if err != nil {
 		http.Error(w, "error from package-storage server", http.StatusInternalServerError)
+		return
 	}
 	defer resp.Body.Close()
 
+	// Set headers before setting the body. If not, first call to w.Write will
+	// add some default values.
```
kpollich, line 44: Good find here!

kpollich (approved):
Code LGTM - thank you for tracking this down. Seems like a very not obvious bug to find. 🚀

PR #932, opened 2022-12-08
Remove range headers, and write status code before writing the body

jsoriano:


storage/resolver.go
```diff
@@ -43,15 +41,14 @@ func (resolver storageResolver) pipeRequestProxy(w http.ResponseWriter, r *http.
 	// Set headers before setting the body. If not, first call to w.Write will
 	// add some default values.
 	addRequestHeadersToResponse(w, resp)
+	w.WriteHeader(resp.StatusCode)
```
jsoriano, line 44: Just in case, according to docs the status code should be written before writing the body.

kpollich (approved):

hop-dev (approved):

PR #939, opened 2022-12-13
Add some APM instrumentation to storage indexer

jsoriano:


mrodm (approved):
Nice! 
This will give us some insights about the storage indexer, thanks!

PR #940, opened 2022-12-13
Update APM libraries to v2

jsoriano:


mrodm (approved):
Great! :muscle: 
According to the docs https://pkg.go.dev/go.elastic.co/apm/v2#readme-requirements, it looks like it is compatible even with Elastic Stack 7.15, so perfect! (same requirements as in V1)

> Tested with Go 1.15+ on Linux, Windows and MacOS.
> Requires [APM Server](https://github.com/elastic/apm-server) v6.5 or newer.

Issue #938, opened 2022-12-13
Reduce scope of the package registry as library


jsoriano:
Sub-packages in the package registry are exposed also as modules. When making refactors in the registry this forces us to maintain backwards compatibility, or risk breaking dependent projects, while the package registry is not generally intended to be used as library.

Move internal functionality to internal modules, and clarify what functionality can be used externally. For example it may make sense to keep public the structs that are mapped to JSON API objects. But almost any other functionality should be probably internal.

This change will be breaking for projects using the registry as library, and require to bump to a v2 version.

kpollich:
We haven't revisited this issue in a while, so we're closing it due to inactivity. Please reopen with a comment if you think this is a mistake. Thanks!

PR #941, opened 2022-12-13
Make logger an explicit dependency and log errors through APM

jsoriano:
This PR contains several related changes:
* Wrap logger so logged errors are captured by APM too.
* Configure APM logger for APM.
* Make logger a explicit dependency instead of being a global object, so we ensure that the expected logger is used in each case.
* Move util package to `internal`, in the context of #938.

It also adds static analysis with `staticcheck`.


internal/util/logging.go
```diff
@@ -24,70 +26,66 @@ const (
...
+	DefaultLoggerType  = ECSLogger
+	DefaultLoggerLevel = zapcore.InfoLevel
 )
 
-var logLevel = zap.LevelFlag("log-level", zap.InfoLevel, "log level (default \"info\")")
-var logType = flag.String("log-type", defaultLoggerType, "log type (ecs, dev)")
-
-var logger *zap.Logger
-var loggerMutex sync.Mutex
+var warnInvalidTypeOnce sync.Once
```
mrodm, line 33: Probably this variable could be removed. I don't see any usage of it, at least in this file (internal/util/logging.go)

main_test.go
```diff
@@ -27,25 +27,22 @@ import (
...
+	"github.com/elastic/package-registry/internal/util"
 	"github.com/elastic/package-registry/packages"
-	"github.com/elastic/package-registry/util"
 )
 
 var (
 	generateFlag       = flag.Bool("generate", false, "Write golden files")
 	testCacheTime      = 1 * time.Second
 	generatedFilesPath = filepath.Join("testdata", "generated")
+	logger             = util.NewTestLogger()
```
mrodm, line 38: Suggestion: To avoid possible confusions, what about renaming this variable to something like `devLogger` or `testLogger` ? As it is defined here in "*_test.go" files, it cannot be used in the other files, but to remark that is just used here in testing.

internal/util/logging.go
jsoriano: Good catch, I have added `staticcheck` to `mage check` so I don't miss these ones again :slightly_smiling_face:

main_test.go
jsoriano: Done :+1:

internal/util/logging.go
mrodm: That's perfect! :+1:

mrodm (approved):

magefile.go
```diff
@@ -131,5 +128,12 @@ func Clean() error {
...
 // ModTidy cleans unused dependencies.
 func ModTidy() error {
+	fmt.Println(">> fmt - go mod tidy: Generating go mod files")
 	return sh.RunV("go", "mod", "tidy")
 }
+
+// Staticcheck runs a static code analyzer.
+func Staticcheck() error {
+	fmt.Println(">> check - staticcheck: Running static code analyzer")
+	return sh.RunV("go", "run", "honnef.co/go/tools/cmd/staticcheck", "./...")
```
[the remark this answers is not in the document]
jsoriano: Fixed, thanks.

Issue #955, opened 2023-02-06
Windows support


querqueq:
I'm running the whole Elastic Stack on Windows in an air-gapped environment. Which wasn't a problem yet but now I want to utilize Elastic Agent as much as possible since it really cuts down on overhead and maintenance. I tried compiling EPR on Windows which did work. But when I run it an error occurs:

```
2023-02-06T13:18:55.123+0100    INFO    package-registry/main.go:304    Packages paths: ./package-storage
2023-02-06T13:18:55.123+0100    INFO    package-registry/main.go:305    Cache time for /: 10s
2023-02-06T13:18:55.123+0100    INFO    package-registry/main.go:306    Cache time for /index.json: 10s
2023-02-06T13:18:55.123+0100    INFO    package-registry/main.go:307    Cache time for /search: 10m0s
2023-02-06T13:18:55.124+0100    INFO    package-registry/main.go:308    Cache time for /categories: 10m0s
2023-02-06T13:18:55.124+0100    INFO    package-registry/main.go:309    Cache time for all others: 10m0s
2023-02-06T13:18:55.124+0100    INFO    package-registry/main.go:140    Package registry started
2023-02-06T13:18:55.249+0100    INFO    packages/packages.go:227        Searching packages in ./package-storage
[...]

PR #956, opened 2023-02-08
Add support for Windows

jillguyonnet:
## Issue description

EPR currently fails to run on Windows due to the use of `filepath` library in an abstracted file system.

## What does this PR do?

* Add a Jenkins testing stage on Windows
* Fix Windows compatiblity

## Why is it important?

EPR currently cannot run on Windows.

## Related issues

* Closes https://github.com/elastic/package-registry/issues/955

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -71,20 +71,42 @@ pipeline {
      Execute unit tests.
      */
     stage('Test') {
-      steps {
-        cleanup()
-        withMageEnv(){
-          dir("${BASE_DIR}"){
-            sh(label: 'Runs the (unit) tests',script: 'mage -debug test|tee tests-report.txt')
+      parallel {
+        stage('Test on Linux') {
+          options { skipDefaultCheckout() }
+          steps {
+            cleanup()
+            withMageEnv(){
+              dir("${BASE_DIR}"){
+                sh(label: 'Runs the (unit) tests',script: 'mage -debug test|tee tests-report.txt')
+              }
+            }
           }
-        }
+          post {
+            always {
+              convertGoTestResults(
+                input: "${BASE_DIR}/tests-report.txt",
+                output: "${BASE_DIR}/junit-report.xml"
+              )
+            }
       }
-      post {
-        always {
-          convertGoTestResults(
-            input: "${BASE_DIR}/tests-report.txt",
-            output: "${BASE_DIR}/junit-report.xml"
-          )
+        }
+        stage('Test on Windows') {
+          agent { label 'windows-2022' }
+          options { skipDefaultCheckout() }
+          steps {
+            cleanup()
+            dir("${BASE_DIR}"){
+              withGoEnv() {
+                goTestJUnit(options: '-v ./...', output: 'junit-report.xml')
+              }
+            }
+          }
+          post {
+            always {
+              junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/junit-report.xml")
+            }
+          }
         }
       }
     }
--- a/main_test.go
+++ b/main_test.go
@@ -574,6 +574,9 @@ func listArchivedFiles(t *testing.T, body []byte) []byte {
 	var listing bytes.Buffer
 
 	for _, f := range zipReader.File {
+		// f.Name is populated from the zip file directly and is not validated for correctness.
+		// Using filepath.ToSlash(f.Name) ensures that the file name has the expected format
+		// regardless of the OS.
 		listing.WriteString(fmt.Sprintf("%d %s\n", f.UncompressedSize64, filepath.ToSlash(f.Name)))
 
 	}
--- a/main_test.go
+++ b/main_test.go
@@ -382,7 +382,7 @@ func TestAllPackageIndex(t *testing.T) {
 	// find all manifests
 	var manifests []string
 	for _, path := range packagesBasePaths {
-		m, err := filepath.Glob(path + "/*/*/manifest.yml")
+		m, err := filepath.Glob(filepath.Join(path, "*", "*", "manifest.yml"))
 		require.NoError(t, err)
 		manifests = append(manifests, m...)
 	}
@@ -574,7 +574,7 @@ func listArchivedFiles(t *testing.T, body []byte) []byte {
 	var listing bytes.Buffer
 
 	for _, f := range zipReader.File {
-		listing.WriteString(fmt.Sprintf("%d %s\n", f.UncompressedSize64, f.Name))
+		listing.WriteString(fmt.Sprintf("%d %s\n", f.UncompressedSize64, filepath.ToSlash(f.Name)))
 
 	}
 	return listing.Bytes()
--- a/packages/datastream.go
+++ b/packages/datastream.go
@@ -111,23 +111,23 @@ type fieldEntry struct {
 }
 
 func NewDataStream(basePath string, p *Package) (*DataStream, error) {
-	fs, err := p.fs()
+	fsys, err := p.fs()
 	if err != nil {
 		return nil, err
 	}
-	defer fs.Close()
+	defer fsys.Close()
 
 	manifestPath := path.Join(basePath, "manifest.yml")
 
 	// Check if manifest exists
-	_, err = fs.Stat(manifestPath)
+	_, err = fsys.Stat(manifestPath)
 	if err != nil && os.IsNotExist(err) {
 		return nil, errors.Wrapf(err, "manifest does not exist for data stream: %s", p.BasePath)
 	}
 
 	dataStreamPath := path.Base(basePath)
 
-	b, err := ReadAll(fs, manifestPath)
+	b, err := ReadAll(fsys, manifestPath)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to read manifest: %s", err)
 	}
@@ -178,7 +178,7 @@ func NewDataStream(basePath string, p *Package) (*DataStream, error) {
 	}
 
 	pipelineDir := path.Join(d.BasePath, "elasticsearch", DirIngestPipeline)
-	paths, err := fs.Glob(path.Join(pipelineDir, "*"))
+	paths, err := fsys.Glob(path.Join(pipelineDir, "*"))
 	if err != nil {
 		return nil, err
 	}
@@ -222,11 +222,11 @@ func (d *DataStream) Validate() error {
 		return fmt.Errorf("type is not valid: %s", d.Type)
 	}
 
-	fs, err := d.packageRef.fs()
+	fsys, err := d.packageRef.fs()
 	if err != nil {
 		return err
 	}
-	defer fs.Close()
+	defer fsys.Close()
 
 	// In case an ingest pipeline is set, check if it is around
 	pipelineDir := path.Join(d.BasePath, "elasticsearch", DirIngestPipeline)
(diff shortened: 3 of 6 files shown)


.ci/Jenkinsfile
```diff
@@ -71,11 +71,28 @@ pipeline {
...
+          }
+        }
+        stage('Test on Windows') {
+          agent { label 'windows-10' }
+          options { skipDefaultCheckout() }
+          steps {
+            cleanup()
+            withMageEnv(){
+              dir("${BASE_DIR}"){
+                sh(label: 'Runs the (unit) tests',script: 'mage -debug test|tee tests-report.txt')
```
mrodm, line 93: As this is run in Windows, it should be called `bat`
it also seems it can be called `cmd` so it uses the right call under the hood.
https://github.com/elastic/apm-pipeline-library/blob/main/vars/cmd.groovy

packages/fs.go
```diff
@@ -35,23 +36,29 @@ func NewExtractedPackageFileSystem(p *Package) (*ExtractedPackageFileSystem, err
 }
 
 func (fs *ExtractedPackageFileSystem) Stat(name string) (os.FileInfo, error) {
-	return os.Stat(filepath.Join(fs.path, name))
+	return os.Stat(path.Join(fs.path, name))
```
jsoriano, line 39: I think that when accessing the file system directly, we should keep using `filepath`, that takes into account the different separator in Windows. This would be for all methods of `ExtractedPackageFileSystem`, that work on the native file system.

As a rule of thumb, if the path is used with a function of `os` package, it should be manipulated with `filepath` package.

packages/fs.go
```diff
@@ -35,23 +36,29 @@ func NewExtractedPackageFileSystem(p *Package) (*ExtractedPackageFileSystem, err
...
 }
 
 func (fs *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, err error) {
-	matches, err = filepath.Glob(filepath.Join(fs.path, pattern))
+	matches, err = filepath.Glob(path.Join(fs.path, pattern))
 	if err != nil {
 		return
 	}
 	for i := range matches {
-		match, err := filepath.Rel(fs.path, matches[i])
```
jsoriano, line 51: If we keep using `filepath` here, we would be able to continue using `filepath.Rel`.

packages/fs.go
```diff
@@ -73,8 +80,8 @@ func NewZipPackageFileSystem(p *Package) (*ZipPackageFileSystem, error) {
 	var root string
 	found := false
 	for _, f := range reader.File {
-		name := filepath.Clean(f.Name)
-		parts := strings.Split(name, string(filepath.Separator))
+		name := path.Clean(f.Name)
+		parts := strings.Split(name, "/")
 		if len(parts) == 2 && parts[1] == "manifest.yml" {
```
jsoriano, line 85: Nit. Maybe we can use path functions here instead of manipulating the path as a string.
```suggestion
		dir, base := path.Split(name)
		if path.Dir(dir) == "." && base == "manifest.yml" {
```

packages/fs.go
jsoriano: If we want to work with `path`, we could use [`os.DirFS`](https://pkg.go.dev/os#DirFS) to build a `fs.FS`.

jillguyonnet:
Hi @jsoriano thanks for the tips, I'm on a learning curve so appreciate it!

packages/fs.go
jillguyonnet: @jsoriano Seems the suggested nit doesn't work as is. I'm not exactly sure what the original intent is, but in the tests the condition
```go
if len(parts) == 2 && parts[1] == "manifest.yml"
```
is only true when the given path is `"example-1.0.1/manifest.yml"`. In this case, `path.Dir(dir)` is equal to `"example-1.0.1"`.

The other case where the file name is equal to `"manifest.yml"` is when the given path is `"example-1.0.1/data_stream/foo/manifest.yml"`, in which case `path.Dir(dir)` is equal to `"example-1.0.1/data_stream/foo"`.

There are no cases where `path.Dir(dir)` is equal to `"."` and the file name is equal to `"manifest.yml"`.

Please let me know if I'm missing something.

jsoriano:
Hey @jillguyonnet,

I am thinking that it could be easier to fix the pending issues if we use `path` everywhere, as you were originally doing. This would probably solve a couple of issues:
* The paths in the API that are currently causing the tests to fail.
* Methods of the `ExtractedPackageFileSystem` type would always work with POSIX paths, also when running on Windows.

For that, we could create the `ExtractedPackageFileSystem`s this way:
```
type ExtractedPackageFileSystem struct {
	path string
	root fs.FS
}

func NewExtractedPackageFileSystem(p *Package) (*ExtractedPackageFileSystem, error) {
	return &ExtractedPackageFileSystem{path: p.BasePath, root: os.DirFS(p.BasePath)}, nil
}
```
And we would make all operations with `path` and `fs` methods over `root`.

packages/fs.go
jillguyonnet: Thank you for the explanations and playground! I would be inclined to add comment to make the intent explicit, thereby clarifying the partial condition `dir != "." && path.Dir(dir) == "."`. Knowing this, I prefer this option.

main_test.go
```diff
@@ -574,7 +574,7 @@ func listArchivedFiles(t *testing.T, body []byte) []byte {
 	var listing bytes.Buffer
 
 	for _, f := range zipReader.File {
-		listing.WriteString(fmt.Sprintf("%d %s\n", f.UncompressedSize64, f.Name))
+		listing.WriteString(fmt.Sprintf("%d %s\n", f.UncompressedSize64, filepath.ToSlash(f.Name)))
```
jsoriano, line 577: Please add a comment here to remember that we cannot trust the format of `f.Name` and then we need to use `ToSlash`.

packages/datastream.go
```diff
@@ -177,16 +177,16 @@ func NewDataStream(basePath string, p *Package) (*DataStream, error) {
 		return nil, fmt.Errorf("invalid release: %q", d.Release)
 	}
 
-	pipelineDir := filepath.Join(d.BasePath, "elasticsearch", DirIngestPipeline)
-	paths, err := fs.Glob(filepath.Join(pipelineDir, "*"))
+	pipelineDir := path.Join(d.BasePath, "elasticsearch", DirIngestPipeline)
+	paths, err := fs.Glob(path.Join(pipelineDir, "*"))
```
jsoriano, line 181: Nit. Maybe we should start using `fsys` in general as abbreviation for filesystems to distingish from `fs` package.

packages/fs.go
```diff
@@ -28,32 +28,47 @@ type PackageFileSystem interface {
...
+	pf, ok := f.(PackageFile)
+	if !ok {
+		defer f.Close()
+		return nil, fmt.Errorf("file does not implement PackageFile interface: %q", name)
+	}
+	return pf, nil
+}
+
+func (fsys *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, err error) {
+	fs.WalkDir(fsys.root, ".", func(p string, d fs.DirEntry, err error) error {
```
jsoriano, line 58: If I recall correctly, `WalkDir` returns an error, we should handle it.

packages/fs.go
```diff
@@ -28,32 +28,47 @@ type PackageFileSystem interface {
 // ExtractedPackageFileSystem provides utils to access files in an extracted package.
 type ExtractedPackageFileSystem struct {
 	path string
```
jsoriano, line 30: I think the only reason to keep the `path` now here would be to use it in error messages so users know to what package in the filesystem we are referring to. Use it when displaying paths in error messages of this package.

packages/fs.go
```diff
@@ -28,32 +28,47 @@ type PackageFileSystem interface {
...
+func (fsys *ExtractedPackageFileSystem) Open(name string) (PackageFile, error) {
+	name = strings.TrimPrefix(name, "/")
+	f, err := fsys.root.Open(name)
 	if err != nil {
-		return
+		return nil, err
 	}
-	for i := range matches {
-		match, err := filepath.Rel(fs.path, matches[i])
+	pf, ok := f.(PackageFile)
```
jsoriano, line 49: Please add a comment here mentioning that we expect this to be a plain file, that implements this interface. It may look a bit random to convert to this interface otherwise.

packages/fs.go
```diff
@@ -28,32 +28,47 @@ type PackageFileSystem interface {
...
+	if !ok {
+		defer f.Close()
+		return nil, fmt.Errorf("file does not implement PackageFile interface: %q", name)
```
jsoriano, line 52: We could think to even `panic()` here, as all the files here should actually implement `PackageFile`. But let's stay on the safer side and keep it as an error.

jsoriano, line 52: This is a place where we may want to use `fsys.path` to display the full path to the user. Something like this:
```suggestion
		defer f.Close()
		return nil, fmt.Errorf("file does not implement PackageFile interface: %q", filepath.Join(f.path, filepath.FromSlash(name)))
```

If we do this on multiple places, you may consider to implement a helper method for that, something like this:
```
func (fsys *ExtractedPackageFileSystem) fullPath(name string) (string) {
	return filepath.Join(f.path, filepath.FromSlash(name))
}
```

jsoriano (commented):
Good job here, there were several tricky cases! Added some nitpicking.

I think the only thing I would like to see reviewed before merging would be to handle the error of `WalkDir`, and to include the `fsys.path` in error messages. For the rest it LGTM.

packages/fs.go
```diff
@@ -28,31 +29,52 @@ type PackageFileSystem interface {
...
+	pf, ok := f.(PackageFile)
+	if !ok {
+		defer f.Close()
+		return nil, fmt.Errorf("file does not implement PackageFile interface: %q", filepath.Join(fsys.path, filepath.FromSlash(name)))
+	}
+	return pf, nil
+}
+
+func (fsys *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, err error) {
+	e := fs.WalkDir(fsys.root, ".", func(p string, d fs.DirEntry, err error) error {
```
jsoriano, line 60: Nit. Error variables in go are usually called `err`.
```suggestion
	err := fs.WalkDir(fsys.root, ".", func(p string, d fs.DirEntry, err error) error {
```
Or you can probably reuse the one already defined in the function.
```suggestion
	err = fs.WalkDir(fsys.root, ".", func(p string, d fs.DirEntry, err error) error {
```
In cases where you need to distinguish between different errors in the same scope, you can use `Err` as suffix.
```suggestion
	walkErr := fs.WalkDir(fsys.root, ".", func(p string, d fs.DirEntry, err error) error {
```

jillguyonnet:
@jsoriano Thanks for the review, I addressed your comments. Very good point about `fsys.path`. I edited a bit the error messages inside
```go
func (fsys *ExtractedPackageFileSystem) Glob(pattern string) (matches []string, err error)
```
and the only place I can see that might make use of `fsys.path` would be the final error returned by `WalkDir`, so currently:
```go
fmt.Errorf("failed to obtain path under package root path (%s): %w", pattern, e)
```
I suppose there could also be a custom error message in the `Open` method.

Can you please let me know your thoughts on this and of anything else I might have missed? Thanks!

packages/fs.go
```diff
@@ -73,10 +95,12 @@ func NewZipPackageFileSystem(p *Package) (*ZipPackageFileSystem, error) {
...
-		parts := strings.Split(name, string(filepath.Separator))
-		if len(parts) == 2 && parts[1] == "manifest.yml" {
-			root = parts[0]
+		name := path.Clean(f.Name)
+		dir, base := path.Split(name)
+		dir = path.Clean(dir)
+		// Find manifest files at the root of the package
+		if dir != "." && path.Dir(dir) == "." && base == "manifest.yml" {
+			root = dir
 			found = true
```
jsoriano, line 104: Nit. Not related to this change, but `found` is redundant here, right? We can also check if `root != ""`.

jsoriano (approved):

packages/fs.go
jillguyonnet: True, I had assumed that `found` was used for extra readability.

mrodm (approved):
Great job here!

PR #975, opened 2023-03-21
Update go version to 1.20.2

mrodm:
Update go versin used by package-registry to 1.20.2:
Notes about this version: https://groups.google.com/g/golang-announce/c/3-TpUx48iQY


jsoriano (approved):

jsoriano:
Find here the full release notes for 1.20: https://go.dev/doc/go1.20

I found interesting that performance regressions introduced in 1.18 and 1.19 for generics support have been fixed.

And not so relevant here, but can be useful in package-spec for example, errors can now wrap multiple errors.

PR #1011, opened 2023-05-05
Update provider settings

mrodm:
This PR adds new settings into the provider settings (Github) for the main Buildkite pipeline:
- `build_tags`: enable to run builds when tags are pushed
- `filter_enable` and `filter_condition`:  Add a new condition on the branches that Buildkite should run ([doc](https://buildkite.com/docs/pipelines/conditionals#conditionals-in-pipelines)) 
    - Allow builds that are not coming from pull requests
    - Allow builds from Pull Requests created just by the `elasticmachine` user
- `branch_configuration`: filter the branches or tags that trigger new builds in Buildkite ([doc](https://buildkite.com/docs/pipelines/branch-configuration))
    - Just allow `main` and tags (or branches) starting with `v1.*`  or `v2.*`. So it allows builds when tags like v1.20.0 or v2.0.1 are pushed to the repository

Related docs: https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -25,6 +25,10 @@ spec:
       provider_settings:
         build_pull_request_forks: false
         build_pull_requests: true
+        build_tags: true
+        filter_enabled: true
+        filter_condition: |
+          build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
       teams:
         ecosystem: {}
         ingest-fp:
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -22,6 +22,7 @@ spec:
     spec:
       repository: elastic/package-registry
       pipeline_file: ".buildkite/pipeline.yml"
+      branch_configuration: "main v1.* v2.*"
       provider_settings:
         build_pull_request_forks: false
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -22,9 +22,14 @@ spec:
     spec:
       repository: elastic/package-registry
       pipeline_file: ".buildkite/pipeline.yml"
+      branch_configuration: "main v1.* v2.*"
```
alexsapran, line 25: My only concern is that every time we release a new major we should update this `branch_configuration`. 
We could avoid this by saying `v*`?

mrodm: Yes, that would allow every tag pushed. The only drawback is that someone pushes a branch starting with `v*` (just an example vulnerability_package), that would also trigger a build in Buildkite.

Maybe, for now we can try that other glob pattern that is more "open". And if needed, we can restrict (add more globs with v1.*, v2.*) later. Agreed ? @alexsapran

alexsapran: +++

alexsapran (approved):

PR #1023, opened 2023-05-22
ci(buildkite): add test on windows step

leo-ri:


.buildkite/scripts/run-tests.ps1
```diff
@@ -0,0 +1,46 @@
...
+    go install github.com/magefile/mage@v$version
+}
+
+fixCRLF
+withGolang $env:SETUP_GOLANG_VERSION
+withMage $env:SETUP_MAGE_VERSION
+withGoJUnitReport
+
+$ErrorActionPreference = "Continue" # set +e
+mage -debug test > test-report.txt
```
alexsapran, line 38: Do we need the `-debug`?

leo-ri: @alexsapran in the Jenkins job we call it with `debug`, I can remove it, just decided do the same since I'm not sure which problems can occur with Mage

PR #1048, opened 2023-06-29
Buildkite: version from .go-version

leo-ri:
A little update for SETUP_GOLANG_VERSION variable > it will be setup on pre-command
Add `fail-build-on-error: true` for plugin annotate, for failing if there are fails in tests (like on go-concert repo)
Test output for win step

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,8 +1,8 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
 env:
+    # SETUP_GOLANG_VERSION from .go-version on post-checkout hook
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,8 +1,8 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
 env:
+  # SETUP_GOLANG_VERSION from .go-version on post-checkout hook
```
mrodm, line 4: Remove this commented line.

mrodm (approved):
Thanks @leo-ri !
Just remove the commented lines (SETUP_GOLANG_VERSION) if they are not used anymore.

PR #1049, opened 2023-07-04
Clean Jenkins resources

jlind23:
As we are migrating to buildkite, we can now clean Jenkins related resources.


alexsapran (approved):

mrodm (changes requested):
This should wait until fleet-server changes the trigger step from Jenkins to Buildkite (removing the trigger step in Jenkinsfile ?).

https://github.com/elastic/fleet-server/blob/14ff6ccee3c26cb479dce86819c0369c426eaaec/.ci/Jenkinsfile#L213



And also, dry-run mode should be removed from this pipeline when that happens as well as prefixes in docker (if they were added, not sure): @leo-ri 

- https://github.com/elastic/package-registry/blob/cf158460a7503395670d0471b056f1d0022fdc2c/.buildkite/release-package-registry-distribution.yml#L40

leo-ri:
> This should wait until fleet-server changes the trigger step from Jenkins to Buildkite (removing the trigger step in Jenkinsfile ?).
> 
> https://github.com/elastic/fleet-server/blob/14ff6ccee3c26cb479dce86819c0369c426eaaec/.ci/Jenkinsfile#L213
> 
> And also, dry-run mode should be removed from this pipeline when that happens as well as prefixes in docker (if they were added, not sure): @leo-ri
> 
> * https://github.com/elastic/package-registry/blob/cf158460a7503395670d0471b056f1d0022fdc2c/.buildkite/release-package-registry-distribution.yml#L40

yes:  DRY_RUN: false (no need to remove from pipeline - if not present it defaults to 'true') and yes, need to remove prefix - I left TODOs there.

If fleet-server is not pointing yet, probably need to create a PR separately with those mini changes first (I can do that), once fleet-server is pointing to buildkite, then this PR can be merged)

jlind23:
@leo-ri would be great if you can prepare the fleet server PR and then we will coordinate both PR to land one after the other.

jlind23:
@mrodm could you please review this again? Fleet-server-package-registry pipeline has been turned in the fleet-server repo.

.ci/jobs/defaults.yml
```diff
@@ -1,23 +0,0 @@
-
```
mrodm, line 1: It looks like this file should be removed in a following PR.

In a first Pull Request, all files except `.ci/defaults.yml` should be removed. That would delete the jobs from the server.

In a second Pull Request, remove the whole `.ci` folder

@jlind23

mrodm:
> @mrodm could you please review this again? Fleet-server-package-registry pipeline has been turned in the fleet-server repo.

@jlind23 In fleet-server repo, the stage to trigger the package-registry job is still there:
https://github.com/elastic/fleet-server/blob/main/.ci/Jenkinsfile#L205-L221

As the pipelines would be removed, probably there would not be any problem.

Regarding 8.9 branch (branch for next releases), there related pipeline is already present. Therefore, it looks that the new docker distribution images would be built and pushed from that branch too :+1: :
https://github.com/elastic/fleet-server/blob/8.9/.buildkite/pipeline.yml#L124-L134

mrodm (commented):
Looking how other Jenkins pipelines were removed. 
It looks like first it needs to be removed all files except `defaults.yaml`. And then once that PR is removed, it can be removed the whole `.ci` folder from the repository.

cc @jlind23 @leo-ri

mrodm:
Missing to change also here DRY_RUN to false so the CI jenkins job to trigger the remote update version performs/creates the needed pull requests:

https://github.com/elastic/package-registry/pull/1049/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R83

cc @leo-ri @jlind23

mrodm (approved):
LGTM

PR #1054, opened 2023-07-12
Add capabilities parameter in search endpoint

### LINKED ISSUE #1051: Support search based on capabilities
Add a new parameter `capabilities`, that can be used to filter out packages that require capabilities not supported by Kibana. Builds of Kibana that include a limited set of capabilities will need to specify the ones they support.

If the parameter is included, packages included in response will be packages:
* Without any capability condition.
* When they have a capability condition, the ones whose required capabilities are included in the parameter.

Capability conditions in packages are defined in https://github.com/elastic/package-spec/issues/560.

mrodm:
Fixes #1051

This PR adds a new filter parameter `capabilities` for the search endpoint.

Current behavior when `capabilities` is defined:
- If package has no capabilities defined
    - package can be added to the response.
- If package has capabilities defined:
    - package can be added to the response if all defined capabilities in the package manifest are present in the `capabilities` query parameter.

For the latter scenario, let's consider the following capabilities for a given package:
```yaml
conditions:
  elastic.capabilities:
    - observability
    - uptime
```
here it is shown if the package could be added to the response or not (let's consider that there are no other filters):
| Query parameter | Package can be added | 
| --- | --- |
| No capability in the query | yes |
| capabilities=observability | no |
| capabilities=observability,security | no |
| capabilities=observability,uptime | yes |
| capabilities=observability,uptime,security | yes |

## How to test

1. Build elastic-package using the latest changeset from package-spec
   ```shell
   cd /path/to/elastic-package
   go mod edit -replace github.com/elastic/package-spec/v2=github.com/elastic/package-spec/v2@72f19e9
   go mod tidy
   make build
   ``` 
2. Update some test packages to include capabilities in its manifest
    - It requires to be `format_version` >= 2.10.0
    - Update also version in manifest and changelog (e.g. 0.0.9), so it is easier to check when querying the Elastic Package Registry.
   ```diff
    $ git diff
   diff --git packages/elastic_package_registry/manifest.yml packages/elastic_package_registry/manifest.yml
   index b8a41f9b5..0d819ff0e 100644
   --- packages/elastic_package_registry/manifest.yml
   +++ packages/elastic_package_registry/manifest.yml
   @@ -1,4 +1,4 @@
   -format_version: 1.0.0
   +format_version: 2.10.0
    name: elastic_package_registry
    title: "Elastic Package Registry"
    version: 0.0.7
   @@ -9,6 +9,9 @@ categories:
    conditions:
      kibana.version: "^8.0.0"
      elastic.subscription: basic
   +  elastic.capabilities:
   +    - observability
   +    - security
    icons:
      - src: /img/elastic_package_registry-logo.svg
        title: Elastic Package Registry logo
   ```
3. Build the package
   ```shell
   elastic-package build
   ```
4. Build a new local docker image for package-registry
   ```shell
   cd /path/to/package-registry   
   ```
5. Run local package-registry using the packages directory with the built packages as a volume mount point and using the Proxy feature:
   ```shell
   docker run --rm -p 8080:8080 -e EPR_FEATURE_PROXY_MODE=true -e EPR_PROXY_TO=https://epr.elastic.co -v /home/mariorodriguez/Coding/work/integrations/build/packages/:/packages/package-registry -it docker.elastic.co/package-registry/package-registry:v1.21.0-rc1
   ```
   ```shell
   # elastic_package_registry package should not be in the response
   # - no capabilities matching
   curl -s "http://localhost:8080/search?capabilities=uptime&prerelease=true" | jq -r '.[]|"\(.name) - \(.version) \(.conditions.elastic.capabilities)"'
   # - missing `security` capability
   curl -s "http://localhost:8080/search?capabilities=observability&prerelease=true" | jq -r '.[]|"\(.name) - \(.version) \(.conditions.elastic.capabilities)"'

   # elastic_package_registry package should be in the response
   # - all the capabilities defined in the package are in the `capabilities` query parameter
   curl -s "http://localhost:8080/search?capabilities=observability,security,uptime&prerelease=true"  | jq -r '.[]|"\(.name) - \(.version) \(.conditions.elastic.capabilities)"'
   curl -s "http://localhost:8080/search?capabilities=observability,security&prerelease=true"  | jq -r '.[]|"\(.name) - \(.version) \(.conditions.elastic.capabilities)"'
   ```


packages/package.go
```diff
@@ -387,6 +388,20 @@ func (p *Package) HasKibanaVersion(version *semver.Version) bool {
...
 }
 
+func (p *Package) HasCapabilities(capabilities []string) bool {
+	if p.Conditions == nil || p.Conditions.Elastic == nil || p.Conditions.Elastic.Capabilities == nil || capabilities == nil {
+		return true
+	}
+
+	// TODO if there are several capabilities in the query, should all of them match with the ones defined in the package ?
+	for _, c := range capabilities {
+		if !util.StringsContains(p.Conditions.Elastic.Capabilities, c) {
```
jsoriano, line 398: I think it should be the other way around. For each capability in the package, there should be a matching one in the declared capabilities.
```suggestion
	for _, requiredCapability := range p.Conditions.Elastic.Capabilities {
		if !util.StringsContains(capabilities, requiredCapabilities)
```

packages/packages_test.go
```diff
@@ -350,6 +384,60 @@ func TestPackagesFilter(t *testing.T) {
...
+			Filter: Filter{
+				Capabilities: []string{"observability"},
+			},
+			Expected: []filterTestPackage{
+				{Name: "apache", Version: "1.0.0"},
+				{Name: "nginx", Version: "2.0.0"},
+				{Name: "redisenterprise", Version: "1.0.0"},
+				{Name: "obs_package", Version: "1.1.0"},
+				{Name: "obs_sec_package", Version: "2.0.0"},
+				{Name: "obs_sec_uptime_package", Version: "2.0.0"},
```
jsoriano, line 409: These packages shouldn't appear here, because they require both observability and security, and only observability is provided here.
```suggestion
				{Name: "obs_sec_package", Version: "2.0.0"},
				{Name: "obs_sec_uptime_package", Version: "2.0.0"},
```

packages/package.go
```diff
@@ -387,6 +388,20 @@ func (p *Package) HasKibanaVersion(version *semver.Version) bool {
 	return p.Conditions.Kibana.constraint.Check(version)
 }
 
+func (p *Package) HasCapabilities(capabilities []string) bool {
```
jsoriano, line 391: Maybe we should use a different name. This could refer to the fact that the package requires the passed capabilties to properly work.
```suggestion
func (p *Package) WorksWithCapabilities(capabilities []string) bool {
```

packages/package.go
mrodm: So all the capabilities defined in the package must be present in the query in order to be added to the response ? @jsoriano

jsoriano: Yes. Clients (aka Kibana) will indicate all the capabilities they support. We should only return packages whose required capabilities are satisfied in this list.

mrodm: Updated logic to do it like this.
Example in the description has also been updated taken into account this logic.

----

For the latter scenario, let's consider the following capabilities for a given package:
```yaml
conditions:
  elastic.capabilities:
    - observability
    - uptime
```
here it is shown if the package could be added to the response or not (let's consider that there are no other filters):
| Query parameter | Package can be added | 
| --- | --- |
| No capability in the query | yes |
| capabilities=observability | no |
| capabilities=observability,security | no |
| capabilities=observability,uptime | yes |
| capabilities=observability,uptime,security | yes |

packages/packages_test.go
```diff
@@ -350,6 +384,60 @@ func TestPackagesFilter(t *testing.T) {
 				{Name: "etcd", Version: "1.0.0-rc2"},
 			},
 		},
+		{
+			Title: "non existing capabilities search",
```
jsoriano, line 388: :+1:

jsoriano (approved):

PR #1057, opened 2023-07-17
Buildkite: avoid buildkite collapsing

leo-ri:
Right now buildkite collapsing groups by '---', this could be confusing reading test output. here I'm changing '---' to '----', only for the output (the file is the same)

### DIFF
--- a/.buildkite/scripts/run-tests.sh
+++ b/.buildkite/scripts/run-tests.sh
@@ -8,10 +8,13 @@ with_mage
 with_go_junit_report
 
 set +e
-mage -debug test | tee tests-report-linux.txt
+mage -debug test > tests-report-linux.txt
 exit_code=$?
-
 set -e
+
+# Buildkite collapse logs under --- symbols
+# need to change --- to anything else or switch off collapsing (note: not available at the moment of this commit)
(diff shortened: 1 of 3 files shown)


.buildkite/scripts/run-tests.sh
```diff
@@ -8,11 +8,14 @@ with_mage
...
 set +e
-mage -debug test | tee tests-report-linux.txt
+mage -debug test > tests-report-linux.txt
 exit_code=$?
-
 set -e
 
+# Buildkite collapse logs under --- symbols
+# need to change --- to anything else or switch off collapsing (note: not available at the moment of this commit)
+awk '{gsub("---", "----"); print }' tests-report-linux.txt
```
mrodm, line 17: Could you try to make it fail some test ? 
As this file is used by `go-junit-report` to be sure that it is everything running as expected @leo-ri

mrodm: Thanks for checking this out @leo-ri 
https://github.com/elastic/package-registry/pull/1057/commits/8a63e8c60a3eb5f720cb26d13f26d0828ca52689
https://buildkite.com/elastic/package-registry/builds/284#0189643b-cc85-44f0-bb89-d2f29fae56d8

PR #1058, opened 2023-07-20
Add spec.min and spec.max query parameter in search endpoint

### LINKED ISSUE #1050: Support search based on spec version
Add a new set of parameters to search packages based on their spec version:
* `spec.max`, maximum version supported by the client.
* `spec.min`, minimum version supported by the client.

If `spec.max` is defined, but `spec.min` and `kibana.version` are not, `spec.min` will be considered to be 3.0.

This will be used by Kibana to search packages on the range of spec version it supports.

mrodm:
Fixes #1050 

This PR adds two new filter parameters `spec.min` and `spec.max` for the search endpoint. These values will be used to filter packages according to their `format_version` value.

If no `spec.min` nor `kibana.version` are in the query, the default value for `spec.min` is going to be `3.0.0`.

## How to test

1. Update some test packages to build different packages with different version of `format_version`
    - Remember to update for each built package the version in manifest and changelog (e.g. 0.0.9) too
   ```diff
    $ git diff
   diff --git packages/elastic_package_registry/manifest.yml packages/elastic_package_registry/manifest.yml
   index b8a41f9b5..0d819ff0e 100644
   --- packages/elastic_package_registry/manifest.yml
   +++ packages/elastic_package_registry/manifest.yml
   @@ -1,4 +1,4 @@
   -format_version: 1.0.0
   +format_version: 2.10.0
    name: elastic_package_registry
    title: "Elastic Package Registry"
    version: 0.0.7
   ```
2. Build the packages
   ```shell
   elastic-package build
   ```
3. Build a new local docker image for package-registry
   ```shell
   cd /path/to/package-registry   
   ```
4. Run local package-registry using the packages directory with the built packages as a volume mount point:
   ```shell
   docker run --rm -p 8080:8080 -v /home/mariorodriguez/Coding/work/integrations/build/packages/:/packages/package-registry -it docker.elastic.co/package-registry/package-registry:v1.21.0-rc1
   ```
   ```shell
   # filter packages with spec.min and spec.max
   curl -s "http://localhost:8080/search?spec.min=2.2.0&spec.max=3.0.0&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'
   # filter packages with spec and kibana.version
   curl -s "http://localhost:8080/search?spec.min=2.2.0&spec.max=3.0.0&kibana.version=8.6.0&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'
   ```


packages/packages.go
```diff
@@ -412,12 +426,6 @@ func (f *Filter) legacyApply(ctx context.Context, packages Packages) Packages {
 			continue
 		}
 
-		if f.Capabilities != nil {
-			if valid := p.WorksWithCapabilities(f.Capabilities); !valid {
-				continue
-			}
-		}
-
```
mrodm, line 420: Removed from legacy. These clients should not send `capabilities` in their queries.

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Add new query parameter "capabilities" in search endpoint [#1054](https://github.com/elastic/package-registry/pull/1054)
+* Add new query parameter "capabilities" in search endpoint [#1054](https://github.com/elastic/package-registry/pull/1054) [#1058](https://github.com/elastic/package-registry/pull/1058)
```
mrodm, line 18: Would it be needed to add this PR into the entry related to capabilities ? It has been removed from the `legacyApply` (using `experimental=true`) here in this PR.

main_test.go
```diff
@@ -90,6 +90,10 @@ func TestEndpoints(t *testing.T) {
 		{"/search?type=input&prerelease=true", "/search", "search-input-packages.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=input&package=integration_input&prerelease=true", "/search", "search-input-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=integration&package=integration_input&prerelease=true", "/search", "search-integration-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=observability,security&prerelease=true", "/search", "search-prerelease-capabilities-observability-security.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=none&prerelease=true", "/search", "search-prerelease-capabilities-none.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?spec.min=1.1.0&spec.max=2.10.0&prerelease=true", "/search", "search-spec-min-max-2.10.0.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?spec.max=3.1.0&prerelease=true", "/search", "search-spec-default-min-max-3.1.0.json", searchHandler(testLogger, indexer, testCacheTime)},
```
mrodm, line 96: Added some tests for search endpoint using capabilities and spec parameters.

Some versions of the `example` and `foo` test packages have been updated:
- adding capabilities
- updating `format_version` number

packages/package.go
```diff
@@ -401,6 +401,32 @@ func (p *Package) WorksWithCapabilities(capabilities []string) bool {
 	return true
 }
 
+func (p *Package) HasCompatibleSpec(specMin, specMax, kibanaVersion *semver.Version) (bool, error) {
+	if specMin == nil && specMax == nil {
+		return true, nil
+	}
+	if specMin == nil && kibanaVersion == nil {
+		specMin = semver.MustParse("3.0.0")
+	}
```
mrodm, line 410: Not totally sure about this, this could be changed depending on what it is decided here https://github.com/elastic/package-registry/issues/1050#issuecomment-1643650408

main_test.go
```diff
@@ -90,6 +90,10 @@ func TestEndpoints(t *testing.T) {
 		{"/search?type=input&prerelease=true", "/search", "search-input-packages.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=input&package=integration_input&prerelease=true", "/search", "search-input-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=integration&package=integration_input&prerelease=true", "/search", "search-integration-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=observability,security&prerelease=true", "/search", "search-prerelease-capabilities-observability-security.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=none&prerelease=true", "/search", "search-prerelease-capabilities-none.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?spec.min=1.1.0&spec.max=2.10.0&prerelease=true", "/search", "search-spec-min-1.1.0-max-2.10.0.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?spec.max=2.10.0&prerelease=true", "/search", "search-spec-max-2.10.0.json", searchHandler(testLogger, indexer, testCacheTime)},
```
mrodm, line 96: Added some tests for search endpoint using capabilities and spec parameters.

Some versions of the `example` and `foo` test packages have been updated:
- adding capabilities
- updating `format_version` number

packages/package.go
mrodm: Removed this default minimum spec.
So in case there is no `spec.min` in the default query, it will return all the packages filtered just by `spec.max`

main_test.go
```diff
@@ -90,6 +90,10 @@ func TestEndpoints(t *testing.T) {
 		{"/search?type=input&prerelease=true", "/search", "search-input-packages.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=input&package=integration_input&prerelease=true", "/search", "search-input-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
 		{"/search?type=integration&package=integration_input&prerelease=true", "/search", "search-integration-integration-package.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=observability,security&prerelease=true", "/search", "search-prerelease-capabilities-observability-security.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?capabilities=none&prerelease=true", "/search", "search-prerelease-capabilities-none.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?spec.min=1.1.0&spec.max=2.10.0&prerelease=true", "/search", "search-spec-min-1.1.0-max-2.10.0.json", searchHandler(testLogger, indexer, testCacheTime)},
```
jsoriano, line 95: I see I didn't copy this to the description of the issue, sorry. In the original proposal we thought about using only <major>.<minor> (as equivalent to <major>.<minor>.0), so if a version of Kibana supports 2.0.0, it also supports 2.0.1 or any other 2.0.X version. Patch versions shouldn't include differences that break compatibility, or new features.

So the API would work like this:
```suggestion
		{"/search?spec.min=1.1&spec.max=2.10&prerelease=true", "/search", "search-spec-min-1.1.0-max-2.10.0.json", searchHandler(testLogger, indexer, testCacheTime)},
```

CHANGELOG.md
jsoriano: No need because this is reverting a change that was never released.

main_test.go
mrodm: Ok, it makes sense.

Updated the tests accordingly with that.

I also added a validation to check that versions in the query parameter are just  `major.minor` (added also a test for that).

search.go
```diff
@@ -132,6 +146,18 @@ func newSearchFilterFromQuery(query url.Values) (*packages.Filter, error) {
...
+	// version must cointain just <major.minor>
+	if len(strings.Split(version, ".")) != 2 {
+		return nil, fmt.Errorf("invalid version '%s': it should be <major.version>", version)
+	}
+	specVersion, err := semver.NewVersion(version)
+	if err != nil {
+		return nil, fmt.Errorf("invalid spec version '%s': %w", version, err)
+	}
+	return specVersion, nil
+}
```
mrodm, line 159: Ensure that versions in query parameters are as expected. That means with the pattern `major.minor`

packages/package.go
```diff
@@ -50,6 +50,9 @@ type Package struct {
 
 	versionSemVer *semver.Version
 
+	specSemVer           *semver.Version
```
jsoriano, line 53: Nit. Not used?

packages/package.go
```diff
@@ -328,6 +331,20 @@ func (p *Package) setRuntimeFields() error {
...
 		}
 	}
+
+	p.specSemVer, err = semver.StrictNewVersion(p.FormatVersion)
+	if err != nil {
+		return errors.Wrap(err, "invalid format spec version")
+	}
+
+	versionSplit := strings.Split(p.FormatVersion, ".")
+	versionSplit[2] = "0"
```
jsoriano, line 341: Nit. Check length to avoid panics?

jsoriano: Or maybe we can just build this version based on the previously parsed one.
```
specMajorMinorVersion := fmt.Sprintf("%d.%d.0", p.specSemVer.Major(), p.specSemVer.Minor())
```

jsoriano (approved):

packages/package.go
mrodm: Right, removed

packages/package.go
```diff
@@ -319,15 +321,33 @@ func (p *Package) setRuntimeFields() error {
...
 		if err != nil {
-			return errors.Wrapf(err, "invalid Kibana versions range: %s", p.Conditions.Kibana.Version)
+			return fmt.Errorf("invalid Kibana versions range %s: %w", p.Conditions.Kibana.Version, err)
 		}
 	}
+
+	// Packages from proxy mode do not have "format_version" field
+	if p.FormatVersion == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
jsoriano: But then `HasCompatibleSpec` won't work for proxied packages, right?

> And these proxied packages do not contain `format_version` in their struct/objects.

Why? Is it because we don't include `format_version` in search responses? Maybe we should add it now that it is meaningful for packages discoverability.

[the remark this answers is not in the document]
mrodm: > But then HasCompatibleSpec won't work for proxied packages, right?

I was thinking that as the query is sent to the remote (proxy) with the same parameters, the packages in the responses should be valid. 

> Why? Is it because we don't include format_version in search responses? Maybe we should add it now that it is meaningful for packages discoverability.

Exactly, currently `format_version` is not included in the responses.
It looks like including now this field in to the responses would be better. Should we added then in this PR ?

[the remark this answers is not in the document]
jsoriano: > I was thinking that as the query is sent to the remote (proxy) with the same parameters, the packages in the responses should be valid.

Good point, that should be enough :+1:

PR #1059, opened 2023-07-20
Add spec.min, spec.max to categories endpoint

mrodm:
This PR adds the query parameters `capabilities` as well as `spec.min` and `spec.max` in the categories endpoint as they were introduced for search endpoint in these issues:
- #1051 
- #1050 

Examples:
- http://localhost:8080/categories?capabilities=observability
- http://localhost:8080/categories?spec.min=2.1.0&spec.max=3.2.0

~This PR is a follow-up of #1058 . At the time of creating this PR was not merged, so this PR contains the same changes.~

## How to test

1. Update some test packages to build different packages with different version of `format_version`
    - Remember to update for each built package the version in manifest and changelog (e.g. 0.0.9) too
   ```diff
    $ git diff
   diff --git packages/elastic_package_registry/manifest.yml packages/elastic_package_registry/manifest.yml
   index b8a41f9b5..0d819ff0e 100644
   --- packages/elastic_package_registry/manifest.yml
   +++ packages/elastic_package_registry/manifest.yml
   @@ -1,4 +1,4 @@
   -format_version: 1.0.0
   +format_version: 2.10.0
    name: elastic_package_registry
    title: "Elastic Package Registry"
    version: 0.0.7
   @@ -9,6 +9,9 @@ categories:
    conditions:
      kibana.version: "^8.0.0"
      elastic.subscription: basic
   +  elastic.capabilities:
   +    - observability
   +    - security
    icons:
      - src: /img/elastic_package_registry-logo.svg
        title: Elastic Package Registry logo
   ```
2. Build the packages
   ```shell
   elastic-package build
   ```
3. Build a new local docker image for package-registry
   ```shell
   cd /path/to/package-registry   
   ```
4. Run local package-registry using the packages directory with the built packages as a volume mount point:
   ```shell
   docker run --rm -p 8080:8080 -v /home/mariorodriguez/Coding/work/integrations/build/packages/:/packages/package-registry -it docker.elastic.co/package-registry/package-registry:v1.21.0-rc1
   ```
   ```shell
   # filter categories with spec.min and spec.max
   curl -s "http://localhost:8080/categories?spec.min=2.2.0&spec.max=3.0.0&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'
   # filter categories with spec and kibana.version
   curl -s "http://localhost:8080/categories?spec.min=2.2.0&spec.max=3.0.0&kibana.version=8.6.0&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'

   # filter categories with capabilities
   curl -s "http://localhost:8080/categories?capabilities=observability,security&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'
    curl -s "http://localhost:8080/categories?capabilities=none&prerelease=true" | jq -r '.[]|"\(.name) - \(.version)"'
   ```


categories.go
```diff
@@ -127,6 +128,24 @@ func newCategoriesFilterFromQuery(query url.Values) (*packages.Filter, error) {
...
+		}
+	}
+
+	if v := query.Get("spec.max"); v != "" {
+		filter.SpecMax, err = semver.NewVersion(v)
+		if err != nil {
+			return nil, fmt.Errorf("invalid 'spec.max' version '%s': %w", v, err)
+		}
+	}
+
```
mrodm, line 148: Main changes included in this PR, adding these new filters to categories endpoint.

packages/packages_test.go
```diff
@@ -442,13 +443,234 @@ func TestPackagesFilter(t *testing.T) {
...
-			result := c.Filter.Apply(context.Background(), packages)
+			result, err := c.Filter.Apply(context.Background(), packages)
+			require.NoError(t, err)
+			assertFilterPackagesResult(t, c.Expected, result)
+		})
+	}
+}
+
+func TestPackagesSpecMinMaxFilter(t *testing.T) {
+	filterTestPackages := []filterTestPackage{
```
jsoriano, line 454: Could you add a test case with a patch version and check that it is included with spec.max set to its minor? For example a package with `FormatVersion: "2.2.1"`, and check that is included when `spec.max` is `2.2`.

packages/package.go
```diff
@@ -401,6 +401,29 @@ func (p *Package) WorksWithCapabilities(capabilities []string) bool {
...
+	}
+
+	fullConstraint := strings.Join(constraints, ",")
+	constraint, err := semver.NewConstraint(fullConstraint)
+	if err != nil {
+		return false, fmt.Errorf("cannot create constraint %s: %w", fullConstraint, err)
+	}
+
+	formatVersion := semver.MustParse(p.FormatVersion)
+	return constraint.Check(formatVersion), nil
```
jsoriano, line 424: We should probably set the patch number to 0, so we only compare with `<major>.<minor>`.

packages/package.go
```diff
@@ -401,6 +401,29 @@ func (p *Package) WorksWithCapabilities(capabilities []string) bool {
...
+		constraints = append(constraints, fmt.Sprintf("<=%s", specMax.String()))
+	}
+
+	fullConstraint := strings.Join(constraints, ",")
+	constraint, err := semver.NewConstraint(fullConstraint)
+	if err != nil {
+		return false, fmt.Errorf("cannot create constraint %s: %w", fullConstraint, err)
+	}
+
+	formatVersion := semver.MustParse(p.FormatVersion)
```
jsoriano, line 423: Nit. Maybe we should pre-parse these versions now that they are going to be more used.

jsoriano (commented):
Looks good, but not sure if this will work for patch versions, please add a test for this case.

packages/packages_test.go
mrodm: Good catch! Current implementation would not include a package with format_version 2.2.1
with the query spec.max 2.2

packages/package.go
mrodm: Updated in https://github.com/elastic/package-registry/pull/1058

packages/package.go
mrodm: Updated in https://github.com/elastic/package-registry/pull/1058

packages/packages_test.go
mrodm: Updated in https://github.com/elastic/package-registry/pull/1058

PR #1061, opened 2023-07-24
[Fleet] Use capabilities filter on categories endpoint

nchaulet:
## Description 

In order to get the correct count in Fleet UI we should probably filter on capabilities for `/categories` endpoint too.

That PR does that 

Test, the capabilities filter is already unit tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Add new query parameter "capabilities" in search endpoint [#1054](https://github.com/elastic/package-registry/pull/1054)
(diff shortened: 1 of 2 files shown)


mrodm (commented):
Hi @nchaulet , thanks for this PR.
Indeed it is needed to add these new query parameters into categories endpoint to filter the packages.

I started a PR a few days ago to include this query parameter (and spec.min/spec.max uery parameters) for the categories endpoint in this other PR https://github.com/elastic/package-registry/pull/1059.

CHANGELOG.md
```diff
@@ -34,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * New Security subcategory "Advanced Analytics (UEBA)" [#997](https://github.com/elastic/package-registry/pull/997)
-
+m
```
mrodm, line 38: Extra character here
```suggestion

```

mrodm (approved):

PR #1101, opened 2023-10-23
migrate package-registry-remote pipeline

sharbuz:
Triggering the package-storage-infra-update-package-registry pipeline instead of using the Jenkins Job


.buildkite/release-package-registry-distribution.yml
```diff
@@ -45,3 +45,17 @@ steps:
...
+    async: true
+    build:
+      branch: "main"
+      env:
+        DRY_RUN: "$DRY_RUN"
+        DRAFT_PR: "false"
+        EPR_VERSION: ""
+    depends_on:
+      - step: "release-distribution"
+        allow_failure: false
```
mrodm, line 61: This pipeline should not have this new step.

This step should be added into the main pipeline `.buildkite/pipeline.yml`
It needs to replace this step:
https://github.com/elastic/package-registry/blob/e074d1811cc5a4e72b7152b802897ef87474e287/.buildkite/pipeline.yml#L77-L91

Also adding the `if` to just run this pipeline when a tag is pushed

To test this, you could trigger a commit with this step ensuring that `DRY_RUN=true`and removing `if` temporarily to trigger this pipeline in a PR build.


As part of this PR, the next files (or folders) could be removed too:
- .buildkite/scripts/trigger-jenkins-job.sh
- .buildkite/scripts/triggerJenkinsJob

sharbuz: done

sharbuz:
@mrodm, There is the test result: https://buildkite.com/elastic/package-registry/builds/394, please take a look at it.

mrodm (approved):
> @mrodm, There is the test result: https://buildkite.com/elastic/package-registry/builds/394, please take a look at it.
LGTM! :+1: 

Thanks @sharbuz!

FYI @elastic/ecosystem , this step, that creates all the required Pull Requests after pushing a new tag, is going to be executed from now on in Buildkite.

PR #1102, opened 2023-10-24
Update Go runtime to 1.21.3

jsoriano:
Update Go runtime to latest 1.21, that includes some fixes in HTTP implementation.


mrodm (commented):
Missing to update `go.mod` with `go 1.21.0` and probably update dependencies in go.sum.

jsoriano:
`go.mod`s updated, as well as the setup in the release pipeline.

.buildkite/release-package-registry-distribution.yml
```diff
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
 env:
-  SETUP_GOLANG_VERSION: '1.20.4'
+  SETUP_GOLANG_VERSION: '1.21.3'
```
mrodm, line 4: This should be a leftover.

It should not be needed here. This environment variable is already defined in the `pre-command` hook: https://github.com/elastic/package-registry/blob/3e2cefe386757e5c371720e55230c05fd170c5a5/.buildkite/hooks/pre-command#L9


And I don't see that `golang` is installed neither in the scripts `validate-tag.sh` or `publish-distribution.sh` of this pipeline.

```suggestion
```

mrodm (approved):
LGTM 

It looks like there were some leftovers in the second Buildkite pipeline (`.buildkite/release-package-registry-distribution.yml`). IMO that var for the golang version (SETUP_GOLANG_VERSION) should not be needed

.buildkite/release-package-registry-distribution.yml
jsoriano: Removed! Thanks.

PR #1147, opened 2024-02-21
Reuse http client on storage resolver

jsoriano:
Reuses http client on storage resolves and sets a generous connect timeout to circuit-break on network issues.

Let's try to get memory profiles before applying this change.


mrodm (approved):
LGTM

mrodm (approved):

PR #1149, opened 2024-02-21
Update in place the package list

mrodm:
Update the package list retrieved from the index in place to avoid keeping twice the list of packages in memory.

In this new process, new packages are added at the end of the list.

Let's try to get memory profiles before applying this change.

Tasks before merging:
- [ ] Remove profiling
- [ ] Review benchmark update if needed or worthy


mrodm:
Added a new benchmark where in each iteration just one package is added (BenchmarkUpdate), but not found significant differences between `main` and this branch :( @jsoriano 

-old 
```
BenchmarkUpdate-16    	      20	1469164964 ns/op	571762132 B/op	 1888984 allocs/op
```
- new
```
BenchmarkUpdate-16    	      20	1466309455 ns/op	571728272 B/op	 1888972 allocs/op
```

Executed bench with this command:
```shell
go test -bench=. -benchtime=20x -benchmem ./storage/  |grep -v -E '(DEBUG|INFO|fakestorage.go|truncated)' |tee file.txt
```

PR #1152, opened 2024-02-23
Avoid propagating APM transaction from init to watcher

jsoriano:
Watcher receives a context when started from init, avoid propagating the APM transaction through this context.
Currently all spans from the watcher are reported as part of the init transaction.


mrodm (approved):
LGTM

PR #1170, opened 2024-04-03
Update Go runtime to 1.22.2

jsoriano:
This version includes fixes for `http.


mrodm:
buildkite test this

mrodm (approved):
:+1:

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This image contains the package-registry binary.
 # It expects packages to be mounted under /packages/package-registry or have a config file loaded into /package-registry/config.yml
 
-ARG GO_VERSION=1.21.7
+ARG GO_VERSION=1.22.7
```
mrodm, line 4: Wrong version

```suggestion
ARG GO_VERSION=1.22.2
```

Could this variable be also set by the script dynamically?

mrodm: Maybe something like this ? 

```shell
build_docker_image() {
    local go_version
    go_version=$(cat .go_version)

    docker buildx build "$@" \
        --platform linux/amd64,linux/arm64/v8 \
        -t "${DOCKER_IMG_TAG}" \
        -t "${DOCKER_IMG_TAG_BRANCH}" \
        --build-arg GO_VERSION="${go_version}" \
        --build-arg BUILDER_IMAGE=docker.elastic.co/wolfi/go \
        --build-arg RUNNER_IMAGE=docker.elastic.co/wolfi/chainguard-base \
        --label BRANCH_NAME="${TAG_NAME}" \
        --label GIT_SHA="${BUILDKITE_COMMIT}" \
        --label GO_VERSION="${SETUP_GOLANG_VERSION}" \
        --label TIMESTAMP="$(date +%Y-%m-%d_%H:%M)" \
        .
}
```

If this is applied, should we set GO_VERSION with an empty value in the Dockerfile ?

jsoriano: Ok, change applied.

> Wrong version

I was thinking that the failure was because the images hasn't been published yet :facepalm:

PR #1195, opened 2024-07-09
Add new label for EPR version/tag in Dockerfile

mrodm:
Add a new label in the docker image built to set specifically the Elastic Package Registry version (tag).


.buildkite/scripts/publish.sh
```diff
@@ -15,6 +15,7 @@ build_docker_image() {
         --build-arg BUILDER_IMAGE=docker.elastic.co/wolfi/go \
         --build-arg RUNNER_IMAGE=docker.elastic.co/wolfi/chainguard-base \
         --label BRANCH_NAME="${TAG_NAME}" \
```
mrodm, line 17: Should it be keep it since it was already in previous docker images ?

PR #1207, opened 2024-08-19
[CI] Add steps in pipeline to run unit tests with MacOS ARM

mrodm:
Part of https://github.com/elastic/elastic-package/issues/539

Add new steps in the CI pipeline to run unit tests in MacOS ARM.

### DIFF
--- a/.buildkite/scripts/pre-install-command.sh
+++ b/.buildkite/scripts/pre-install-command.sh
@@ -60,6 +60,7 @@ with_mage() {
 }
 
 with_go_junit_report() {
+    echo "--- Install go-junit-report"
     go install github.com/jstemmer/go-junit-report/v2@latest
 }
 
--- a/.buildkite/scripts/pre-install-command.sh
+++ b/.buildkite/scripts/pre-install-command.sh
@@ -55,7 +55,7 @@ with_mage() {
     fi
 
     echo "--- Install mage"
-    go install "github.com/magefile/mage@${SETUP_MAGE_VERSION}"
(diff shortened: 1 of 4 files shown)


.buildkite/scripts/pre-install-command.sh
```diff
@@ -12,16 +33,34 @@ add_bin_path(){
...
+    go version
+    which go
+    PATH="${PATH}:$(go env GOPATH)/bin"
+    export PATH
+}
+
+with_mage() {
+    if ! command -v go &> /dev/null ; then
+        with_go
+    fi
```
mrodm, line 55: There are steps that run in golang containres, that would not need to install go again.

.buildkite/scripts/pre-install-command.sh
```diff
@@ -12,16 +33,34 @@ add_bin_path(){
     export PATH="${WORKSPACE}/bin:${PATH}"
 }
 
-with_mage() {
+with_go() {
     create_bin_folder
-    retry 5 curl -sL -o "${WORKSPACE}/bin/mage.tar.gz" "https://github.com/magefile/mage/releases/download/v${SETUP_MAGE_VERSION}/mage_${SETUP_MAGE_VERSION}_Linux-64bit.tar.gz"
```
mrodm, line 17: It looks like there is no artifact for Darwin and ARM in 1.15.0, so I changed the approach to install `mage` using `go install` command.

jsoriano (approved):

jsoriano (approved):

PR #1220, opened 2024-09-10
Add discovery fields and tests for content packages

mrodm:
Part of https://github.com/elastic/package-spec/issues/351

Added tests related to content packages and the new discovery fields into the BasePackage manifest definition.

Tested also running locally Elastic Package Registry:

```shell
# use test package from package-spec repository
cd /path/to/package-spec/
cd test/packages/good_content
elastic-package build

docker build --build-arg GO_VERSION="$(cat .go-version)" --rm -t docker.elastic.co/package-registry/package-registry:test .
docker run --rm -it -p 8080:8080 -v /path/to/package-spec/build/packages:/packages/package-registry docker.elastic.co/package-registry/package-registry:main

curl "http://localhost:8080/search?type=content&all=true&prerelease=true"
curl "http://localhost:8080/search?type=content&package=good_content&prerelease=true"
curl "http://localhost:8080/search?package=good_content&prerelease=true"
```


jsoriano (approved):
:+1:

Issue #1229, opened 2024-09-19
Add support to index and search packages based on discovery features


mrodm:
Follow-up for https://github.com/elastic/package-spec/issues/351

Once Elastic Package Registry (EPR) has added support for content packages that contain discovery properties:
```yaml
discovery:
  fields:
    - name: process.pid
```
the package registry should also be able to index and search packages based on the information included in the discovery property of each package. 

For example the following request to EPR would return packages that can leverage documents that include the `process.pid` or the `user.id` fields.

```
GET /search?discovery=fields:process.pid,user.id
```

PR #1235, opened 2024-10-08
Implement content packages discovery

jsoriano:
Fixes https://github.com/elastic/package-registry/issues/1229.

Add a new `discovery` parameter to `/search` and `/categories` endpoints, to filter packages based on their discovery fields. It supports a list of fields that has to be passed in the form `fields:<comma-separated list of fields>`. To pass the filter, packages must include discovery fields in their manifest, and all of them must be included in the list included in the parameter.

For example the following request will return prerelease packages of content type, whose discovery subfields are a subset of ['http.method', 'apache.status.total_bytes'].
```
GET /search?prerelease=true&type=content&discovery=fields:http.method,apache.status.total_bytes
```


jsoriano:
@mrodm do you know how the [search-index-all-full.json](https://github.com/elastic/package-registry/blob/9c3021436f098baca86b3d986110a65df093937b/storage/testdata/search-index-all-full.json) can be regenerated?

I would like to add content packages there.

categories.go
```diff
@@ -219,7 +227,7 @@ func getCategoriesOutput(ctx context.Context, categories map[string]*packages.Ca
 	}
 	sort.Strings(keys)
 
-	var outputCategories []*packages.Category
+	outputCategories := []*packages.Category{}
```
jsoriano, line 230: Unrelated change. It changes the response of the `/categories` endpoint to generate `[]` instead of `null` when no packages satisfy the conditions.

mrodm:
> @mrodm do you know how the [search-index-all-full.json](https://github.com/elastic/package-registry/blob/9c3021436f098baca86b3d986110a65df093937b/storage/testdata/search-index-all-full.json) can be regenerated?
> 
> I would like to add content packages there.

The last time I had to add new packages (and update this file), I did it manually.
I don't know if there is any automated way to achieve this, sorry :(

packages/packages.go
```diff
@@ -297,11 +297,63 @@ type Filter struct {
...
 	SpecMax        *semver.Version
+	Discovery      *DiscoveryFilter
 
 	// Deprecated, release tags to be removed.
 	Experimental bool
 }
 
+type DiscoveryFilter struct {
+	Fields DiscoveryFilterFields
+}
```
mrodm, line 308: Would it be ok to set those two structs `DiscoveryFilter` and `DiscoveryFilterFields` as private ?

packages/packages.go
```diff
@@ -297,11 +297,63 @@ type Filter struct {
...
+		return nil, fmt.Errorf("unknown discovery filter %q", filterType)
+	}
+
+	return &result, nil
+}
+
+func (f *DiscoveryFilter) Matches(p *Package) bool {
+	if f == nil {
+		return true
+	}
```
mrodm, line 334: Just for checking, would this mean that there is no filter defined in the query parameter?
Or is it there just to avoid raising a panic error ?

packages/packages.go
```diff
@@ -297,11 +297,63 @@ type Filter struct {
...
 	// Deprecated, release tags to be removed.
 	Experimental bool
 }
 
+type DiscoveryFilter struct {
+	Fields DiscoveryFilterFields
+}
+
+func NewDiscoveryFilter(filter string) (*DiscoveryFilter, error) {
+	filterType, args, found := strings.Cut(filter, ":")
```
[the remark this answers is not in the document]
jsoriano: Currently we only have a type of discovery method, so I didn't add support for more.

If we had, a request using more than one would set the parameter multiple times, yes.

I can add support for this, though it wouldn't be really useful at the moment. What do you prefer?

packages/packages.go
jsoriano: Yes, this is code safety. This would mean that `Matches` is called with a `nil` value, for example [here](https://github.com/elastic/package-registry/pull/1235/files#diff-20fd7b81f7ff43064db94851e0fc5ad6a9852949b07d72a02d48751be912534fR407).

mrodm: Ok, thanks for the explanation!

packages/packages.go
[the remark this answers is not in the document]
jsoriano: Do you think I should add support for `discovery` appearing multiple times?

So these two queries would be equivalent:
```
GET /search?discovery=fields:process.pid&discovery=fields:user.id
```
```
GET /search?discovery=fields:process.pid,user.id
```

This would also serve as base to add other methods. Though I don't think this would be very useful by now.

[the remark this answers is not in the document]
mrodm: Probably, it is not required to do it now. I guess Kibana could put all the fields required in one query parameter, WDYT?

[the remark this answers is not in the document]
jsoriano: Yes, there would be no reason to separate in multiple parameters.

PR #1242, opened 2024-10-25
Update go runtime to 1.23.2

mrodm:
Updated to the latest version of Golang 1.23


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/elastic/package-registry
 
-go 1.22.1
+go 1.22.8
```
mrodm, line 3: Updated also to the latest Golang 1.22 version.

Not sure if this is required, since the version defined in the toolchain is going to be downloaded and used.

WDYT ?

.go-version
```diff
@@ -1 +1 @@
-1.23.1
+1.23.2
```
mrodm, line 1: Would it be better to set here the same as in the go.mod (go directive) 1.22.8?

jsoriano: This is the version used for building, right? I would use the latest one here.

go.mod
jsoriano: I would keep it as 1.22.1 if possible. This will be the minimum version required for projects importing this.

.go-version
mrodm: Exactly, this is going to be used in CI and also following the indications of the README to build the package-registry binary too: https://github.com/elastic/package-registry?tab=readme-ov-file#go-command

go.mod
mrodm: Ok! Reverting it.

jsoriano (approved):

Issue #1243, opened 2024-10-28
Expose `policy_template_behavior` in search and package endpoints


mrodm:
Relates https://github.com/elastic/package-spec/issues/802

Add as part of the API response in search and package endpoints, the `policy_templates_behavior` introduced in  https://github.com/elastic/package-spec/pull/825

Issue #1238, opened 2024-10-14
Expose `deployment_modes` in search and info endpoints


jen-huang:
`deployment_modes` was added in https://github.com/elastic/package-spec/pull/738 to support agentless packages. This property should be returned in the `/search` endpoint and the package info endpoint ([example](https://epr.elastic.co/package/elastic_connectors/0.0.2/)) so that Fleet can provide a better UX around agentless-only packages, as in https://github.com/elastic/kibana/pull/194885.

PR #1244, opened 2024-10-28
Expose policy_templates_behavior and deployment modes for policy templates in search and package endpoints

mrodm:
Closes https://github.com/elastic/package-registry/issues/1238
Closes https://github.com/elastic/package-registry/issues/1243

Expose `policy_templates_behavior` and the `deployment_modes` for the policy templates (if any) in both search and package endpoints.

## How to test

1. Compile and run Elastic Package Registry with the test packages (using `config.yml` from the repo):
   ```shell
   eval "$(gvm $(cat .go-version))"
   go run . -config config.yml
   ```
2. In another terminal, run these queries and ensure those new fields are present:
   ```shell
   curl -s "http://127.0.0.1:8080/search?package=deployment_modes&prerelease=true"
   curl -s "http://127.0.0.1:8080/package/deployment_modes/0.0.1/"
   ```


packages/package.go
```diff
@@ -184,6 +187,15 @@ type Download struct {
 	Type string `config:"type" json:"type" validate:"required"`
 }
 
+type DeploymentModes struct {
+	Default   *DeploymentMode `config:"default,omitempty" json:"default,omitempty" yaml:"default,omitempty"`
+	Agentless *DeploymentMode `config:"agentless,omitempty" json:"agentless,omitempty" yaml:"agentless,omitempty"`
```
mrodm, line 192: Not added the `organization`, `division` and `team` fields that can also be set under `agentless` key:

https://github.com/elastic/package-spec/blob/cc5282374a5d062497bf0517d77a449d7176a35e/spec/integration/manifest.spec.yml#L177-L197

Should they be added? Are they necessary for Fleet?

jsoriano: According to the description these fields are only intended to be used internally, when the package is installed: https://github.com/elastic/package-spec/issues/795
So I don't think we need to expose them here. cc @seanrathier just in case.

jsoriano (approved):

packages/package.go
seanrathier: They need to be added because Kibana will send those fields to the Agentless API and then the Agentless API will tag the agentless agent deployment with that information.

Does that answer the question?

mrodm: Looking at this comment, I thought those values would be read directly from the manifest in Kibana (Fleet):
https://github.com/elastic/package-spec/issues/795#issuecomment-2337993455

If the package is installed before sending those values to the Agentless API, could those be read from the package manifest? If that is possible, probably it's not needed to add those values to the API.

WDYT ? @seanrathier @jsoriano

seanrathier: If I am misunderstanding this request please let me know.

Are we looking to expose `deployment_modes` when searching for packages?  If so,  we don't need to expose the org, division and team in the search.

mrodm: > Are we looking to expose `deployment_modes` when searching for packages? If so, we don't need to expose the org, division and team in the search.

Exactly, I think I didn't give the full context here, sorry. This question was more related to check whether or not those fields should be added in to the search and package API responses from the package registry.

So, I think it's good to proceed as it is the PR now (just exposing if the mode is enabled or not).

Thanks @seanrathier !

PR #1260, opened 2024-12-26
Fix dependency problems with envoy package

jsoriano:
Dependabot is failing because of problems with a couple of dependencies. Exclude the affected versions and upgrade other packages.

Upgrade also the Go version.

See https://github.com/envoyproxy/go-control-plane/issues/1074


mrodm:
> Dependabot is failing because of problems with a couple of dependencies. Exclude the affected versions and upgrade other packages.
> 
> Upgrade also the Go version.
> 
> See [envoyproxy/go-control-plane#1074](https://github.com/envoyproxy/go-control-plane/issues/1074)

Looks like that this could have been resolved by https://github.com/envoyproxy/go-control-plane/pull/1079

Would it be possible to test without those `exclude` again?

jsoriano:
Ah yes, it looks like it was fixed in https://github.com/elastic/package-registry/pull/1261, closing.

Issue #1265, opened 2025-01-09
Add support for CORS headers


jsoriano:
Without CORS headers, responses from the Package Registry cannot be used in javascript applications running in browsers.

Given that all the Package Registry APIs are public and read-only, we don't really need restricting its cross origin access.

It would be probably fine to always allow any origin:
```
Access-Control-Allow-Origin: *
```

PR #1266, opened 2025-01-09
Add CORS headers to package registry

jsoriano:
Add `Access-Control-Allow-Origin: *` header to all responses.

Fixes https://github.com/elastic/package-registry/issues/1265.


kpollich (approved):
Thanks for jumping on this! Much appreciated ❤️

Issue #1277, opened 2025-02-18
Test packages have invalid `solution` type


jen-huang:
Various test packages declare `type: solution` which is not a valid package type according to the spec. This doesn't prevent the registry from loading up, but does prevent Kibana from parsing the registry response correctly.

These are the packages I found:
```
testdata/package/foo/1.0.0/manifest.yml
testdata/package/defaultrelease/0.0.1/manifest.yml
testdata/package/ilm_policy/1.0.0/manifest.yml
testdata/package/hidden/1.0.0/manifest.yml
testdata/package/elasticsearch_privileges/1.0.0/manifest.yml
testdata/package/agent_privileges/1.0.0/manifest.yml
testdata/package/input_level_templates/1.0.0/manifest.yml
```

jsoriano:
Most of the test packages there are invalid according to the spec (32 are invalid, only 2 are valid), probably due to a mix of legacy reasons and not needing that packages are valid from the Package Registry perspective.

> does prevent Kibana from parsing the registry response correctly.

For development or testing with Kibana we should probably use a better set of integrations, from the integrations repository or from EPR. Or maybe we could use the proxy mode with the public registry.

We can fix the type if this unblocks some dev workflow, but packages will remain invalid, I have opened a PR with this: https://github.com/elastic/package-registry/pull/1278

@jen-huang would this fix enough to unblock you?

Not sure though if we should fix all test packages as we want to make Package Registry resilient to invalid or legacy packages.

jen-huang:
> Not sure though if we should fix all test packages as we want to make Package Registry resilient to invalid or legacy packages.

This is a good point, as this issue certainly surfaced a real issue in Fleet: https://github.com/elastic/kibana/issues/211666

That has since been fixed and unblocks dev, so I am comfortable closing this one and keeping invalid packages around for the reasons you mention.

PR #1278, opened 2025-02-19
Fix types of test packages

jsoriano:
Related to https://github.com/elastic/package-registry/issues/1277

Use valid types in test packages to reduce problems when developing with Kibana.

Packages with wrong types also cause issues with `elastic-package`, that expects the manifest at the root of the package to have a valid type.

Most test packages here are actually invalid. This is in part intentional, as Package Registry is expected to be resilient to invalid or legacy packages.


jsoriano:
Issue closed, closing this too.

PR #1286, opened 2025-03-19
Prepare release 1.28.0

jsoriano:
Prepare release 1.28.0 with the new `siem` category, and take the opportunity to update the Go runtime.


.go-version
```diff
@@ -1 +1 @@
-1.23.4
+1.24.1
```
mrodm, line 1: I see that it was already updated in a previous PR (https://github.com/elastic/package-registry/pull/1285) the `toolchain` directive in `go.mod` :+1: 
https://github.com/elastic/package-registry/blob/07ca1502d2d426c844f4f8519890ad7f4f5d9710/go.mod#L5

mrodm (approved):

PR #1292, opened 2025-04-16
Add new benchmarks for storage indexer

mrodm:
This PR adds new benchmarks to help get data about how the storage indexer works.

It adds two different benchmarks:
- Related to updating indexer
    - It could show the data (e.g. memory) required to replace the current list of packages used by the indexer.
- Related to query the indexer all the packages
    - It could show the data (e.g. memory) required to build the answer for the search API request.

PR #1293, opened 2025-04-16
Use storage index package structs as pointers

mrodm:
This PR changes how the structs related to the search index and package manifests are used.
In this PR, they are updated to be pointers.

Using the benchmarks introduced in #1292, these are the stats obtained related to updating the index:

```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/storage
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                      │ ../package-registry-main/main_full_sample.txt │   pointers_full_sample.txt    │
                      │                    sec/op                     │   sec/op     vs base          │
IndexerUpdateIndex-16                                    292.6m ± 21%   301.4m ± 6%  ~ (p=0.165 n=10)

                      │ ../package-registry-main/main_full_sample.txt │      pointers_full_sample.txt       │
                      │                     B/op                      │     B/op      vs base               │
IndexerUpdateIndex-16                                    97.33Mi ± 0%   95.31Mi ± 0%  -2.08% (p=0.000 n=10)

                      │ ../package-registry-main/main_full_sample.txt │   pointers_full_sample.txt    │
                      │                   allocs/op                   │  allocs/op   vs base          │
IndexerUpdateIndex-16                                     473.2k ± 0%   473.2k ± 0%  ~ (p=0.927 n=10)
```

The more relevant data is the B/op values, there is a 2.08% improvement in that field.


storage/indexer.go
```diff
@@ -213,13 +209,13 @@ func (i *Indexer) Get(ctx context.Context, opts *packages.GetOptions) (packages.
 	return i.packageList, nil
 }
 
-func (i *Indexer) transformSearchIndexAllToPackages(sia searchIndexAll) (packages.Packages, error) {
+func (i *Indexer) transformSearchIndexAllToPackages(sia *searchIndexAll) packages.Packages {
 	var transformedPackages packages.Packages
 	for j := range sia.Packages {
 		m := sia.Packages[j].PackageManifest
```
mrodm, line 215: `PackageManifest` is a pointer. As this list comes from `loadSearchIndexAll`  function is supposed to not be nil.

Do you think it would  be needed to check whether or not that field is nil ?

storage/indexer.go
```diff
@@ -183,9 +183,13 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
 		metrics.StorageIndexerUpdateIndexErrorsTotal.Inc()
 		return fmt.Errorf("can't load the search-index-all index content: %w", err)
 	}
+	if anIndex == nil {
+		i.logger.Info("Downloaded new search-index-all index. No packages found.")
+		return nil
+	}
```
mrodm, line 189: Not sure if this is totally necessary. 
It looks like that `loadSearchIndexAll` function will not return a nil value
https://github.com/elastic/package-registry/blob/288bf96f2b7e520d333db91693fbb156395b8939/storage/index.go#L51-L91

If it is needed, taking into account that `err` is nil here, should it be returned an error instead in this scenario with no packages?

storage/indexer.go
```diff
@@ -213,13 +217,16 @@ func (i *Indexer) Get(ctx context.Context, opts *packages.GetOptions) (packages.
 	return i.packageList, nil
 }
 
-func (i *Indexer) transformSearchIndexAllToPackages(sia searchIndexAll) (packages.Packages, error) {
+func (i *Indexer) transformSearchIndexAllToPackages(sia *searchIndexAll) (packages.Packages, error) {
```
jsoriano, line 220: Maybe we can get rid of the `searchIndexAll` struct, see this patch over your changes: https://github.com/jsoriano/package-registry/commit/6875d6f1d335825f453859624dec1cecb04da323

jsoriano (approved):
LGTM, I wonder if we should take the opportunity to get rid of `searchIndexAll` struct too.

storage/indexer_test.go
```diff
@@ -34,14 +36,57 @@ func BenchmarkInit(b *testing.B) {
...
+	defer fs.Stop()
+	storageClient := fs.Client()
+
+	logger := util.NewTestLoggerLevel(zapcore.FatalLevel)
+	indexer := NewIndexer(logger, storageClient, FakeIndexerOptions)
+	err := indexer.Init(context.Background())
+	require.NoError(b, err)
+
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
```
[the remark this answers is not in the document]
mrodm: I was trying to simulate concurrent queries of getting all packages using `b.RunParallel`.
References:
- https://pkg.go.dev/testing#example-B.RunParallel
- https://konradreiche.com/blog/how-to-use-b-run-parallel/

storage/indexer_test.go
```diff
@@ -34,14 +36,57 @@ func BenchmarkInit(b *testing.B) {
...
+	defer fs.Stop()
+	storageClient := fs.Client()
+
+	logger := util.NewTestLoggerLevel(zapcore.FatalLevel)
+	indexer := NewIndexer(logger, storageClient, FakeIndexerOptions)
+	err := indexer.Init(context.Background())
+	require.NoError(b, err)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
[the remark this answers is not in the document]
mrodm: I've tried to use `b.Loop()`, but for the `BenchmarkIndexerUpdateIndex` benchmark, there is no output and the benchmark does not finish (it needs to be cancelled). I'll keep for now using the current implementation (`b.N`).

storage/indexer.go
mrodm: At least, returning early here will make sure that the package-registry service will not delete all the packages from its memory. That will ensure that the service could continue with the previous packages if it finds `anIndex == nil`
WDYT ?

jsoriano (approved):

Issue #1296, opened 2025-04-28
Add a supported EPR version to package schema and have the package-registry check this value before trying to load an integration


axrayn:
We ran into an issue recently where we updated all the packages (including the prerelease ones) onto our air-gapped EPR server, but made the mistake of not updating the version of the EPR server container itself.

This caused the container to crash because of the new 'siem' category that was introduced in a later version of the EPR server.

This was the log entry that we caught:
```json
{
    "log.level":"fatal",
    "@timestamp":"2025-04-27T22:54:15.870Z",
    "log.origin": {
        "function":"main.ensurePackagesAvailable",
        "file.name":"package-registry/main.go",
        "file.line":332
    },
    "message":"Init failed",
    "error": {
        "message":"reading packages from filesystem failed: loading package failed (path: /packages/package-registry/google_secops/0.2.0): invalid category: siem accessing config"
    },
    "ecs.version":"1.6.0"
}
```

As a simple solution, could we get this made into an ERROR-level log entry and skip indexing the errant package instead of a fatal error that stops the whole server loading?

As a more advanced solution, could we add an 'epr' section in the package manifest conditions section and then have the EPR check against this and either index or not index the package based on that check passing?
e.g.
```yaml
conditions:
  kibana:
    version: "^8.13.0 || ^9.0.0"
  elastic:
    subscription: basic
  epr:
    version: "^1.27.0"
```
If a package fails this check it could then be logged as a WARN/ERROR that it wasn't loaded because it required a newer version of EPR.

jsoriano:
Hey @axrayn, thanks for reporting. This failure should not be fatal, we will prepare a fix. In principle the registry should be quite tolerant to package formats.

axrayn:
Thanks for the quick fix!

PR #1297, opened 2025-04-28
Ignore unknown categories instead of producing fatal errors

jsoriano:
Fix https://github.com/elastic/package-registry/issues/1296

Unknown categories produce fatal errors now. This can be a problem when using new packages in old versions of the registry.

In general the Package Registry is quite tolerant to different package formats, so probably an unknown category should not prevent the registry from starting. Still, these categories should probably not being exposed in the registry APIs.

This change ignores unknown categories, and avoid producing fatal errors on these cases. Unknown categories are logged at the warning level.


packages/package.go
```diff
@@ -592,11 +598,17 @@ func (p *Package) Validate() error {
 		return fmt.Errorf("no description set")
 	}
 
-	for _, c := range p.Categories {
+	j := 0
+	for i, c := range p.Categories {
 		if _, ok := Categories[c]; !ok {
-			return fmt.Errorf("invalid category: %s", c)
+			p.logger.Warn("package uses an unknown category, will be ignored",
+				zap.String("category", c))
```
mrodm, line 605: Testing this branch locally, in this log message package name and package version are not shown:
```json
{"log.level":"warn","@timestamp":"2025-04-29T12:07:58.785+0200","log.origin":{"function":"github.com/elastic/package-registry/packages.(*Package).Validate","file.name":"packages/package.go","file.line":604},"message":"package uses an unknown category, will be ignored","category":"unknown","ecs.version":"1.6.0"}
```

I've tested this by running these commands:
```shell
mage build
./package-registry
```

As in [line 242](https://github.com/elastic/package-registry/pull/1297/files#diff-793186ea40ef241016210b03078529829f86c5f00367642830663928d7445126R242) a child logger is created including both fields, I don't know why those fields (package name and version) are not shown here.

As this `Validate` function is used via `Unpack`, could it be that the logger is not available there?

Should it be added the package name and version directly as attributes here ? 

```suggestion
			p.logger.Warn("package uses an unknown category, will be ignored",
				zap.String("package", p.Name),
				zap.String("version", p.Version),
				zap.String("category", c))
```

This would result in this log message (following the same commands as above):
```json
{"log.level":"warn","@timestamp":"2025-04-29T12:27:35.398+0200","log.origin":{"function":"github.com/elastic/package-registry/packages.(*Package).Validate","file.name":"packages/package.go","file.line":604},"message":"package uses an unknown category, will be ignored","package":"nodirentries","version":"1.0.0","category":"unknown","ecs.version":"1.6.0"}
```

packages/package_test.go
```diff
@@ -11,108 +11,112 @@ import (
...
+					},
+					Categories: []string{"custom", "foo"},
+					Version:    "1.2.3",
 				},
-				Categories: []string{"custom", "foo"},
+				FormatVersion: "1.0.0",
+				logger:        zap.Must(zap.NewDevelopment()),
 			},
+			true,
+			"unknown category ",
```
mrodm, line 68: Nit.
```suggestion
			"unknown category",
```

packages/package.go
jsoriano: Oh, I moved the tags from here to the constructor at some point, but yes, you are right, this is called during unpacking, so the settings are not there yet.
Adding the tags in both places now.

mrodm (approved):

PR #1319, opened 2025-06-23
Replace stdlib encoding/json by goccy/go-json

mrodm:
Replace "encoding/json" from stdlib with a fast JSON library "github.com/goccy/go-json".


jsoriano:
As discussed offline we could use this library in the less memory sensitive parts, but maybe it is safer to continue using the standard library.

PR #1334, opened 2025-07-04
Improvements to the SQL storage indexer for observability and reliability

mrodm:
Follows #1313 and #1326

This PR tries to run some improvements to the SQL Storage Indexer:
- Ensure that all Spans related to Update database are nested under the same parent span.
- Run cleanup database just if cursor has been updated in that iteration.
- Add the current cursor as a new column of the packages inserted into the database.
    - This could help in the scenario where there is an error dropping the table.
- Unify how JSONs are served in search, categories and packages endpoints.
    - Added new spans in those functions.
- Include options structs to set the variables of the SQL objects.
- Allow to customize the default batch sizes via environment variables:
    - batch size to read packages from index `EPR_SQL_INDEXER_READ_PACKAGES_BATCH_SIZE`.
    - batch size to insert packages into the database `EPR_SQL_INDEXER_DB_INSERT_BATCH_SIZE`.


internal/database/sqliterepository.go
```diff
@@ -159,19 +179,18 @@ func (r *SQLiteRepository) Initialize(ctx context.Context) error {
...
+	span.Context.SetLabel("database.path", r.File(ctx))
 	defer span.End()
 
 	if len(pkgs) == 0 {
 		return nil
 	}
 
 	totalProcessed := 0
-	args := make([]any, 0, r.maxBulkAddBatch*r.numberFields)
+	args := make([]any, 0, r.maxTotalArgs)
```
mrodm, line 191: Added a new field `maxTotalArgs` since this value is always constant.

internal/storage/sqlindexer.go
```diff
@@ -199,20 +206,23 @@ func (i *SQLIndexer) updateIndex(ctx context.Context) error {
 		metrics.StorageIndexerUpdateIndexDurationSeconds.Observe(time.Since(start).Seconds())
 	}()
 
-	defer func() {
+	defer func(initialCursor string) {
+		if initialCursor == i.cursor {
+			return
+		}
```
mrodm, line 212: Just clean the database (drop + initialize), if there has been a new index read (new cursor value).

internal/database/sqliterepository.go
```diff
@@ -240,6 +260,9 @@ func (r *SQLiteRepository) BulkAdd(ctx context.Context, database string, pkgs []
 		if totalProcessed >= len(pkgs) {
 			break
 		}
+
+		// reuse args slice
+		args = args[:0]
```
jsoriano, line 265: :+1:

internal/database/sqliterepository.go
```diff
@@ -370,16 +395,34 @@ type FilterOptions struct {
...
 
 func (o *SQLOptions) Where() (string, []any) {
-	if o == nil || o.Filter == nil {
+	if o == nil {
 		return "", nil
 	}
 	var sb strings.Builder
 	var args []any
+	// Always filter by cursor
+	if o.CurrentCursor != "" {
```
jsoriano, line 410: `CurrentCursor` is always going to contain a value in normal operation, right?

jsoriano (approved):

internal/database/sqliterepository.go
mrodm: Yes, that variable should always have value.
It has been updated the `SQLIndexer` constructor to set the `cursor` field as `init` ([link](https://github.com/elastic/package-registry/pull/1334/files#diff-3ce35671f69ba00f2b08c1c5e9eff09f9f1d87ac0e61b009461120297c3db99aR73)) to ensure it always has a value different from empty string:

```go
	indexer := &SQLIndexer{
		storageClient:         storageClient,
		options:               options,
		logger:                logger,
		database:              options.Database,
		swapDatabase:          options.SwapDatabase,
		label:                 fmt.Sprintf("storage-%s", options.PackageStorageEndpoint),
		readPackagesBatchSize: defaultReadPackagesBatchSize,
		cache:                 options.Cache,
		cursor:                "init",
	}
```

And the filter is always set with `i.cursor` ([link](https://github.com/elastic/package-registry/pull/1334/files#diff-3ce35671f69ba00f2b08c1c5e9eff09f9f1d87ac0e61b009461120297c3db99aR408-R410)):
```go
		options := &database.SQLOptions{
			CurrentCursor: i.cursor,
		}
```

PR #1382, opened 2025-08-07
Return error if requests contain unknown query parameters

### LINKED ISSUE #1373: Return error if there is an unknown query parameter set in requests
Currently, if an unknown query parameter is set in a request for EPR those parameters are skipped and it is returned a response taking into account the other query parameters.

As example:

```shell
 $ curl -s "https://epr.elastic.co/search?package=elastic_package_registry&prerelease=true&unknown=foo"
[
  {
    "name": "elastic_package_registry",
    "title": "Elastic Package Registry",
    "version": "0.3.1",
    ...
  }
]
```

This happens for all endpoints (`/search`, `/categories`, ...).

This issue is for returning a 400 error code (Bad Request) if there is any unknown query parameter used.

An environment variable (`EPR_ALLOW_UNKNOWN_QUERY_PARAMETERS`) or flag (`allow-unknown-query-parameters`) should be available to allow these unknown query parameters if it is required to keep the same previous behavior. By default, it should be disallowed.

mrodm:
Closes #1373 

This PR updates the handlers of all endpoints to fail if there is any unknown query parameter.

This can be disabled by setting the parameter `-allow-unknown-query-parameters` or the environment variable `EPR_ALLOW_UNKNOWN_QUERY_PARAMETERS`

## Author's Checklist
- [x] Add tests enabling and disabling the flag `-allow-unknown-query-parameters`.
- [x] Run manual tests to validate that works as expected EPR with and without the flag.

### Test disallowing unknown query parameters

Run in a terminal an instance of the EPR service with the default value (fail if there is any unknown query parameter):
```shell
mage build
./package-registry
```
Run some queries to validate that it fails if there is any unknown query parameter:
```shell
curl -s "http://localhost:8080/search?package=yamlpipeline&unknown=foo"
curl -s "http://localhost:8080/categories?prerelease=true&unknown=foo"
curl -s "http://localhost:8080/package/yamlpipeline/1.0.0/?unknown=foo"
curl -s "http://localhost:8080/health?unknown=foo"
curl -s "http://localhost:8080/?unknown=foo"

curl -s "http://localhost:8080/epr/yamlpipeline/yamlpipeline-1.0.0.zip?unknown=foo"
``` 

### Test disallowing unknown query parameters
Run an instance of the EPR service allowing unknown query parameters:
```shell
mage build
./package-registry -allow-unknown-query-parameters
```
Run the following queries:
 ```shell
curl -s "http://localhost:8080/search?package=yamlpipeline&unknown=foo"
curl -s "http://localhost:8080/categories?prerelease=true&unknown=foo"
curl -s "http://localhost:8080/package/yamlpipeline/1.0.0/?unknown=foo"
curl -s "http://localhost:8080/health?unknown=foo"
curl -s "http://localhost:8080/?unknown=foo"

curl -o file.zip -s "http://localhost:8080/epr/yamlpipeline/yamlpipeline-1.0.0.zip?unknown=foo"
unzip -l file.zip
rm file.zip
```

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -37,8 +37,8 @@ func newArtifactsHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Dur
 	if indexer == nil {
 		return nil, errors.New("indexer is required for artifacts handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for artifacts handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	a := &artifactsHandler{
--- a/categories.go
+++ b/categories.go
@@ -44,8 +44,8 @@ func newCategoriesHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Du
 	if indexer == nil {
 		return nil, errors.New("indexer is required for categories handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for categories handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	h := &categoriesHandler{
--- a/index.go
+++ b/index.go
@@ -27,8 +27,8 @@ type indexHandler struct {
 type indexOption func(*indexHandler)
 
 func newIndexHandler(cacheTime time.Duration, opts ...indexOption) (*indexHandler, error) {
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for search handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 	data := indexData{
 		ServiceName: serviceName,
--- a/search.go
+++ b/search.go
@@ -42,8 +42,8 @@ func newSearchHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Durati
 	if indexer == nil {
 		return nil, errors.New("indexer is required for search handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for search handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	h := &searchHandler{
--- a/favicon.go
+++ b/favicon.go
@@ -23,8 +23,8 @@ type faviconHandler struct {
 type faviconOption func(*faviconHandler)
 
 func newFaviconHandler(cacheTime time.Duration, opts ...faviconOption) (*faviconHandler, error) {
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for favicon handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	h := &faviconHandler{
--- a/package_index.go
+++ b/package_index.go
@@ -42,8 +42,8 @@ func newPackageIndexHandler(logger *zap.Logger, indexer Indexer, cacheTime time.
 	if indexer == nil {
 		return nil, errors.New("indexer is required for package index handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for package index handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	h := &packageIndexHandler{
--- a/signatures.go
+++ b/signatures.go
@@ -37,8 +37,8 @@ func newSignaturesHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Du
 	if indexer == nil {
 		return nil, errors.New("indexer is required for categories handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for categories handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	s := &signaturesHandler{
--- a/static.go
+++ b/static.go
@@ -41,8 +41,8 @@ func newStaticHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Durati
 	if indexer == nil {
 		return nil, errors.New("indexer is required for static handler")
 	}
-	if cacheTime == 0 {
-		return nil, errors.New("cache time must be set for static handler")
+	if cacheTime <= 0 {
+		return nil, errors.New("cache time must be greater than 0s")
 	}
 
 	s := &staticHandler{

### CODE REVIEW

**Comment by prodsecmachine** [comment]
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/651f0348-3125-4895-949e-0ba3fd77d19d?refs=8f7a3ebb0b995dac873458ec63d699f365a0ec5f&source=prChecksComment)

File: artifacts.go [hunk 1]
```diff
@@ -23,14 +22,26 @@ const artifactsRouterPath = "/epr/{packageName}/{packageName:[a-z0-9_]+}-{packag
 
 var errArtifactNotFound = errors.New("artifact not found")
 
-func artifactsHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Duration) func(w http.ResponseWriter, r *http.Request) {
-	return artifactsHandlerWithProxyMode(logger, indexer, proxymode.NoProxy(logger), cacheTime)
-}
-
-func artifactsHandlerWithProxyMode(logger *zap.Logger, indexer Indexer, proxyMode *proxymode.ProxyMode, cacheTime time.Duration) func(w http.ResponseWriter, r *http.Request) {
```
mrodm, line 30: Incuding this `handlerOptions` struct, I think that this `artifactsHandlerWithProxyMode` could be removed and just keep one `artifactsHandler`. WDYT ?

main_test.go
```diff
@@ -80,59 +104,63 @@ func TestEndpoints(t *testing.T) {
...
+		{"/search?prerelease=true&discovery=datasets:good_content.errors&discovery=fields:process.path", "/search", "search-discovery-multiple-no-match.txt", defaultSearchHandler},
 		{"/favicon.ico", "", "favicon.ico", faviconHandleFunc},
 
 		// Removed flags, kept to ensure that they don't break requests from old versions.
-		{"/search?internal=true", "/search", "search-package-internal.json", searchHandler(testLogger, indexer, testCacheTime)},
+		{"/search?internal=true", "/search", "search-package-internal.json", defaultSearchHandler},
+
+		// Test queries with unknown query parameters
+		{"/search?package=yamlpipeline&unknown=true", "/search", "search-unknown-query-parameter-error.txt", defaultSearchHandler},
+		{"/search?package=yamlpipeline&unknown=true", "/search", "search-allowed-unknown-query-parameter.json", allowUnknownQueryParamsSearchHandler},
```
mrodm, line 163: Added a few tests to ensure that works as expected depending on the value of `allowUnknownQueryParameters`

index.go
```diff
@@ -25,7 +25,16 @@ func indexHandler(cacheTime time.Duration) (func(w http.ResponseWriter, r *http.
...
+	if options.cacheTime == 0 {
+		return nil, errors.New("cache time must be set for index handler")
+	}
 	return func(w http.ResponseWriter, r *http.Request) {
-		serveJSONResponse(r.Context(), w, cacheTime, body)
+		// Return error if any query parameter is present
+		if !options.allowUnknownQueryParameters && len(r.URL.Query()) > 0 {
+			badRequest(w, "not supported query parameters")
+			return
+		}
```
mrodm, line 36: For endpoints that do not expect any query parameter, they have been updated to fail if there is any query parameter provided in the request.

WDYT?

main.go
```diff
@@ -615,4 +675,18 @@ func getRouter(logger *zap.Logger, config *Config, indexer Indexer, cache *expir
...
+				// Ready check, currently same as live check
+			default:
+				if !options.allowUnknownQueryParameters {
+					badRequest(w, fmt.Sprintf("unknown query parameter: %s", k))
+				}
+				return
+			}
+		}
+	}
+}
```
mrodm, line 692: For this handler, I've followed the description in the comment of this function. There it is stated that it can be used `?ready=true` even if there is no different logic currently for that.
Therefore, this endpoint just allows one query parameter (`ready`), any other parameter is going to return a 400 error code.

categories.go
```diff
@@ -27,18 +27,23 @@ import (
...
+	if options.proxyMode == nil {
+		logger.Warn("artifactsHandlerWithProxyMode called without proxy mode, defaulting to no proxy")
+		options.proxyMode = proxymode.NoProxy(logger)
+	}
+	if options.indexer == nil {
+		return nil, fmt.Errorf("indexer is required for categories handler")
+	}
+	if options.cacheTime == 0 {
+		return nil, fmt.Errorf("cache time must be set for categories handler")
+	}
```
mrodm, line 40: With this change of using `handlerOptions` struct, now each handler must check which options they require themselves and return error if there is anything unexpected.

In case of the proxy mode, if it is not set, I though to set the default proxy mode (NoProxy).

mrodm: As mentioned in https://github.com/elastic/package-registry/pull/1382#discussion_r2266846186, it has been updated `Enabled()` to check if struct pointer is nil and in that case it will be considered as disabled.

search.go
```diff
@@ -87,91 +91,104 @@ func searchHandlerWithProxyMode(logger *zap.Logger, indexer Indexer, proxyMode *
...
+				}
+				filter.Discovery = append(filter.Discovery, discovery)
+			}
+		case "internal":
+			// Parameter removed in https://github.com/elastic/package-registry/pull/765
+			// Keep it here to avoid breaking existing clients.
+		default:
+			if !allowUnknownQueryParameters {
+				return nil, fmt.Errorf("unknown query parameter: %q", key)
+			}
```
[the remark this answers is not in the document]
mrodm: I've applied the changes to use `http.Handler` instead of `http.HandlerFunc` in https://github.com/elastic/package-registry/pull/1382/commits/876c5602bd0226d4abf55b9222e85103a17e10d9

Currently, I'm not sure about how to add those middleware that you mentioned, but in any case this change should help us moving in that direction.

Moreover, it has been added for each handler just the options they require/need. The ones that are mandatory have been added as parameter in the constructors and the optional ones via "Options".

artifacts.go
mrodm: Related to this comment https://github.com/elastic/package-registry/pull/1382#discussion_r2266856353

With the latest change in https://github.com/elastic/package-registry/commit/876c5602bd0226d4abf55b9222e85103a17e10d9 , each handler can specify its own options. With that change, it also is noted which parameters are mandatory and which ones are optional.

artifacts.go
```diff
@@ -22,72 +22,101 @@ const artifactsRouterPath = "/epr/{packageName}/{packageName:[a-z0-9_]+}-{packag
...
+	return a, nil
+}
 
-		vars := mux.Vars(r)
-		packageName, ok := vars["packageName"]
-		if !ok {
-			badRequest(w, "missing package name")
-			return
-		}
+func ArtifactsWithProxy(pm *proxymode.ProxyMode) artifactsOption {
```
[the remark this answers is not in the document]
mrodm: > We are making all these options public, is this intended?

Oh, good catch! No, I didn't want to make these options public. Thanks!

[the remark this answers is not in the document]
mrodm: Applied change to be those methods private in https://github.com/elastic/package-registry/pull/1382/commits/18cd8062e507cf6c69f5f85cef5fbf1e40fe2cff

jsoriano (approved):
Change requested for the cache time checks, for the rest it LGTM.

PR #1393, opened 2025-08-13
Remove unknown query parameters check

mrodm:
Reopens https://github.com/elastic/package-registry/issues/1373

This PR removes the breaking change added that validates if requests contain or not unknown
query parameters n https://github.com/elastic/package-registry/pull/1382. With that change, those requests were failed with a 400 error. This PR allows again to have unknown query parameters (and they will be ignored as before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Fix filtering by category in search endpoint (SQL Storage Indexer). [#1394](https://github.com/elastic/package-registry/pull/1394)
 * Revert failing requests with 400 error code if there are unknown query parameters. [#1393](https://github.com/elastic/package-registry/pull/1393)
 
 ### Added
--- a/internal/database/model.go
+++ b/internal/database/model.go
@@ -13,7 +13,6 @@ type Package struct {
 	Prerelease      bool
 	KibanaVersion   string
 	Capabilities    string
-	Categories      string
 	DiscoveryFields string
 	Type            string
 	Path            string
--- a/internal/database/sqliterepository.go
+++ b/internal/database/sqliterepository.go
@@ -38,7 +38,6 @@ var keys = []keyDefinition{
 	{"release", "TEXT NOT NULL"},
 	{"prerelease", "INTEGER NOT NULL"},
 	{"kibanaVersion", "TEXT NOT NULL"},
-	{"categories", "TEXT NOT NULL"},
 	{"capabilities", "TEXT NOT NULL"},
 	{"discoveryFields", "TEXT NOT NULL"},
 	{"type", "TEXT NOT NULL"},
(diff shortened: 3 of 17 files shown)


kpollich (approved):

main.go
```diff
@@ -114,7 +112,6 @@ func init() {
 	// This flag is experimental and might be removed in the future or renamed
 	flag.BoolVar(&dryRun, "dry-run", false, "Runs a dry-run of the registry without starting the web service (experimental).")
 	flag.BoolVar(&packages.ValidationDisabled, "disable-package-validation", false, "Disable package content validation.")
-	flag.BoolVar(&allowUnknownQueryParameters, "allow-unknown-query-parameters", false, "Allow unknown query parameters in the request. If set to false, the server will return an error if any unknown query parameter is present in the request.")
```
jsoriano, line 117: Could we change this default to `true` instead of removing all the code?

jsoriano, line 117: Maybe we can also add an "allow list" of headers, that should be always accepted and ignored, to quickly accept headers in custom deployments, or common headers sent by clients.

jsoriano (approved):
I think we should enable this at least for `/search` and `/categories`, where unknown parameters are problematic in two vectors: clients don't have any feedback if they use invalid parameters, and less filters use to lead to more expensive requests in EPR. So we end up having more expensive requests that are also not useful.

This is also the behavior of Elasticsearch for example:
```
$ curl -v -u elastic:changeme localhost:9200/logs-*/_search?query=foo
*   Trying 127.0.0.1:9200...
* Connected to localhost (127.0.0.1) port 9200 (#0)
* Server auth using Basic with user 'elastic'
> GET /logs-*/_search?query=foo HTTP/1.1
> Host: localhost:9200
> Authorization: Basic ZWxhc3RpYzpjaGFuZ2VtZQ==
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< X-elastic-product: Elasticsearch
< content-type: application/json
< content-length: 269
< 
* Connection #0 to host localhost left intact
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"request [/logs-*/_search] contains unrecognized parameter: [query]"}],"type":"illegal_argument_exception","reason":"request [/logs-*/_search] contains unrecognized parameter: [query]"},"status":400}
```

But I am also fine with delaying this change if we think there are other risks, or we want to reconsider what endpoints should have this check.

mrodm:
> I think we should enable this at least for `/search` and `/categories`, where unknown parameters are problematic in two vectors: clients don't have any feedback if they use invalid parameters, and less filters use to lead to more expensive requests in EPR. So we end up having more expensive requests that are also not useful.

This approach looks good to me too, WDYT @kpollich ?

We can also set the default value to `true` for the feature as suggested here: https://github.com/elastic/package-registry/pull/1393#discussion_r2275855644

So, by default EPR will keep the same behavior (accepting unknown query parameters). That would help to avoid breaking changes now.
WDYT ? @jsoriano @kpollich 

I'll prepare another PR with this approach so we can check both and decide which one could help us better here.

kpollich:
Thanks @mrodm and @jsoriano - the approach described here looks sound to me. Aligning with the Elasticsearch behavior is a good argument, and limiting it to the two endpoints where the behavior is problematic makes sense to limit risk. Let's move forward with this in mind.

kpollich (approved):

PR #1395, opened 2025-08-14
Apply validation of unknown query parameters just to search and categories endpoints

mrodm:
Relates #1393
Relates #1373

This PR changes the default behavior to allow unknown query parameters. The default value for `allow-unknown-query-parameters` is now `true`.

Moreover, if the unknown query parameters are disallowed, this PR narrows the scope of this validation of disallowing the unknown query parameters to just the `/search` and `/categories` EPR endpoints.

### DIFF
--- a/categories.go
+++ b/categories.go
@@ -49,9 +49,10 @@ func newCategoriesHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Du
 	}
 
 	h := &categoriesHandler{
-		logger:    logger,
-		indexer:   indexer,
-		cacheTime: cacheTime,
+		logger:                      logger,
+		indexer:                     indexer,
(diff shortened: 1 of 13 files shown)


categories.go
```diff
@@ -49,9 +49,10 @@ func newCategoriesHandler(logger *zap.Logger, indexer Indexer, cacheTime time.Du
 	}
 
 	h := &categoriesHandler{
-		logger:    logger,
-		indexer:   indexer,
-		cacheTime: cacheTime,
+		logger:                      logger,
+		indexer:                     indexer,
+		cacheTime:                   cacheTime,
+		allowUnknownQueryParameters: true,
```
mrodm, line 55: Allowed by default in search and categories unknown query parameters

PR #1396, opened 2025-08-14
[SQL storage indexer] Fix responses from requests including capabilities query parameter

mrodm:
This PR fixes the search and categories endpoints to return the expected packages when the `capabilities` query parameter is set.

As the discovery filter is not used neither in the SQL storage indexer it has been removed too.

This PR also updates the tests:
- including the content and input packages for package storage related tests.
- including packages defining discovery fields in package storage related tests.
- defining the same set of tests (requests) in storage indexer (`package_storage_test.go`) as it is in `main_test.go` file
- remove the files from `internal/storage/testdata` and just use the ones from `storage/testdata`. Those files were the same.


internal/database/sqliterepository.go
```diff
@@ -38,8 +38,6 @@ var keys = []keyDefinition{
 	{"release", "TEXT NOT NULL"},
 	{"prerelease", "INTEGER NOT NULL"},
 	{"kibanaVersion", "TEXT NOT NULL"},
-	{"capabilities", "TEXT NOT NULL"},
-	{"discoveryFields", "TEXT NOT NULL"},
```
mrodm, line 42: `discoveryFields` is not used currently as a filter, and it could have the same issues as `categories` and `capabilities`.
So I think it is better to remove it from here for now.

internal/storage/sqlindexer_test.go
```diff
@@ -32,7 +32,7 @@ func TestSQLInit(t *testing.T) {
 	options, err := CreateFakeIndexerOptions(db, swapDb)
 	require.NoError(t, err)
 
-	fs := PrepareFakeServer(t, "testdata/search-index-all-full.json")
+	fs := PrepareFakeServer(t, "../../storage/testdata/search-index-all-full.json")
```
mrodm, line 35: Changed all tests to use the files from `storage/testdata` folder.
They were the same files, and it was needed to update both at the same time. Let's keep just one copy.

storage/testdata/search-index-all-full.json
```diff
@@ -462492,6 +462687,67 @@
      ]
     }
    },
+   {
+    "package_manifest": {
+     "name": "sql",
+     "title": "SQL Input",
```
mrodm, line 462693: Example of input package included.

storage/testdata/search-index-all-full.json
```diff
@@ -365805,6 +365940,66 @@
      ]
     }
    },
+   {
+    "package_manifest": {
+     "name": "kubernetes_otel",
+     "title": "Kubernetes OpenTelemetry Assets",
+     "version": "0.0.6",
```
mrodm, line 365947: Example of content package included.

jsoriano (approved):

PR #1411, opened 2025-08-28
Use testing context when possible

jsoriano:
Go 1.24.0 introduced a context in `testing` objects that is cancelled after the test is finished, but before registered cleanup functions are executed. See https://tip.golang.org/doc/go1.24#testingpkgtesting.

Apart from providing a non-nil context, this new context helps ensuring that resources that should be finished by context cancellation are actually finished.

In general there is no reason to don't use this context, the only reasons would be:
* Contexts in cleanup functions, as this context will be cancelled at this point.
* Contexts for resources that should persist along tests. Something I think we don't have in this project.

Indexer cleanups have been reorganized to use Cleanup instead of defer, so tests can be more easily parallelized.

### DIFF
--- a/categories_test.go
+++ b/categories_test.go
@@ -5,7 +5,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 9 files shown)


prodsecmachine:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/f2dfbd4d-e172-4580-bcae-926d0a7925d8?refs=080f7e79c4a707e98c2ab3e92803195a507df778&source=prChecksComment)

mrodm (approved):

Issue #1436, opened 2025-09-25
[ER] Make `serviceName` configurable via `ELASTIC_APM_SERVICE_NAME` environment variable


fred-maussion:
## Description

The current implementation uses a hardcoded constant for the APM `serviceName`:

```go
const serviceName = "package-registry"
```

This prevents dynamic configuration of the APM service name, which can be useful in:

- multi-instance deployments
- containerized environments
- observability use cases (e.g., grouping services by env or cluster)
- debugging or staging environments where custom service naming is required

## Expected behavior

The service name used in APM should:

- Default to `"package-registry"` if the environment variable `ELASTIC_APM_SERVICE_NAME` is not set
- Use the value from `ELASTIC_APM_SERVICE_NAME` if present

##  Proposed implementation

Replace the `const` with a top-level `var` initialized with the value of `os.Getenv`, falling back to the default.

### Code snippet:
```go
// Replace this:
const serviceName = "package-registry"

// With this:
var serviceName = getServiceName()

func getServiceName() string {
    if name := os.Getenv("ELASTIC_APM_SERVICE_NAME"); name != "" {
        return name
    }
    return "package-registry"
}
```

This ensures:
- The env variable is only read once
- The same value is consistently used across APM tracer and logger initialization
- Performance and maintainability are preserved

PR #1437, opened 2025-09-25
#1436 : Add serviceName as variable

fred-maussion:
## Summary

This PR makes the `serviceName` used by Elastic APM configurable via the `ELASTIC_APM_SERVICE_NAME` environment variable.

By default, the service name remains `"package-registry"`, but it can now be overridden at runtime, improving flexibility for deployments across environments (e.g., staging, prod, multi-cluster setups).

## Changes

- Replaced the `const serviceName = "package-registry"` with a `var` that reads from `os.Getenv("ELASTIC_APM_SERVICE_NAME")`
- Ensured the value is read once at startup and reused consistently across APM tracer and logger initialization

### DIFF
--- a/main.go
+++ b/main.go
@@ -47,7 +47,6 @@ import (
 )
 
 const (
-	serviceName         = "package-registry"
 	version             = "1.32.2"
 	defaultInstanceName = "localhost"
 )
@@ -83,6 +82,7 @@ var (
 
 	featureProxyMode bool
 	proxyTo          string
+	serviceName      = getServiceName()
 
 	defaultConfig = Config{
 		CacheTimeIndex:               10 * time.Second,
@@ -334,6 +334,13 @@ func getHostname() string {
 		return defaultInstanceName
 	}
 	return hostname
+}
+
+func getServiceName() string {
+	if name := os.Getenv("ELASTIC_APM_SERVICE_NAME"); name != "" {
+		return name
+	}
+	return "package-registry"
 }
 
 func initMetricsServer(logger *zap.Logger) {


fred-maussion:
Thanks to this change we can differentiate the different instance while keeping all the APM features working.
package-registry-servicename-variable

And gain access to distributed tracing in case of proxymode usage
image

main.go
```diff
@@ -336,6 +336,13 @@ func getHostname() string {
 	return hostname
 }
 
+func getServiceName() string {
+	if name := os.Getenv("ELASTIC_APM_SERVICE_NAME"); name != "" {
+		return name
+	}
```
mrodm, line 342: I thought this environment variable would be honored as it is mentioned by the documentation:
https://www.elastic.co/docs/reference/apm/agents/go/configuration#config-service-name

But it looks like that this environment variable gets overwritten by the value set in:

https://github.com/elastic/package-registry/blob/cf4c2b8f9fc588afcbf2602ddc122c2727f8f504/main.go#L504-L507

Please, could you add a comment indicating that this environment variable comes from the official docs of APM Go Agent, and it is required since the value is overwritten?

I'm wondering if the service name should just be changed at the APM level and keep `package-registry` as service name for logs.

Would that make sense? Or, should logs and APM data set the same service name?
WDYT @elastic/ecosystem ? 

What I thought is if it should be updated that variable just in the `initAPMTracer` function or not. Something like:
```golang
func initAPMTracer() *apm.Tracer {
	apm.DefaultTracer().Close()
	if _, found := os.LookupEnv("ELASTIC_APM_SERVER_URL"); !found {
		// Don't report anything if the Server URL hasn't been configured.
		return apm.DefaultTracer()
	}

	service := serviceName
	if value, found := os.LookupEnv("ELASTIC_APM_SERVICE_NAME"); found {
		service = value
	}

	tracer, err := apm.NewTracerOptions(apm.TracerOptions{
		ServiceName:    service,
		ServiceVersion: version,
	})
	if err != nil {
		log.Fatalf("Failed to initialize APM agent: %v", err)
	}
	return tracer
}
```

jsoriano: > it looks like that this environment variable gets overwritten by the value set in:

Perhaps we shouldn't set the service name in the tracer. It defaults to the name of the binary, that for production builds is going to be the same.

fred-maussion:
@jsoriano I agree that we should use `service.environment` and `service.node.name` for their intended purposes of differentiating environments and individual instances. My goal with this change is to provide flexibility for more advanced observability scenarios where the logical **service identity** is distinct from those concepts.

There are two primary use cases where this becomes necessary:

1.  **Enterprise Naming Conventions:** In large organizations, it's common practice to have service naming conventions that reflect business context (e.g., `ENTITY-LOCATION-SERVICE`). This logical name remains the same across all environments (dev, staging, prod), while the `service.environment` tag is used to filter between them. The diagram below shows how a cluster of nodes forms a single logical service. Allowing `ELASTIC_APM_SERVICE_NAME` to be set gives users the ability to match these internal conventions.

```mermaid
flowchart TD
    A[Kibana Entity 1] --> |service.name : ENTITY1-FRANCE-EPR| B(NGINX)
    B --> D[EPR Node 1]
    B --> E[EPR Node 2]
    B --> F[EPR Node 3]

    G[Kibana Entity 2] --> |service.name : ENTITY4-UK-EPR| H(NGINX)
    H --> I[EPR Node 4]
    H --> J[EPR Node 5]
    H --> K[EPR Node 6]
```

2.  **Service Map Visualization for Distributed Tracing:** This is the most important driver. The Service Map is one of the main reason for this change along with the pivot capacity. The APM Service Map primarily uses `service.name` to render distinct nodes. To get a clear visualization of a call from `ENTITY1-FRANCE-EPR` to `epr.elastic.co` or  `ENTITY2-UK-EPR` to `internal-epr.elastic.co` they must have different service names.

To answer your question, while you can filter the view by environment, this doesn't answer the above use case. If both services share the same name, the Service Map will only render a single service node, blurring the benefit of the view and making it impossible to distinguish the call flow between the logically separate services.

jsoriano (approved):

PR #1442, opened 2025-09-30
[SQL indexer] Add functions for filtering packages at database level

mrodm:
This PR adds new functions into SQL to be able to compare fields using semantic version:
- semver_compare_constraint
- semver_compare_ge
- semver_compare_le

These functions allow the indexer to retrieve less packages when query parameters like `kibana.version`, `spec.min` or `spec.max` are used.

- Benchmark:
```
 $ benchstat sql_prod_indexer_main.txt sql_prod_indexer_functions.txt 
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/internal/storage
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                 │ sql_prod_indexer_main.txt │   sql_prod_indexer_functions.txt   │
                 │          sec/op           │   sec/op    vs base                │
SQLIndexerGet-16                  2.018 ± 2%   1.607 ± 2%  -20.35% (p=0.000 n=10)

                 │ sql_prod_indexer_main.txt │    sql_prod_indexer_functions.txt    │
                 │           B/op            │     B/op      vs base                │
SQLIndexerGet-16                520.1Mi ± 0%   414.1Mi ± 0%  -20.37% (p=0.000 n=10)

                 │ sql_prod_indexer_main.txt │   sql_prod_indexer_functions.txt    │
                 │         allocs/op         │  allocs/op   vs base                │
SQLIndexerGet-16                 7.281M ± 0%   5.863M ± 0%  -19.48% (p=0.000 n=10)
```
- Comparison

| Query | Main (no functions) | Using new functions | 
| --- | --- | --- | 
| `/search?kibana.version=9.0.0` | 8898 | 1345 |
| `/search?spec.min=2.0&spec.max=3.4` | 8898 | 5733 |

## Author's checklist 

- [x] Run benchmark to analyze the benefits.
- [x] Compare number of packages retrieved.

## How to test this locally

Launch Package Registry service:
```shell
bash dev/launch_epr_service_storage_indexer.sh -s -i storage/testdata/search-index-all-full.json -p localhost:8081 -d
```

Run some queries against that service launched previously, some examples:
```shell
curl -s "http://localhost:8081/search?kibana.version=8.16.0"
curl -s "http://localhost:8081/search?kibana.version=7.17.0"

curl -s "http://localhost:8081/search?spec.min=2.10&spec.max=3.1"
```


internal/database/functions.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	secondVersion, ok := args[2].(string)
+	if !ok {
+		return nil, fmt.Errorf("secondVersion argument must be a string")
+	}
+
+	newArgs := []driver.Value{firstVersion, fmt.Sprintf("%s%s", operation, secondVersion)}
+
+	return semverCompareConstraint(ctx, newArgs)
```
jsoriano, line 73: Compiling constraints is relatively expensive, as we saw in some other profiling. 

Maybe it is better to have two different functions, `semver_compare_le` and `semver_compare_ge` for less than and greater than respectively, that would only require to compile the two versions, and use their methods.

jsoriano (commented):
Nice to see we can add functions in Go!

internal/database/functions.go
mrodm: I'll add those two new functions :+1: 
Even it is better since using constraints there are some caveats when using pre-release versions, since by default they are excluded:
https://github.com/Masterminds/semver?tab=readme-ov-file#checking-version-constraints

> When constraint checking is used for checks or validation it will follow a different set of rules that are common for ranges with tools like npm/js and Rust/Cargo. This includes considering pre-releases to be invalid if the ranges does not include one. If you want to have it include pre-releases a simple solution is to include -0 in your range.

mrodm:
> > Latest benchmark results obtained with the changes introduced in this PR:
> 
> Are these benchmarks compared with the previous version or with main? They look worse than the ones in the description, right?

> Are these benchmarks compared with the previous version or with main? They look worse than the ones in the description, right?

Those were compared with the contents in the main branch. It's right that they are a bit worse. I cannot tell you exactly why it could be, but these benchmarks as they are run in my machine, maybe some other program was interfering. At least, in general there are some benefits.

This is a new benchmark repeated locally in my machine for both main and this branch:
```
goos: linux
goarch: amd64
pkg: github.com/elastic/package-registry/internal/storage
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                 │ ../package-registry-sql-queries/sql_prod_get_main_3.txt │ sql_prod_get_compare_functions_3.txt │
                 │                         sec/op                          │    sec/op      vs base               │
SQLIndexerGet-16                                               2.313 ± 27%      2.148 ± 1%  -7.14% (p=0.023 n=10)

                 │ ../package-registry-sql-queries/sql_prod_get_main_3.txt │ sql_prod_get_compare_functions_3.txt │
                 │                          B/op                           │     B/op       vs base               │
SQLIndexerGet-16                                              437.8Mi ± 0%    413.0Mi ± 0%  -5.67% (p=0.000 n=10)

                 │ ../package-registry-sql-queries/sql_prod_get_main_3.txt │ sql_prod_get_compare_functions_3.txt │
                 │                        allocs/op                        │   allocs/op    vs base               │
SQLIndexerGet-16                                               5.873M ± 0%     5.694M ± 0%  -3.05% (p=0.000 n=10)
```

PR #1456, opened 2025-10-20
Typo for local packages

philippkahr:
The `$(` call always gave me CLI errors, removing it worked.


jsoriano (approved):
Thanks! I think these are leftovers from previous versions of the docs.

PR #1482, opened 2025-11-17
Allow empty package paths on startup

### LINKED ISSUE #1351: [ER] Allow Elastic Package Registry to start without any installed packages
## Description

Currently, the Elastic Package Registry fails to start if no package is present in the configured directory. This behavior makes it difficult to deploy the registry in scenarios where packages are not yet available at installation time, a common case in lightweight installations or automated provisioning pipelines.

## Use case

In environments where we build minimal Docker images or binaries for air-gapped or CI/CD-driven deployments, the registry should be allowed to start even if no packages are installed yet. This enables customers to bootstrap the infrastructure before loading packages through custom automation or manual uploads.

## Expected behavior

The service should start successfully and return an empty package list (or an appropriate response), rather than fail or exit. Once packages are added, the registry should dynamically reflect the changes without requiring a restart.

## Benefits

- Enables light installations and infrastructure-as-code deployments
- Improves developer experience and early-stage automation flows
- Reduces coupling between package availability and service lifecycle

## Proposed solution

- Allow `package-registry` to start with zero packages present
- Log a warning such as: _"No packages found at startup. The registry is running but no content is available yet."_

A new flag would be added, that when used, would start the local indexer with a watcher for the configured directories, that would update the index when packages are added or removed.

teresaromero:
Resolves #1351 

This PR adds a flags to the package-registry in order to allow empty package directory when the registry starts up.
When this is enabled by flag, the filesystem indexers are watching for changes at the given directories. The watcher uses fsnotify, so any change on the configured directories will trigger an event.
 
The flag also enables an env to activate this behavour.

## Testing

Tested manually creating an empty folder and add it to the config file. Running the registry without the flag should fail. 
Using the flag and using the environment variables both make the registry run with empty folder.

When packages are added to the directory, the registry packages list is updated and they are available at the indexer

### DIFF
--- a/packages/packages.go
+++ b/packages/packages.go
@@ -220,7 +220,8 @@ func (i *FileSystemIndexer) Init(ctx context.Context) (err error) {
 	}
 
 	if i.enablePathsWatcher {
-		go i.watchPackageFileSystem(ctx)
+		// removing current transaction as we are starting a new one at watcher
+		go i.watchPackageFileSystem(apm.ContextWithTransaction(ctx, nil))
 	}
 	return nil
 }


prodsecmachine:

| Status | Scanner |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/0a74a942-c04f-4785-b9c3-962c8a71db09?refs=624a848f069268f3eae5aea8dfa879a09112e42b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/0a74a942-c04f-4785-b9c3-962c8a71db09/license?refs=624a848f069268f3eae5aea8dfa879a09112e42b&source=prChecksComment) |

main.go
```diff
@@ -128,6 +131,9 @@ func init() {
 	// The following proxy-indexer related flags are technical preview and might be removed in the future or renamed
 	flag.BoolVar(&featureProxyMode, "feature-proxy-mode", false, "Enable proxy mode to include packages from other endpoint (technical preview).")
 	flag.StringVar(&proxyTo, "proxy-to", "https://epr.elastic.co/", "Proxy-to endpoint")
+
+	flag.BoolVar(&packagePathsAllowEmpty, "package-paths-allow-empty", false, "Allow empty package paths on startup")
```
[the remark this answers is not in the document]
jsoriano: I think this is fine as flags/environment variables.

main.go
```diff
@@ -128,6 +131,9 @@ func init() {
...
+	flag.DurationVar(&packagePathsWatchInterval, "package-paths-watch-interval", 1*time.Minute, "Watch interval for package paths")
```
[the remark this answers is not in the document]
teresaromero: answered here https://github.com/elastic/package-registry/pull/1482/files#r2536937949

teresaromero:
As i test the solution, i see the config.PackagePaths is not being validated, meaning i can use an empty config file so the filesystem indexers dont have any reference to watch. I guess before this change the "validation" was done because there was an empty list of packages and the init indexer failed.

I think i should include here a validation for the config PackagePaths item, if the we allow empty these should not be empty, if we dont allow empty it will fail as it has been doing as the program wont find packages at an empty dir. WDYT? @mrodm @jsoriano

packages/packages.go
```diff
@@ -193,9 +200,44 @@ func (i *FileSystemIndexer) Init(ctx context.Context) (err error) {
...
 
+func (i *FileSystemIndexer) watchPackageFileSystem(ctx context.Context) error {
+	if i.watchInterval == 0 {
+		i.logger.Debug(fmt.Sprintf("No watcher configured for %s indexer, packageList will not be updated", i.label))
+		return nil
+	}
+	i.logger.Debug(fmt.Sprintf("Watching %s indexer for changes", i.label))
+	ticker := time.NewTicker(i.watchInterval)
+	defer ticker.Stop()
+	for {
```
jsoriano: Have you considered using file system notifications instead of an active loop? https://github.com/fsnotify/fsnotify

With this we would only read the files if they have actually changed, and as a future change we could even update the list of packages instead of rebuilding it, as we would know what has been added, removed or modified.

Though any this could be added later too, we will need the option of the active loop in any case for situations where notifications are not available.

jsoriano:
> I think i should include here a validation for the config PackagePaths item, if the we allow empty these should not be empty, if we dont allow empty it will fail as it has been doing as the program wont find packages at an empty dir. WDYT? @mrodm @jsoriano

Yeah, sounds good to check at least that these directories exist.

packages/packages.go
```diff
@@ -231,6 +276,9 @@ func (i *FileSystemIndexer) getPackagesFromFileSystem(ctx context.Context) (Pack
 	span.Context.SetLabel("indexer", i.label)
 	defer span.End()
 
+	i.m.RLock()
+	defer i.m.RUnlock()
+
```
mrodm: Not sure this method requires a lock.
This method is used by Init() and by the new method to update the package list `watchPackageFileSystem`. IIUC this usage would be more related to a writer lock, since it builds the new package list.

However, I think it would be needed a writer lock inside the `watchPackageFileSystem` that it is already present to update the lists.

packages/packages.go
```diff
@@ -193,9 +200,44 @@ func (i *FileSystemIndexer) Init(ctx context.Context) (err error) {
...
+		func() {
+			newPackageList, err := i.getPackagesFromFileSystem(ctx)
+			if err != nil {
+				i.logger.Error("reading packages from filesystem failed", zap.Error(err))
+				return
+			}
+			i.m.Lock()
+			i.packageList = newPackageList
+			i.m.Unlock()
+		}()
```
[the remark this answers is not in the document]
jsoriano: `backend.fswatcher` looks good to me.

Just as food for thought, maybe we should stop using namespaced types, like `backend.init`, or `backend.watcher`. And use a single a type such as `backend` or `background`, and use other fields to identify the indexer or other metadata that may be relevant.

packages/packages.go
```diff
@@ -189,10 +202,90 @@ var ZipFileSystemBuilder = func(p *Package) (PackageFileSystem, error) {
...
+		return err
+	}
+
+	if i.enablePathsWatcher {
+		go i.watchPackageFileSystem(ctx)
+	}
+	return nil
+}
+
+func (i *FileSystemIndexer) watchPackageFileSystem(ctx context.Context) {
```
teresaromero: implemented the watcher using fsnotify and working as expected. I would need to validate the debounce time as if is too low the update can't read the manifest of the packages when reading, and if is too high the operation might take too much time where there are lots of packages.

offline discussion about this pr with @mrodm he shared that fsnotify might be not-compatible with certain OS systems. @jsoriano do you know if [this](https://github.com/fsnotify/fsnotify?tab=readme-ov-file#why-dont-notifications-work-with-nfs-smb-fuse-proc-or-sys) could be an issue? an option was to enable a flag/config to set the "mode" of watcher, implement notify but also the active loop? WDYT?

packages/packages.go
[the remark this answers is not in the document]
teresaromero: regarding the apm tracer, why do we use transaction instead of span? should this also include some label with the "type" of indexer? zip or fs?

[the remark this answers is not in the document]
mrodm: Each iteration of updating index should create a new transaction, and each transaction could have different spans within it.

Some example:
example transaction


There could be added labels too, we do that when loading the search index in the storage indexer:
https://github.com/elastic/package-registry/blob/4e41cdf194e90754814e4f4bdf8402a6d37149bd/internal/storage/index.go#L24-L26

Or in the file system indexer in the `Get` operation:
https://github.com/elastic/package-registry/blob/4e41cdf194e90754814e4f4bdf8402a6d37149bd/packages/packages.go#L210-L212

Here it looks like that it could be added a label similar to that `Get` operation.

packages/packages.go
jsoriano: > offline discussion about this pr with @mrodm he shared that fsnotify might be not-compatible with certain OS systems. @jsoriano do you know if [this](https://github.com/fsnotify/fsnotify?tab=readme-ov-file#why-dont-notifications-work-with-nfs-smb-fuse-proc-or-sys) could be an issue? an option was to enable a flag/config to set the "mode" of watcher, implement notify but also the active loop? WDYT?

I don't think this will be an issue with usual local filesystems, but it would be interesting to be able to use the active loop as a escape hatch. This could be added later too.

packages/packages.go
```diff
@@ -189,10 +206,98 @@ var ZipFileSystemBuilder = func(p *Package) (PackageFileSystem, error) {
...
+func (i *FileSystemIndexer) updatePackageFileSystemIndex(ctx context.Context) error {
+	i.m.Lock()
+	defer i.m.Unlock()
+
+	newPackageList, err := i.getPackagesFromFileSystem(ctx)
 	if err != nil {
-		return fmt.Errorf("reading packages from filesystem failed: %w", err)
+		i.logger.Error("reading packages from filesystem failed",
+			zap.Error(err),
+			zap.String("indexer", i.label))
```
jsoriano: Redundant given that the caller is also going to receive the error and log it.
```suggestion
```

jsoriano (commented):
Looks good overall.

packages/packages_test.go
```diff
@@ -779,6 +788,247 @@ func TestPackagesSpecMinMaxFilter(t *testing.T) {
...
+	zipFilepath := filepath.Join(dest, pkgName+"-1.0.0.zip")
+	file, err := os.Create(zipFilepath)
+	require.NoError(t, err)
+
+	err = archiver.ArchivePackage(file, archiver.PackageProperties{
+		Name:    pkgName,
+		Version: "1.0.0",
+		Path:    filepath.Join(tmpMockFSDir, pkgName, "1.0.0"),
+	})
+	require.NoError(t, err)
```
jsoriano: The problem with tests in Windows might come from this function. It is creating the headers with native paths:
```go
                header, err := buildArchiveHeader(info, filepath.Join(rootDir, relativePath))
                if err != nil {
                        return fmt.Errorf("building archive header failed (path: %s): %w", relativePath, err)
                }
...
func buildArchiveHeader(info os.FileInfo, relativePath string) (*zip.FileHeader, error) {
        header, err := zip.FileInfoHeader(info)
        if err != nil {
                return nil, fmt.Errorf("reading file info header failed (info: %s): %w", info.Name(), err)
        }

        header.Method = zip.Deflate
        header.Name = relativePath
        if info.IsDir() && !strings.HasSuffix(header.Name, "/") {
                header.Name = header.Name + "/"
        }
        return header, nil
}

```

But the `Name` in these headers should use forward slashes https://pkg.go.dev/archive/zip#FileHeader

You can try with:
```go
        header.Name = filepath.ToSlash(relativePath)
```

jsoriano (approved):

packages/packages.go
```diff
@@ -196,46 +212,104 @@ var ZipFileSystemBuilder = func(p *Package) (PackageFileSystem, error) {
...
+	}
 	return nil
 }
 
-func (i *FileSystemIndexer) watchPackageFileSystem(ctx context.Context) error {
-	if i.watchInterval == 0 {
-		i.logger.Debug(fmt.Sprintf("No watcher configured for %s indexer, packageList will not be updated", i.label))
-		return nil
+func (i *FileSystemIndexer) watchPackageFileSystem(ctx context.Context) {
+	// TODO: https://github.com/elastic/package-registry/issues/1488
```
mrodm: :+1:

packages/packages.go
```diff
@@ -189,10 +212,103 @@ var ZipFileSystemBuilder = func(p *Package) (PackageFileSystem, error) {
...
-	i.packageList, err = i.getPackagesFromFileSystem(ctx)
+	if err := i.updatePackageFileSystemIndex(ctx); err != nil {
+		i.logger.Error("initializing package filesystem index failed",
+			zap.Error(err),
+			zap.String("indexer", i.label))
+		return err
+	}
+
+	if i.enablePathsWatcher {
+		go i.watchPackageFileSystem(ctx)
```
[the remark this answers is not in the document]
teresaromero: looking into the call hierarchy i see there is a transaction initialized at `initIndexer` that is passed along the ctx object...  so short answer is yes.

i've asked Copilot about it and it proposed me to use `apm.DetachedContext(ctx)` instead, giving me the following explanation, but from my pov this is not what we want, as we want to remove previous trace reference and init a new one inside the go routine. Also, the goroutine should be canceled if parent ctx is canceled.
WDYT?


> The key differences between [apm.DetachedContext](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [apm.ContextWithTransaction(ctx, nil)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):
> 
> [apm.DetachedContext(ctx)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
> Creates a new background context that is independent of the original context's lifecycle
> Preserves APM metadata from the original context (like trace/span IDs) while detaching from cancellation signals
> Ideal for goroutines that should continue after parent completes - perfect for your go i.watchPackageFileSystem() use case
> The goroutine won't be cancelled when [ctx](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is cancelled
> Visual representation:
> 
> [apm.ContextWithTransaction(ctx, nil)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
> Associates a transaction with a context (passing nil means no transaction)
> Maintains parent context's lifecycle - if [ctx](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is cancelled, the returned context is also cancelled
> Used to explicitly set/override transaction tracking
> Passing nil essentially returns a context with no active transaction
> Visual representation:
> 
> For Your Use Case
> Using [apm.DetachedContext(ctx)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is correct here because:
> 
> The watcher goroutine is long-running and should survive beyond the initiating request
> You want to maintain APM tracing links without inheriting cancellation
> The watcher has its own lifecycle management
> If you used ContextWithTransaction(ctx, nil), the watcher would be cancelled prematurely when the parent context completes.

mrodm (approved):

PR #1503, opened 2025-12-16
specify non-root user in Dockerfile

pkoutsovasilis:
## Summary

Specify a non-root user (UID 1000) in the Dockerfile to improve security and compatibility with Kubernetes security contexts.

## Motivation

The current Dockerfile runs as root (UID 0) by default, which:
- Conflicts with `RunAsNonRoot: true` security contexts in Kubernetes
- Requires explicit `runAsUser` overrides in pod specifications
- Is generally discouraged as a security best practice

## Changes

- Added `USER 1000` directive before the ENTRYPOINT to run the package-registry process as a non-root user

## Benefits

- **Kubernetes**: Eliminates the need to explicitly set `runAsUser` in pod specs when using `RunAsNonRoot: true`
- **OpenShift**: Works seamlessly with the default `restricted` SCC, which automatically remaps the UID to the namespace's allocated range
- **Security**: Follows the principle of least privilege by not running as root

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+* Package registry container image runs by default as a non-root user (UID 1000). [#1503](https://github.com/elastic/package-registry/pull/1503)
+
 ### Bugfixes
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ docker run --rm -it -p 8080:8080 \
   docker.elastic.co/package-registry/package-registry:main
 ```
 
+> [!IMPORTANT]
+> When you mount local package directories into the container (e.g., `-v /path/to/packages:/packages/package-registry`), ensure the mounted directory has proper filesystem permissions allowing UID 1000 to read the files. Alternatively, you can run the container as root using `docker run --user 0 ...` if you encounter permission issues.
+
 **Listening on HTTPS**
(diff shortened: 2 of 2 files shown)


jsoriano:
> > Could this be breaking if the user is running the registry with local packages? https://github.com/elastic/package-registry?tab=readme-ov-file#docker
> 
> how do you mean that @jsoriano ? mounting them directly in the container?

If executed with something like this:
```
docker run --rm -it -p 8080:8080 \
  -v /path/to/local/packages:/packages/package-registry \
  docker.elastic.co/package-registry/package-registry:main
```

Could it become an issue if `/path/to/local/packages` is not readable by all users?

pkoutsovasilis:
thanks for the confirmation @jsoriano , updated both CHANGELOG and README here 3698c8684ad333ac0019b2a51b9a91996aea2db7

Dockerfile
```diff
@@ -39,6 +39,9 @@ WORKDIR /package-registry
 # Get in config which expects packages in /packages
 COPY config.docker.yml /package-registry/config.yml
 
+# Run as non-root user
+USER 1000
```
jsoriano: I see some recommendations on setting UIDs > 10000 to avoid matching ids of local users. For example Chainguard uses UID 65532.

Do you think we could use a different UID here?

pkoutsovasilis: > Well, regarding to obscurity I don't see it any different to setting it to 1000 🙂

that's why I said still valuable. That said, I think we both agree that the original image running with UID 0 isn't secure

PR #1585, opened 2026-02-19
Sort packages by semver

### LINKED ISSUE #1584: Search API: versions sorted lexicographically instead of by semver when all=true
**Describe the bug**

When querying the search endpoint with `all=true` (to retrieve all versions of a package), versions within the same package are sorted lexicographically rather than by semantic version order. This causes versions with multi-digit patch/minor components to appear out of order.

**To reproduce**

```
GET /search?package=security_detection_engine&all=true
```

Observe that the results include, for example:

```
{ "version": "8.19.15" }  ← index 195
{ "version": "8.19.2" }   ← index 196
{ "version": "8.19.3" }   ← index 197
...
{ "version": "8.2.1" }    ← index 204
```

`8.19.15` appears before `8.19.2` because `"1" < "2"` lexicographically.

**Expected behavior**

Versions should be sorted in ascending semantic version order, for the same package:

```
1.0.1, ..., 8.2.1, ..., 8.19.2, 8.19.3, ..., 8.19.15
```

**Root cause**

`Packages.Less` in packages.go used a plain string comparison as the tiebreaker when two packages share the same title:

```go
return p[i].Version < p[j].Version
```

It should use `semver` comparison, as `byNameVersion.Less` in indexer.go already does correctly.

**Impact**

Any consumer iterating the `all=true` results and relying on version order (e.g. to find the newest compatible version) will get incorrect results.

HeroGamers:
This PR fixes #1584 
- After performing the existing package title check, packages are sorted by semver.

Testing
- Added tests to check correct sorting of packages.

### DIFF
--- a/testdata/generated/storage-indexer/search-all.json
+++ b/testdata/generated/storage-indexer/search-all.json
@@ -506,6 +506,913 @@
   {
     "name": "aws",
     "title": "AWS",
+    "version": "0.5.0",
+    "release": "beta",
+    "description": "AWS Integration",
+    "type": "integration",
+    "download": "/epr/aws/aws-0.5.0.zip",
+    "path": "/package/aws/0.5.0",
+    "icons": [
+      {
+        "src": "/img/logo_aws.svg",
+        "path": "/package/aws/0.5.0/img/logo_aws.svg",
+        "title": "logo aws",
+        "size": "32x32",
+        "type": "image/svg+xml"
+      }
+    ],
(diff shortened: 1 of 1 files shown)


HeroGamers:
@mrodm 
> Please, add also a changelog entry for the changes in this PR here:

Done :)

mrodm (approved):
Thanks @HeroGamers !

PR #1689, opened 2026-04-27
Fix package sort logic to use a more stable property name instead of title

nchaulet:
Resolve https://github.com/elastic/package-registry/issues/1601 

## Description

  - Sort packages by Name instead of Title for a more stable and deterministic sort order
  - Test fixtures and generated golden files updated to reflect the new ordering

##  Test plan

  - Unit tests updated in packages/packages_test.go — sort by name alphabetically, by semver within the same name, and combined name+version ordering
  - Golden file testdata/generated/search-all-proxy.json regenerated to match new sort order
  - Existing test suite passes

### DIFF
--- a/packages/packages.go
+++ b/packages/packages.go
@@ -41,8 +41,8 @@ type Packages []*Package
 func (p Packages) Len() int      { return len(p) }
 func (p Packages) Swap(i, j int) { p[i], p[j] = p[j], p[i] }
 func (p Packages) Less(i, j int) bool {
-	if p[i].Title != nil && p[j].Title != nil && *p[i].Title != *p[j].Title {
-		return *p[i].Title < *p[j].Title
+	if p[i].Name != p[j].Name {
+		return p[i].Name < p[j].Name
 	}
 	iSemVer, iErr := semver.NewVersion(p[i].Version)
 	jSemVer, jErr := semver.NewVersion(p[j].Version)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bugfixes
 
+* Sort packages by name instead of title for a more stable and deterministic sort order. [#1689](https://github.com/elastic/package-registry/pull/1689)
+
 ### Added
 
 ### Deprecated
--- a/packages/packages_test.go
+++ b/packages/packages_test.go
@@ -1310,21 +1310,19 @@ func assertFilterPackagesResult(t *testing.T, expected []filterTestPackage, foun
 func strPtr(s string) *string { return &s }
 
 func TestPackagesSort(t *testing.T) {
-	t.Run("sorts by title alphabetically", func(t *testing.T) {
+	t.Run("sorts by name alphabetically", func(t *testing.T) {
 		pkgs := Packages{
-			{BasePackage: BasePackage{Name: "c_pkg", Title: strPtr("Zebra Integration"), Version: "1.0.0"}},
-			{BasePackage: BasePackage{Name: "a_pkg", Title: strPtr("Apple Integration"), Version: "1.0.0"}},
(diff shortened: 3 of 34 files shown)


prodsecmachine:

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/599b3f95-f3ec-4e49-947a-7e956c2244e5?refs=d5b7a73f8212fb2de266086b5cc4a04360e00f7c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/599b3f95-f3ec-4e49-947a-7e956c2244e5/license?refs=d5b7a73f8212fb2de266086b5cc4a04360e00f7c&source=prChecksComment) |

packages/packages.go
```diff
@@ -41,8 +41,8 @@ type Packages []*Package
 func (p Packages) Len() int      { return len(p) }
 func (p Packages) Swap(i, j int) { p[i], p[j] = p[j], p[i] }
 func (p Packages) Less(i, j int) bool {
```
mrodm, line 43: If we change this ordering to be first based on name instead of title, could this other ordering methods be removed too ?

https://github.com/elastic/package-registry/blob/f6fd41ded5492ecdc7e97ceb8c232ba12c2f0df6/indexer.go#L82-L98

They look the same now.

mrodm (commented):
I'd like also to get the opinion from someone else from @elastic/ecosystem in case I'm missing any other side-effect.

packages/packages.go
nchaulet: Looks like the sort in indexer.go sort by newer version first, and we sort in packages by older first, but we probably could refacto to share some code

nchaulet: Actually tried to do the refacto not sure it's worth it

teresaromero: the order change was introduced [here](https://github.com/elastic/package-registry/pull/1585)

>and we sort in packages by older first

@nchaulet do you mean at fleet? maybe the sorting has to be aligned but we could do it on other pr?

mrodm: > Looks like the sort in indexer.go sort by newer version first, and we sort in packages by older first

Oh! I didn't realize that the ordering for the semantic version was different in those two, thanks for the heads up!

This sort method (the one from `[‎packages/packages.go](https://github.com/elastic/package-registry/pull/1689/changes#diff-20fd7b81f7ff43064db94851e0fc5ad6a9852949b07d72a02d48751be912534f)`) is related to queries that include `all=true` like (older versions first):
- https://epr.elastic.co/search?all=true
- https://epr.elastic.co/search?all=true&package=elastic_package_registry

Changing now that ordering for those requests, could that be interpreted as a breaking change?
Not sure if it should be done. WDYT @jsoriano @teresaromero  ? Should we keep for now the same ordering here ? Between older first or newer first, what would it the best for the users using this API ?

nchaulet: On the breaking change question: I think we should not change the version ordering for all=true queries in this PR and focus on fixing the bug with package title.

jsoriano: > Should we keep for now the same ordering here ? Between older first or newer first, what would it the best for the users using this API ?

I'd say we should be consistent and use the same ordering in all endpoints. Though this doesn't need to be fixed in this PR.

jsoriano (approved):

mrodm (approved):

teresaromero (approved):
can we create an issue to address de followup?

packages/packages.go
nchaulet: Created https://github.com/elastic/package-registry/issues/1693 for follow up

PR #1692, opened 2026-04-28
Add steps to copy release images to Docker Hub in Buildkite pipeline

### LINKED ISSUE #1270: Release images on Docker Hub instead of just docker.elastic.co
As per title, it would be interesting to have the EPR images published to hub.docker.com/u/elastic

teresaromero:
## Summary

Closes #1270

Adds Buildkite pipeline steps to copy released Docker images from `docker.elastic.co` to DockerHub (`hub.docker.com/u/elastic`) as part of the release process.

### `package-registry/package-registry` (empty image)

On every semver release tag (`v*.*.*`), two new trigger steps copy the standard and UBI images to DockerHub after the `publish` step completes:

- `package-registry/package-registry:<tag>`
- `package-registry/package-registry:<tag>-ubi`

### `package-registry/distribution` (distribution image)

The existing release distribution pipeline now also copies all four image variants to DockerHub after the release steps complete:

- `package-registry/distribution:<tag>` (production)
- `package-registry/distribution:<tag>-ubi` (production UBI)
- `package-registry/distribution:lite-<tag>` (lite)
- `package-registry/distribution:lite-<tag>-ubi` (lite UBI)

## Notes

The original issue scoped only the **empty `package-registry/package-registry` image** for DockerHub publishing, due to concerns about the large size of distribution images. This PR also includes the distribution image variants to cover all images released by these pipelines — however, whether `package-registry/distribution` (and in particular the `lite` variants) should be published to DockerHub was not explicitly confirmed. This is flagged for reviewer awareness.

- Related https://github.com/elastic/platform-engineering-productivity/issues/954#issuecomment-4305256985


prodsecmachine:

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/c6435a62-57b3-4544-91c5-a7d4343fecf7?refs=2e048e1e109a7f59fdb75b8a4b22d7af0c3dc7db&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/beats/pr-checks/c6435a62-57b3-4544-91c5-a7d4343fecf7/license?refs=2e048e1e109a7f59fdb75b8a4b22d7af0c3dc7db&source=prChecksComment) |

.buildkite/release-package-registry-distribution.yml
```diff
@@ -49,6 +49,52 @@ steps:
       - elastic/vault-docker-login#v0.6.0:
           secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
 
+  - label: ":docker: Copy release images to Docker Hub"
+    key: copy-release-images-to-dockerhub
+    if: "build.source == 'ui' || build.env('BUILDKITE_TRIGGERED_FROM_BUILD_PIPELINE_SLUG') == 'fleet-server'"
+    command:
+      - |-
```
mrodm: According to the docs https://buildkite.com/docs/pipelines/configure/step-types/command-step
I think this should be like:

```suggestion
    command: |
```

.buildkite/release-package-registry-distribution.yml
```diff
@@ -49,6 +49,52 @@ steps:
...
+        set -euo pipefail
```
mrodm, line 56: Not sure this set -o pipefail would be supported. I'm not sure if Buildkite agent runs the commands with sh or bash. If it uses sh, that flag does not exist.

Could you add a dummy test in  `.buildkite/pipeline.yml` so it gets executed in this PR that uses this same flag to validate if that would work successfully?

```yaml
  - label: "Dummy test"
    key: dummy-test
    command: |
      set -euo pipefail
      echo "Test message"
```

.buildkite/pipeline.yml
```diff
@@ -112,6 +112,32 @@ steps:
       - elastic/vault-docker-login#v0.6.0:
           secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
 
+  - trigger: unified-release-copy-elastic-images-to-dockerhub
```
mrodm, line 115: Does this pipeline offer a dry run option ? It would be interesting being able to test these pipelines before a new tag is pushed.

navyau09: Hey @mrodm `unified-release-copy-elastic-images-to-dockerhub` does not currently support a dry-run mode. When the job runs, it performs an actual crane copy from docker.elastic.co to Docker Hub ( index.docker.io).

The pipeline is driven by `IMAGES_NAMES` (comma-separated namespace/name) and IMAGES_TAG.
To test before using a production tag, you can try triggering the pipeline with a non-production tag that already exists upstream and verify the results on Docker Hub.

teresaromero:
Added some changes with the feedback from the official buildkite [skill](https://github.com/buildkite/skills) to review pipelines

for https://github.com/elastic/package-registry/pull/1692/changes#diff-c7cfaf86811e41aa2b847f18fdce17cb8285ab9fd1e337faa1431e0457bfa08f
```
  • Make pipeline intent visible in the UI: Swapped a log-only echo step for a buildkite-agent annotate message so the “who
    should run this and how” contract is visible on the build page without digging into logs.

  • Avoid conditional dependency traps: The validate step depended on an input step that is skipped for non-UI triggers.
    Adding allow_dependency_failure: true prevents builds from getting stuck or behaving inconsistently when a dependency is
     skipped.

  • Improve resilience to transient infra failures: Added retry.automatic for common transient exit statuses (-1, 143, 255)
    so agent loss/spot termination/connection hiccups don’t require manual reruns.

  • Prevent concurrent release races: Added concurrency + concurrency_group so two releases can’t run at once and fight over
     the same Docker tags (a common release-pipeline hazard).

  • Make downstream trigger results observable and retries safer: Switched Docker Hub copy triggers to async: false so the
    build reflects their success/failure, and added stable key: values to make retried uploads/steps more deterministic and
    easier to reason about.
```

for https://github.com/elastic/package-registry/pull/1692/changes#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53

```
  • Added timeouts to prevent hung jobs consuming agents indefinitely:
    • lint/build: 15 min
    • unit tests + smoke test + publish: 30 min
  • Added automatic retries for transient infra issues (exit_status: -1, 143, 255) on
    lint/build/smoke/tests/publish.
  • Added per-branch concurrency guard on publish to prevent tag/branch image races:
    • concurrency_group: "package-registry/publish-${BUILDKITE_BRANCH}", ordered
  • Saved CI compute on failures: cancel_on_build_failing: true on each OS unit-test step.
  • Made Docker Hub copy triggers observable: switched the two dockerhub copy triggers to async: false so the
    build reflects copy success/failure.
  • Documented release-tag gating: added a comment explaining the tag regex is release-only and excludes
    prereleases by design.
```

.buildkite/pipeline.yml
```diff
@@ -44,6 +71,16 @@ steps:
         key: test-linux
         command:
           - ".buildkite/scripts/run-tests.sh"
+        cancel_on_build_failing: true
```
mrodm: I think I will keep this with the default value (false).

If these steps are run, we get more information about all the errors in test even if the first steps fail.

.buildkite/pipeline.yml
```diff
@@ -65,6 +112,16 @@ steps:
...
+        timeout_in_minutes: 30
+        retry:
+          automatic:
+            - exit_status: -1
+              limit: 2
+            - exit_status: 143
+              limit: 2
+            - exit_status: 255
+              limit: 2
```
mrodm: I think I would keep these steps as before.
Not sure about the exit codes that the scripts will raise to allow retrying.

.buildkite/release-package-registry-distribution.yml
```diff
@@ -25,6 +35,7 @@ steps:
     depends_on:
       - step: "input"
         allow_failure: false
+    allow_dependency_failure: true
```
mrodm: Adding this `allow_dependency_failure` setting, would it be the same as the current `depends_on` field?

.buildkite/release-package-registry-distribution.yml
mrodm: Thanks for testing it !

.buildkite/release-package-registry-distribution.yml
```diff
@@ -49,6 +49,54 @@ steps:
...
+            label: ":docker: Copy release lite UBI images to Docker Hub"
+            key: copy-lite-ubi
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/distribution"
+                IMAGES_TAG: "lite-${DOCKER_TAG}-ubi"
+        YAML
+    agents:
+      provider: "gcp"
```
mrodm: It should not be needed to run this in GCP. As this just builds a new pipeline using unix commands, it could be used the default agent.



```suggestion
```

.buildkite/release-package-registry-distribution.yml
teresaromero: nop, this was a wrong recommendation, as adding this will continue with the release regardless of any dependency failing...

.buildkite/release-package-registry-distribution.yml
```diff
@@ -49,6 +49,52 @@ steps:
...
+            label: ":docker: Copy release lite UBI images to Docker Hub"
+            key: copy-lite-ubi
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/distribution"
+                IMAGES_TAG: "lite-${DOCKER_TAG}-ubi"
+        YAML
+    depends_on:
+      - step: "release-distribution"
```
mrodm, line 96: It would be better to not trigger these new steps if there is a failure in the `release-distribution` step.

```suggestion
    depends_on:
      - step: "release-distribution"
        allow_failure: false
```

If there is a failure in that previous step, it means there has been some issue re-tagging the distribution docker images, and it could fail these steps since maybe those tags do not exist. Depending on the error, maybe there is some tag pushed, but not all of them I guess.

WDYT?

mrodm (approved):
LGTM

It will require to pay attention to these pipeline when new tags are pushed:
- New tag pushed to package-registry via GitHub releases that triggers the pipeline https://buildkite.com/elastic/package-registry
- New tag pushed to fleet-server [that triggers](https://github.com/elastic/fleet-server/blob/0400cf321e29ce2c8659f5edceeabff560ab320f/.buildkite/pipeline.yml#L356-L363) the pipeline https://buildkite.com/elastic/package-registry-release-package-registry-distribution

PR #1735, opened 2026-06-04
Update stalled indirect dependencies

jsoriano:
Dependabot doesn't update indirect dependencies by default.

If they are not updated by the direct dependencies they may remain in old versions.

Update all of them and include their updates in dependabot executions. Including them in dependabot may be noisy, let's try during some days.


.github/dependabot.yml
```diff
@@ -8,14 +8,21 @@ updates:
     labels:
       - automation
     open-pull-requests-limit: 10
+    allow:
+      - dependency-type: "all" # Also indirect dependencies.
```
mrodm, line 12: According to the comment, should this be "indirect" or "all" ?

We can try with "all" as you suggested, and if it is too noisy there is also the option to set `indirect`:

> Dependencies of direct dependencies (also known as sub-dependencies, or transitive dependencies).

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#dependency-type-allow

mrodm (approved):

.github/dependabot.yml
jsoriano: `indirect` would only target indirect dependencies. For `gomod`, `all` is direct and indirect dependencies. I was using "Also" in the comment because by default it only updates direct dependencies.

PR #1801, opened 2026-06-30
Add agentless release field support

MichelLosier:
In support of: https://github.com/elastic/package-spec/issues/1098

* Adds support of agentless deployment mode `release` field so its exposed in search responses

### DIFF
--- a/packages/package.go
+++ b/packages/package.go
@@ -236,8 +236,9 @@ type DeploymentModes struct {
 }
 
 type DeploymentMode struct {
-	Enabled   bool  `config:"enabled" json:"enabled" yaml:"enabled" validate:"required"`
-	IsDefault *bool `config:"is_default" json:"is_default,omitempty" yaml:"is_default,omitempty"`
+	Enabled   bool   `config:"enabled" json:"enabled" yaml:"enabled" validate:"required"`
+	IsDefault *bool  `config:"is_default" json:"is_default,omitempty" yaml:"is_default,omitempty"`
+	Release   string `config:"release,omitempty" json:"release,omitempty" yaml:"release,omitempty"`
 }
 
 type Deprecated struct {
@@ -368,6 +369,10 @@ func newPackage(basePath string, fsBuilder FileSystemBuilder) (*Package, error) 
 		// Fill ingestion method for input packages.
 		if p.Type == "input" && p.PolicyTemplates[i].Input != "" {
 			p.PolicyTemplates[i].IngestionMethod = IngestionMethods.Get(p.PolicyTemplates[i].Input)
+		}
+
+		if err := validateDeploymentModeRelease(p.PolicyTemplates[i].DeploymentModes); err != nil {
+			return nil, fmt.Errorf("policy template %q: %w", p.PolicyTemplates[i].Name, err)
 		}
 	}
 
@@ -623,6 +628,16 @@ func (p *Package) IsNewerOrEqual(pp *Package) bool {
 
 func (p *Package) IsPrerelease() bool {
 	return isPrerelease(p.versionSemVer)
+}
+
+func validateDeploymentModeRelease(modes *DeploymentModes) error {
+	if modes == nil || modes.Agentless == nil || modes.Agentless.Release == "" {
+		return nil
+	}
+	if !IsValidAgentlessRelease(modes.Agentless.Release) {
+		return fmt.Errorf("invalid agentless release: %q", modes.Agentless.Release)
+	}
+	return nil
 }
 
 func (p *Package) isTechPreview() bool {
--- a/packages/package.go
+++ b/packages/package.go
@@ -638,7 +638,7 @@ func validateDeploymentModeRelease(modes *DeploymentModes) error {
 	if modes == nil || modes.Agentless == nil || modes.Agentless.Release == "" {
 		return nil
 	}
-	if !IsValidAgentlessRelease(modes.Agentless.Release) {
+	if !isValidAgentlessRelease(modes.Agentless.Release) {
 		return fmt.Errorf("invalid agentless release: %q", modes.Agentless.Release)
 	}
 	return nil
--- a/packages/package.go
+++ b/packages/package.go
@@ -231,11 +231,15 @@ type Download struct {
 }
 
 type DeploymentModes struct {
-	Default   *DeploymentMode `config:"default,omitempty" json:"default,omitempty" yaml:"default,omitempty"`
-	Agentless *DeploymentMode `config:"agentless,omitempty" json:"agentless,omitempty" yaml:"agentless,omitempty"`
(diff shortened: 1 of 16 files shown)


packages/package.go
```diff
@@ -236,8 +236,9 @@ type DeploymentModes struct {
 }
 
 type DeploymentMode struct {
```
teresaromero: The current DeploymentMode struct is shared between default and agentless, but the [package-spec](https://github.com/elastic/package-spec/blob/main/spec/integration/manifest.spec.yml) defines them as two distinct schemas with different fields: default only has enabled, while agentless has enabled, is_default, organization, division, team, release, and resources.

This was fine originally when DeploymentMode only held Enabled, but as agentless-specific fields have been added (is_default, and now release), they land on the shared struct — making deployment_modes.default.release and deployment_modes.default.is_default grammatically valid in the YAML and silently accepted by the parser, even though neither field exists in the spec under default. The spec enforces this with additionalProperties: false on the agentless block; the EPR has no equivalent guard.

Would it make sense to split this into two separate structs aligned with what the spec actually defines?

```
type DefaultDeploymentMode struct {
    Enabled bool `config:"enabled" json:"enabled" yaml:"enabled" validate:"required"`
}

type AgentlessDeploymentMode struct {
    Enabled   bool   `config:"enabled" json:"enabled" yaml:"enabled" validate:"required"`
    IsDefault *bool  `config:"is_default" json:"is_default,omitempty" yaml:"is_default,omitempty"`
    Release   string `config:"release,omitempty" json:"release,omitempty" yaml:"release,omitempty"`
}

type DeploymentModes struct {
    Default   *DefaultDeploymentMode   `config:"default,omitempty" json:"default,omitempty" yaml:"default,omitempty"`
    Agentless *AgentlessDeploymentMode `config:"agentless,omitempty" json:"agentless,omitempty" yaml:"agentless,omitempty"`
}
```

packages/package.go
```diff
@@ -625,6 +630,16 @@ func (p *Package) IsPrerelease() bool {
 	return isPrerelease(p.versionSemVer)
 }
 
+func validateDeploymentModeRelease(modes *DeploymentModes) error {
```
teresaromero, line 637: now that we have validation modes at package-spec i think it will be interesting to evaluate if we could import the package-spec validator itself, using BuildMode. All packages on EPR should be in build form. This way we avoid duplicating rules and the need of mantaining them. Opened an issue https://github.com/elastic/package-registry/issues/1805 

@elastic/ecosystem wdyt?

testdata/package/deployment_modes/0.0.1/manifest.yml
teresaromero, line 1: Hey, small thing I noticed while looking at the fixture — the format_version in testdata/package/deployment_modes/0.0.1/manifest.yml is set to 3.3.0, but the manifest is using fields that came in later spec versions: is_default landed in 3.3.2, input-level deployment_modes arrays in 3.4.1, and the new release field in 3.6.3. So the fixture is technically claiming to be an older package than it actually is.

Also worth updating the kibana version constraint from ^8.15.2 — the spec itself notes that release is only evaluated in Kibana 9.5.0 or later, so ^9.5.0 would be more accurate there.

Neither of these breaks anything at runtime, but keeping the fixture honest makes it a better reference for future contributors 🙂

If we go ahead with the package-spec validation migration we would probably spot a lot of errors on the fixtures! :D

teresaromero (commented):
thanks! i left a couple of comments

packages/package.go
mrodm: Yes, it makes sense to apply that change. This will make to be in sync with what package-spec has 👍 

It should be validated that this function still copies everything (that I think it should):

```
func (p *Package) setBasePolicyTemplates() {
```

This function allows to copy those fields so they are also available in the response of queries like 
https://epr.elastic.co/package/aws/6.20.2/

testdata/package/deployment_modes/0.0.1/manifest.yml
mrodm: > If we go ahead with the package-spec validation migration we would probably spot a lot of errors on the fixtures! :D 

I guess that would be true

packages/package.go
mrodm: I guess this could be worth to investigate. It will force us to use always correct packages and it could make sense.

mrodm (commented):
LGTM after applying the change in structs proposed by @teresaromero 

And, a new entry in the [CHANGELOG](https://github.com/elastic/package-registry/blob/main/CHANGELOG.md) should be added for the change introduced in this PR.

MichelLosier:
> It should be validated that this function still copies everything (that I think it should):

> func (p *Package) setBasePolicyTemplates() {

Validated with the generated response in https://github.com/elastic/package-registry/pull/1801/changes#diff-b6077d0574b85ce331f833a246bc0f7eef3db5831e9ba707b02f4b13b26d8c25

`DeploymentModes` struct refactor and changelog entry addressed in https://github.com/elastic/package-registry/pull/1801/commits/05bff18902aa0bb75492c376dfcb15b2fec1bde3

packages/releases.go
```diff
@@ -26,11 +26,23 @@ var ReleaseTypes = map[string]interface{}{
 	ReleaseGa:           nil,
 }
 
+// AgentlessReleaseTypes contains valid release values for deployment_modes.agentless.release.
+// Unlike the package-level release, "experimental" is not valid here.
+var AgentlessReleaseTypes = map[string]interface{}{
+	ReleaseBeta: nil,
+	ReleaseGa:   nil,
+}
```
mrodm: Currently, this variable is not used in any other package. I would make it private:


```
// agentlessReleaseTypes contains valid release values for deployment_modes.agentless.release.
// Unlike the package-level release, "experimental" is not valid here.
var agentlessReleaseTypes = map[string]interface{}{
	ReleaseBeta: nil,
	ReleaseGa:   nil,
}
```

packages/releases.go
```diff
@@ -26,11 +26,23 @@ var ReleaseTypes = map[string]interface{}{
...
+
 func IsValidRelease(release string) bool {
 	_, exists := ReleaseTypes[release]
 	return exists
 }
 
+func IsValidAgentlessRelease(release string) bool {
+	_, exists := AgentlessReleaseTypes[release]
+	return exists
+}
```
mrodm: Same as the other comment, this variable is not used in other packages. We could keep it private for now.


```
func isValidAgentlessRelease(release string) bool {
	_, exists := AgentlessReleaseTypes[release]
	return exists
}
```

mrodm (commented):
LGTM, just added some comments to avoid making public some structs and functions.

mrodm (approved):
Thanks!

teresaromero (approved):

PR #1808, opened 2026-07-03
Publish distribution image to DockerHub as package-registry-distribution

teresaromero:
## Summary
- Retag the internal `docker.elastic.co/package-registry/distribution` image as `package-registry/package-registry-distribution` (same digest) so it can be published under a non-misleading name.
- Update the DockerHub copy step to publish as `elastic/package-registry-distribution` instead of `elastic/distribution`.

## Related issue
Related to elastic/platform-engineering-productivity#2953, which tracks deleting the old `elastic/distribution` image from DockerHub once this is live.

Follow-up of #1692, which added the DockerHub copy step and introduced the `elastic/distribution` name.

### DIFF
--- a/.buildkite/release-package-registry-distribution.yml
+++ b/.buildkite/release-package-registry-distribution.yml
@@ -56,7 +56,7 @@ steps:
     command: |
         set -euo pipefail
 
-        DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="${DRY_RUN:-"false"}")"
+        DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="$${DRY_RUN:-"false"}")"
         if [[ "$${DRY_RUN:-"false"}" == "true" ]]; then
             echo "Skip triggering steps to release images to Docker Hub"
             exit 0
--- a/.buildkite/release-package-registry-distribution.yml
+++ b/.buildkite/release-package-registry-distribution.yml
@@ -50,11 +50,17 @@ steps:
       - elastic/vault-docker-login#v0.6.0:
           secret_path: 'kv/ci-shared/platform-ingest/elastic_docker_registry'
 
-  - label: ":docker: Copy release images to Docker Hub"
-    key: copy-release-images-to-dockerhub
+  - label: ":docker: Trigger copy release images to Docker Hub"
+    key: trigger-copy-release-images-to-dockerhub
     if: "build.source == 'ui' || build.env('BUILDKITE_TRIGGERED_FROM_BUILD_PIPELINE_SLUG') == 'fleet-server'"
     command: |
         set -euo pipefail
+
+        DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="${DRY_RUN:-"false"}")"
+        if [[ "$${DRY_RUN:-"false"}" == "true" ]]; then
+            echo "Skip triggering steps to release images to Docker Hub"
+            exit 0
+        fi
 
         DOCKER_TAG="$$(buildkite-agent meta-data get DOCKER_TAG --default="")"
         if [[ -z "$${DOCKER_TAG:-""}" ]]; then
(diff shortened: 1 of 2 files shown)


.buildkite/release-package-registry-distribution.yml
mrodm: I would rename this label and key to add that this is related to triggering those steps.

Proposal:


```suggestion
- label: ":docker: Trigger copy release images to Docker Hub"
key: trigger-copy-release-images-to-dockerhub
```

mrodm, line 58: Looking at the previous step, the script manages the case of the environment variable `DRY_RUN` being true to skip the publication.

Here we should do the same to keep consistency.

What about adding this ? 

```sh
DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="${DRY_RUN:-"false"}")"
if [[ "$${DRY_RUN:-"false"} == "true" ]]; then
    echo "Skip triggering steps to release images to Docker Hub"
    exit 0
fi
```

In case we want to trigger this step from the UI in DRY_RUN true mode.

mrodm: Maybe this could be moved to another pull request.

what I'm thinking is to add another input item

```yaml
  - input: ":truffleruby: Input version for docker tag"
    key: input
    if: "build.source == 'ui'"
    fields:
      - text: "The docker tag to be published (format: major.minor.patch(-prerelease)?)."
        hint: "e.g. 0.0.1"
        key: "DOCKER_TAG"
        required: true
      - text: "DRY_RUN."
        key: DRY_RUN
        options:
          - label: "True"
            value: "true"
          - label: "False"
            value: "false"
        default: "true"
```

Then remove the hard-coded value for `DRY_RUN` from the `env` key in the pipeline definition:

https://github.com/elastic/package-registry/blob/3f85ee1619a58a20a4c33fcd3026fdcf007aae92/.buildkite/release-package-registry-distribution.yml#L40

And update [the script](https://github.com/elastic/package-registry/blob/3f85ee1619a58a20a4c33fcd3026fdcf007aae92/.buildkite/scripts/publish-distribution.sh#L7-L13) for that to include the retrieval of that variable

```sh
DRY_RUN="$(buildkite-agent meta-data get DRY_RUN --default="${DRY_RUN:-"false"}")"
if [[ "${DRY_RUN:-"false"} == "true" ]]; then
    echo "Skip triggering steps to release images to Docker Hub"
    exit 0
fi
```

So, if the pipeline is triggered from another pipeline (using a `trigger` step) by default `DRY_RUN` will get the value `false` if not defined.

When triggered using the UI, by default will be `true` this variable (from the input step). Values filled from the input step are retrieved from the meta-data.

teresaromero: i've added the DRY_RUN guard, i think the other should be part of a dedicated PR as a followup improvement

.buildkite/release-package-registry-distribution.yml
```diff
@@ -49,12 +50,18 @@ steps:
...
 
-  - label: ":docker: Copy release images to Docker Hub"
-    key: copy-release-images-to-dockerhub
+  - label: ":docker: Trigger copy release images to Docker Hub"
+    key: trigger-copy-release-images-to-dockerhub
     if: "build.source == 'ui' || build.env('BUILDKITE_TRIGGERED_FROM_BUILD_PIPELINE_SLUG') == 'fleet-server'"
     command: |
         set -euo pipefail
 
+        DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="${DRY_RUN:-"false"}")"
```
mrodm: Sorry, just realized that I missed a `$` in my suggestion


```suggestion
        DRY_RUN="$$(buildkite-agent meta-data get DRY_RUN --default="$${DRY_RUN:-"false"}")"
```

.buildkite/release-package-registry-distribution.yml
teresaromero: followup https://github.com/elastic/package-registry/issues/1809

mrodm (approved):
🚀

PR #1858, opened 2026-07-15
feat: add FIPS 140-3 build support

teresaromero:
## Why

Elastic Package Registry must be deployable in FIPS 140-3 regulated environments
(government, defense, financial). Following the Go ecosystem team's direction
(elastic/beats#51345), this uses Go's native certified FIPS module (`GOFIPS140=v1.0.0`,
CMVP cert #4735) rather than microsoft-go/OpenSSL — no CGO, no special toolchain
image, the entire certified module lives in the binary. Go 1.26.4 (already in use)
supports `GOFIPS140` natively; no toolchain upgrade was needed.

Closes elastic/ingest-dev#7234

---

### Build

- **`Makefile`**: `FIPS ?= 0`; when `FIPS=1`, prefixes `go build` with `GOFIPS140=v1.0.0`.
- **`Dockerfile`**: `ARG FIPS=0` passed through to `make release-... FIPS=${FIPS}`. Same builder image, no CGO.
- **`magefile.go`**: Three new targets — `BuildFIPS`, `DockerBuildFIPS`, `TestFIPS`.
  `TestFIPS` runs `go test ./...` against the `package-registry` module only,
  with `GOFIPS140=v1.0.0` and strict `GODEBUG=fips140=only`.

### FIPS compliance fixes

Under `GODEBUG=fips140=only`, three test packages panicked on `crypto/md5` — traced to
`github.com/fsouza/fake-gcs-server` hashing object content with MD5 when
`ObjectAttrs.Md5Hash` is empty (both on startup and on `Server.CreateObject`).

Fixed in **`internal/storage/fakestorage.go`**:
- `fakeObjectMD5Hash`: placeholder pre-populated on every fake object so the
  library never computes MD5 itself. Harmless — the real GCS client never reads it.
- `UpdateFakeServer`: rewrote to stop and restart the server with fresh
  `InitialObjects` instead of calling `Server.CreateObject`, keeping object
  creation off the MD5-recomputing path.
- Updated 8 call sites in `storage/indexer_test.go` and
  `internal/storage/sqlindexer_test.go`.

`go test ./...` now passes clean under strict `GODEBUG=fips140=only` across every
package in the `package-registry` module.

Note: `internal/storage/fakestorage.go` compiles into the shipped binary (not
test-only). The `EPR_EMULATOR_INDEX_PATH` production path uses the same
`PrepareServerObjects` code — verified to run without panic under `fips140=only`.

### TLS guardrail

`-tls-min-version=1.1` would violate FIPS 140-3. FIPS builds now reject startup
with a clear error if this flag is set below TLS 1.2. Non-FIPS builds are
unaffected. Implemented via `isFIPSBinary()` (reads `debug.ReadBuildInfo()` for
the `GOFIPS140` build setting) and three unit tests in `flags_test.go`.

### CI

- **`.buildkite/pipeline.yml`**: New `fips-build-test` step (`allow_failure: false`,
  required gate on `publish`).
- **`.buildkite/scripts/test-fips.sh`** (new): builds FIPS binary, asserts
  `GOFIPS140=v1.0.0*` and `DefaultGODEBUG=fips140=on` in buildinfo, runs `mage testFIPS`.
- **`.buildkite/scripts/publish.sh`**: pushes `-fips` tagged image using
  `docker.elastic.co/wolfi/chainguard-base-fips` as the runtime base.

### Documentation

- **`docs/fips.md`** (new): build/verify instructions, what is covered, known limitations.
- **`README.md`**: FIPS 140-3 section linking to `docs/fips.md`.
- **`CHANGELOG.md`**: entry under `[unreleased] / Added`.

---

### `cmd/distribution` is not FIPS-covered

`cmd/distribution` verifies packages against Elastic's release-signing key via
`github.com/ProtonMail/go-crypto/openpgp`. That key is an OpenPGP v4 key; RFC 4880
§12.2 mandates SHA-1 for all v4 fingerprints, computed unconditionally at parse
time. Under `fips140=only` this panics before any verification takes place.

Fixing it requires Elastic to re-issue the signing key as a v6 key (SHA-256
fingerprint) — an org-level change outside this repo. Note: SHA-1 is only used
for the key identifier; actual package integrity verification uses SHA-256.
Documented in `docs/fips.md`.

---

## How to verify

```bash
mage buildFIPS
go version -m package-registry | grep -E 'GOFIPS140|DefaultGODEBUG'
# build  DefaultGODEBUG=fips140=on
# build  GOFIPS140=v1.0.0-...

mage testFIPS
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,9 +1,9 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
 env:
-  SETUP_GVM_VERSION: 'v0.6.0' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
-  SETUP_MAGE_VERSION: '1.15.0'
-  JQ_VERSION: '1.7'
+  SETUP_GVM_VERSION: "v0.6.0" # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
+  SETUP_MAGE_VERSION: "1.15.0"
+  JQ_VERSION: "1.7"
   DOCKER_IMG: "docker.elastic.co/package-registry/package-registry"
   DOCKER_IMG_PR: "docker.elastic.co/observability-ci/package-registry"
   # Agent images used in pipeline steps
@@ -145,6 +145,19 @@ steps:
       env:
         IMAGES_NAMES: "package-registry/package-registry"
         IMAGES_TAG: "${BUILDKITE_TAG}-ubi"
+    depends_on:
+      - step: "publish"
+        allow_failure: false
+    if: build.tag =~ /^v\d+\.\d+\.\d+$$/
+
+  - trigger: unified-release-copy-elastic-images-to-dockerhub
+    label: ":docker: Copy release FIPS images to Docker Hub"
+    key: "copy-release-fips-images-to-dockerhub"
+    async: false
+    build:
+      env:
+        IMAGES_NAMES: "package-registry/package-registry"
+        IMAGES_TAG: "${BUILDKITE_TAG}-fips"
     depends_on:
       - step: "publish"
         allow_failure: false
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,6 +37,15 @@ steps:
     agents:
       provider: "gcp"
 
+  - label: ":lock: FIPS 140-3 build & test"
+    key: fips-build-test
+    command:
+      - ".buildkite/scripts/test-fips.sh"
+    agents:
+      image: "${LINUX_GOLANG_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
   - group: ":go: Run Unit tests"
     key: unit-tests
     steps:
@@ -107,6 +116,8 @@ steps:
       - step: "lint"
         allow_failure: false
       - step: "smoke-test-fake-gcs-server"
+        allow_failure: false
+      - step: "fips-build-test"
         allow_failure: false
     plugins:
       - elastic/vault-docker-login#v0.6.0:
--- a/.buildkite/release-package-registry-distribution.yml
+++ b/.buildkite/release-package-registry-distribution.yml
@@ -87,6 +87,14 @@ steps:
                 IMAGES_NAMES: "package-registry/package-registry-distribution"
                 IMAGES_TAG: "$${DOCKER_TAG}-ubi"
           - trigger: unified-release-copy-elastic-images-to-dockerhub
+            label: ":docker: Copy release FIPS images to Docker Hub"
+            key: copy-prod-fips
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/package-registry-distribution"
+                IMAGES_TAG: "$${DOCKER_TAG}-fips"
+          - trigger: unified-release-copy-elastic-images-to-dockerhub
             label: ":docker: Copy release lite images to Docker Hub"
             key: copy-lite
             async: false
@@ -102,6 +110,14 @@ steps:
               env:
                 IMAGES_NAMES: "package-registry/package-registry-distribution"
                 IMAGES_TAG: "lite-$${DOCKER_TAG}-ubi"
+          - trigger: unified-release-copy-elastic-images-to-dockerhub
+            label: ":docker: Copy release lite FIPS images to Docker Hub"
+            key: copy-lite-fips
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/package-registry-distribution"
+                IMAGES_TAG: "lite-$${DOCKER_TAG}-fips"
         PIPELINE
     depends_on:
       - step: "release-distribution"
--- a/.buildkite/release-package-registry-distribution.yml
+++ b/.buildkite/release-package-registry-distribution.yml
@@ -68,6 +68,34 @@ steps:
             exit 1
         fi
 
+        upload_fips_copy_step() {
+            local tag_name="$$1"
+            local image_tag="$$2"
+            local step_key="$$3"
+            local availability
+            availability="$$(buildkite-agent meta-data get "fips-distribution-$${tag_name}-available" --default="false")"
+
+            if [[ "$${availability}" != "true" ]]; then
+                echo "Skip triggering FIPS image copy for $${tag_name}: source image was not available"
+                return
+            fi
+
+            buildkite-agent pipeline upload << PIPELINE
+        steps:
+          - trigger: unified-release-copy-elastic-images-to-dockerhub
+            label: ":docker: Copy release $${tag_name} FIPS images to Docker Hub"
+            key: "$${step_key}"
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/package-registry-distribution"
+                IMAGES_TAG: "$${image_tag}"
+        PIPELINE
+        }
+
+        upload_fips_copy_step "production" "$${DOCKER_TAG}-fips" "copy-prod-fips"
+        upload_fips_copy_step "lite" "lite-$${DOCKER_TAG}-fips" "copy-lite-fips"
+
         buildkite-agent pipeline upload << PIPELINE
         steps:
           - trigger: unified-release-copy-elastic-images-to-dockerhub
@@ -87,14 +115,6 @@ steps:
                 IMAGES_NAMES: "package-registry/package-registry-distribution"
                 IMAGES_TAG: "$${DOCKER_TAG}-ubi"
           - trigger: unified-release-copy-elastic-images-to-dockerhub
-            label: ":docker: Copy release FIPS images to Docker Hub"
-            key: copy-prod-fips
-            async: false
-            build:
-              env:
-                IMAGES_NAMES: "package-registry/package-registry-distribution"
-                IMAGES_TAG: "$${DOCKER_TAG}-fips"
-          - trigger: unified-release-copy-elastic-images-to-dockerhub
             label: ":docker: Copy release lite images to Docker Hub"
             key: copy-lite
             async: false
@@ -110,14 +130,6 @@ steps:
               env:
                 IMAGES_NAMES: "package-registry/package-registry-distribution"
                 IMAGES_TAG: "lite-$${DOCKER_TAG}-ubi"
-          - trigger: unified-release-copy-elastic-images-to-dockerhub
-            label: ":docker: Copy release lite FIPS images to Docker Hub"
-            key: copy-lite-fips
-            async: false
-            build:
-              env:
-                IMAGES_NAMES: "package-registry/package-registry-distribution"
-                IMAGES_TAG: "lite-$${DOCKER_TAG}-fips"
         PIPELINE
     depends_on:
       - step: "release-distribution"
--- a/.buildkite/scripts/publish-distribution.sh
+++ b/.buildkite/scripts/publish-distribution.sh
@@ -22,6 +22,7 @@ DOCKER_IMAGE_TARGETS=("${DOCKER_IMAGE}")
 if [[ -n "${DOCKER_IMAGE_RENAMED:-""}" ]]; then
     DOCKER_IMAGE_TARGETS+=("${DOCKER_IMAGE_RENAMED}")
 fi
+IMAGE_SUFFIXES=("" "-ubi" "-fips")
 
 echo "Docker retag"
 docker buildx create --use
@@ -33,13 +34,16 @@ for image in "${DOCKER_IMAGE_TARGETS[@]}"; do
         DOCKER_IMG_TARGET="${image}:${TAG_NAME}-${DOCKER_TAG}"
     fi
 
-    # do not push if DRY_RUN is true
-    if [[ ${DRY_RUN:-true} == "true" ]]; then
-        docker buildx imagetools create --dry-run -t "${DOCKER_IMG_TARGET}" "${DOCKER_IMG_SOURCE}"
-        docker buildx imagetools create --dry-run -t "${DOCKER_IMG_TARGET}-ubi" "${DOCKER_IMG_SOURCE}-ubi"
-    else
-        retry 3 docker buildx imagetools create -t "${DOCKER_IMG_TARGET}" "${DOCKER_IMG_SOURCE}"
-        retry 3 docker buildx imagetools create -t "${DOCKER_IMG_TARGET}-ubi" "${DOCKER_IMG_SOURCE}-ubi"
-        echo "Docker image pushed: ${DOCKER_IMG_TARGET}"
-    fi
+    for suffix in "${IMAGE_SUFFIXES[@]}"; do
+        source_image="${DOCKER_IMG_SOURCE}${suffix}"
+        target_image="${DOCKER_IMG_TARGET}${suffix}"
+
+        # do not push if DRY_RUN is true
+        if [[ ${DRY_RUN:-true} == "true" ]]; then
+            docker buildx imagetools create --dry-run -t "${target_image}" "${source_image}"
+        else
+            retry 3 docker buildx imagetools create -t "${target_image}" "${source_image}"
+            echo "Docker image pushed: ${target_image}"
+        fi
+    done
 done
--- a/.buildkite/scripts/publish-distribution.sh
+++ b/.buildkite/scripts/publish-distribution.sh
@@ -22,7 +22,18 @@ DOCKER_IMAGE_TARGETS=("${DOCKER_IMAGE}")
 if [[ -n "${DOCKER_IMAGE_RENAMED:-""}" ]]; then
     DOCKER_IMAGE_TARGETS+=("${DOCKER_IMAGE_RENAMED}")
 fi
-IMAGE_SUFFIXES=("" "-ubi" "-fips")
+IMAGE_SUFFIXES=("" "-ubi")
+FIPS_SOURCE_IMAGE="${DOCKER_IMG_SOURCE}-fips"
+FIPS_SOURCE_AVAILABLE=false
+
+# The FIPS distribution source does not exist until the first release that
+# includes a FIPS package-registry image. Skip it during that transition.
+if retry 3 docker buildx imagetools inspect "${FIPS_SOURCE_IMAGE}" > /dev/null; then
+    IMAGE_SUFFIXES+=("-fips")
+    FIPS_SOURCE_AVAILABLE=true
+else
+    echo "FIPS source image is not available yet, skipping: ${FIPS_SOURCE_IMAGE}"
+fi
 
 echo "Docker retag"
 docker buildx create --use
@@ -47,3 +58,5 @@ for image in "${DOCKER_IMAGE_TARGETS[@]}"; do
         fi
     done
 done
+
+buildkite-agent meta-data set "fips-distribution-${TAG_NAME}-available" "${FIPS_SOURCE_AVAILABLE}"
--- a/.buildkite/scripts/publish.sh
+++ b/.buildkite/scripts/publish.sh
@@ -6,6 +6,7 @@ set -euo pipefail
 build_push_docker_image() {
 	local runner_image="${1}"
 	local tag_suffix="${2:-}"
+	local fips="${3:-0}"
 	local docker_img_tag="${DOCKER_IMG_TAG}${tag_suffix}"
 	local docker_img_tag_branch="${DOCKER_IMG_TAG_BRANCH}${tag_suffix}"
 	go_version=$(cat .go-version)
@@ -20,6 +21,7 @@ build_push_docker_image() {
 		--build-arg GO_VERSION="${go_version}" \
 		--build-arg BUILDER_IMAGE=docker.elastic.co/wolfi/go \
 		--build-arg RUNNER_IMAGE="${runner_image}" \
+		--build-arg FIPS="${fips}" \
 		--label BRANCH_NAME="${TAG_NAME}" \
 		--label GIT_SHA="${BUILDKITE_COMMIT}" \
 		--label GO_VERSION="${SETUP_GOLANG_VERSION}" \
@@ -47,3 +49,4 @@ DOCKER_IMG_TAG_BRANCH="${DOCKER_NAMESPACE}:${TAG_NAME}"
 
 build_push_docker_image "docker.elastic.co/wolfi/chainguard-base" ""
 build_push_docker_image "registry.access.redhat.com/ubi9/ubi-minimal:9.6" "-ubi"
+build_push_docker_image "docker.elastic.co/wolfi/chainguard-base-fips" "-fips" "1"
--- a/.buildkite/scripts/publish.sh
+++ b/.buildkite/scripts/publish.sh
@@ -6,7 +6,10 @@ set -euo pipefail
 build_push_docker_image() {
 	local runner_image="${1}"
 	local tag_suffix="${2:-}"
-	local fips="${3:-0}"
+	local fips=0
+	if [[ "${tag_suffix}" == "-fips" ]]; then
+		fips=1
+	fi
 	local docker_img_tag="${DOCKER_IMG_TAG}${tag_suffix}"
 	local docker_img_tag_branch="${DOCKER_IMG_TAG_BRANCH}${tag_suffix}"
 	go_version=$(cat .go-version)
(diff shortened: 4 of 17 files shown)


.buildkite/scripts/publish.sh
```diff
@@ -47,3 +49,4 @@ DOCKER_IMG_TAG_BRANCH="${DOCKER_NAMESPACE}:${TAG_NAME}"
 
 build_push_docker_image "docker.elastic.co/wolfi/chainguard-base" ""
 build_push_docker_image "registry.access.redhat.com/ubi9/ubi-minimal:9.6" "-ubi"
+build_push_docker_image "docker.elastic.co/wolfi/chainguard-base-fips" "-fips" "1"
```
teresaromero: @cmacknz Does `chainguard-base-fips` satisfy the OS-level FIPS posture you
raised in the issue? Should it be pinned to a specific digest rather than rolling
`latest`? A silent OpenSSL update in the base image could change the
CMVP-listed provider version.

.buildkite/pipeline.yml
teresaromero, line 127: @nimarezainia  Should `package-registry:vX.Y.Z-fips` be on Docker Hub
alongside the standard images, or is the Elastic registry the right distribution
channel for FIPS-regulated customers?

.buildkite/pipeline.yml
```diff
@@ -108,6 +117,8 @@ steps:
         allow_failure: false
       - step: "smoke-test-fake-gcs-server"
         allow_failure: false
+      - step: "fips-build-test"
+        allow_failure: false
```
teresaromero, line 121: @mrodm claudes review flagged this change:

> The new `fips-build-test` step is `allow_failure: false` (gates the `publish`
> step), and `publish.sh` pushes the `-fips` image using the same `vault-docker-login`
> vault secret as the existing images. This touches shared release infra.

is this correct?

teresaromero (commented):
**Does `cmd/distribution` need a FIPS build target?**

A FIPS build of `cmd/distribution` is not provided. The blocker is structural:
Elastic's release-signing key is OpenPGP v4, and RFC 4880 §12.2 mandates SHA-1
for v4 fingerprints — `go-crypto` computes this unconditionally at parse time,
panicking under `fips140=only` before any verification happens. No code change in
this repo fixes it; it requires re-issuing the key as v6 (org-level change).


**Should the shipped FIPS binary default to `fips140=only` instead of `fips140=on`?**

`DefaultGODEBUG=fips140=on` is what the Go toolchain automatically embeds when
building with `GOFIPS140=v1.0.0` — not an explicit choice in this code.

- `fips140=on`: certified module active; non-approved algorithm calls fall back silently.
- `fips140=only`: any non-approved algorithm call panics immediately, in production.

The test suite runs at `fips140=only` and passes clean, but test coverage is
never complete. A production code path not exercised by tests could silently call
a non-FIPS algorithm under `on`; `only` would surface it as a crash.

Does the compliance posture for regulated customer environments require `only`, or is `on` sufficient?

cc @cmacknz @mrodm

.buildkite/scripts/publish.sh
cmacknz: Using `GOFIPS140=v1.0.0` we do not depend on OpenSSL at all anymore and this locks the crypto module at the CMVP validated v1.0.0 Google Go one.

Using chainguard-base-fips is good regardless.

.buildkite/pipeline.yml
cmacknz: We publish the elastic-agent-fips container and artifacts alongside the regular elastic-agent ones so this should be fine, there is no need for any kind of special handling of the release artifacts just use the same release channels you use now with a fips name or tag.

internal/storage/fakestorage.go
```diff
@@ -74,17 +88,29 @@ func ClientNoAuth(server *fakestorage.Server) *storage.Client {
...
 	indexContent, err := os.ReadFile(indexPath)
 	require.NoError(tb, err, "index file must be populated")
 
 	serverObjects, numPackages, err := PrepareServerObjects(revision, indexContent)
 	require.NoError(tb, err, "failed to prepare server objects")
 	tb.Logf("Prepared %d packages with total %d server objects.", numPackages, len(serverObjects))
 
-	for _, so := range serverObjects {
-		server.CreateObject(so)
-	}
```
mrodm, line 87: It would be good to try to keep the previous behavior.

As it is now in this PR, `UpdateFakeServer` will just contain the new index and the new cursor file, but not the previous ones (via `PrepareServerObjects`).

This forces also to update the internal storage client form the indexer struct.

.buildkite/scripts/test-fips.sh
```diff
@@ -0,0 +1,33 @@
...
+mage -debug buildFIPS
+
+echo "--- Verify FIPS 140-3 build settings"
+build_info="$(go version -m package-registry)"
+echo "${build_info}"
+
+fips_version="$(echo "${build_info}" | grep -oE 'GOFIPS140=\S+' | cut -d= -f2-)"
+default_godebug="$(echo "${build_info}" | grep -oE 'DefaultGODEBUG=\S+' | cut -d= -f2-)"
+
+if [[ "${fips_version}" != v1.0.0* ]]; then
```
mrodm: Checking the value for GOFIPS140 locally, I see that it also contain a commit hash at the end
```
GOFIPS140=v1.0.0-c2097c7c
```

It could be added a comment about this to not that it is not just `v1.0.0`.

It could be used `=~` to validate the value, in case it is needed to be stricter:

```suggestion
if [[ "${fips_version}" =~ ^v1\.0\.0(-[a-f0-9]{5,40})?$ ]]; then
```

.buildkite/pipeline.yml
mrodm: I see this definition correct.
If the step `fips-build-test` fails, this step to publish the docker images will not be executed.

.buildkite/scripts/test-fips.sh
```diff
@@ -0,0 +1,33 @@
...
+
+echo "--- Build FIPS 140-3 binary"
+mage -debug buildFIPS
+
+echo "--- Verify FIPS 140-3 build settings"
+build_info="$(go version -m package-registry)"
+echo "${build_info}"
+
+fips_version="$(echo "${build_info}" | grep -oE 'GOFIPS140=\S+' | cut -d= -f2-)"
+default_godebug="$(echo "${build_info}" | grep -oE 'DefaultGODEBUG=\S+' | cut -d= -f2-)"
```
mrodm: Just a suggestion, this could be changed by 
```
fips_version=$(go version -m -json package-registry | jq -r '.Settings[] | select(.Key == "GOFIPS140") | .Value')

default_godebug=$(go version -m -json package-registry | jq -r '.Settings[] | select(.Key == "DefaultGODEBUG") | .Value')
```

But it would require to install `jq` probably in the step. Similar to what it was done for the `test-epr-gcs`:

https://github.com/elastic/package-registry/blob/c8a9bf66af006bca304d6ed718aca5c52e025f11/.buildkite/scripts/test-epr-gcs.sh#L19-L28

.buildkite/pipeline.yml
teresaromero: ok, i will add a block to publish this too under dockerhub then

internal/storage/fakestorage.go
teresaromero: I've done a change in order to mantain the objects when using Update.

.buildkite/scripts/publish.sh
```diff
@@ -6,6 +6,7 @@ set -euo pipefail
 build_push_docker_image() {
 	local runner_image="${1}"
 	local tag_suffix="${2:-}"
+	local fips="${3:-0}"
```
mrodm: Right now is not a problem, but in order to use this 3rd parameter, it is required to define the 2nd parameter too.

internal/storage/fakestorage.go
```diff
@@ -74,17 +88,48 @@ func ClientNoAuth(server *fakestorage.Server) *storage.Client {
...
+// converter (used on that path) strips Md5Hash before handing off to the
+// backend, which then recomputes it from scratch — a path that panics under
+// strict FIPS 140-3 enforcement (GODEBUG=fips140=only) since MD5 isn't an
+// approved algorithm. InitialObjects bypass that recomputation (see
+// PrepareFakeServer), so restarting the server keeps all object creation on
+// that path instead.
+//
+// New revision objects are appended after the existing ones so they win any
+// name conflicts (e.g. the cursor object is overwritten with the new revision).
+func UpdateFakeServer(tb testing.TB, server *fakestorage.Server, revision, indexPath string) *fakestorage.Server {
```
mrodm: What about adding a new parameter here to the indexer (with reference) so the function is in charge of updating the `storageClient` ?  This would hide from the caller that this must be performed.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add FIPS 140-3 build support via `GOFIPS140=v1.0.0`; new `mage buildFIPS`, `mage dockerBuildFIPS`, and `mage testFIPS` targets; `-fips` Docker image published per release. See [docs/fips.md](./docs/fips.md).
+* FIPS builds now reject startup if `-tls-min-version` is set below TLS 1.2.
+* `internal/storage`: `UpdateFakeServer` now restarts the fake GCS server instead of adding objects to the running one, keeping object creation on the `InitialObjects` path to avoid MD5 recomputation under strict FIPS mode.
```
mrodm: Missing the Pull Request links at the end of the message.

flags.go
```diff
@@ -70,6 +71,21 @@ func flagsFromEnv(flagSet *flag.FlagSet) error {
...
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return false
+	}
+	for _, s := range bi.Settings {
+		if s.Key == "GOFIPS140" && s.Value != "" {
+			return true
+		}
+	}
+	return false
```
mrodm: Nit.

```suggestion
	return slices.ContainsFunc(bi.Settings, func(s debug.BuildSetting) bool {
		return s.Key == "GOFIPS140" && s.Value != ""
	})
```

flags_test.go
```diff
@@ -53,3 +54,44 @@ func TestFlagEnvName(t *testing.T) {
...
+	t.Cleanup(func() {
+		isFIPSBinary = origIsFIPS
+		tlsMinVersionValue = origTLSMin
+		tlsCertFile = origCert
+		tlsKeyFile = origKey
+	})
+
+	// Provide dummy cert/key paths so the existing cert-presence check passes.
+	tlsCertFile = "cert.pem"
+	tlsKeyFile = "key.pem"
```
mrodm: Could these overrides have any side effect when running all the tests in package-registry ? Could it affect other tests in other packages ?

Another approach is to test this as part of the shell script that triggers the FIPS tests with something like:

```
 $ mage buildFIPS
 $ ./package-registry -tls-min-version 1.1  -tls-cert cert.pem -tls-key key.pem -dry-run -require-package-signatures=false
2026/07/16 15:07:38 FIPS 140-3 build: -tls-min-version 1.1 is not permitted; minimum allowed version is 1.2

 $ ./package-registry -tls-min-version 1.1  -tls-cert cert.pem -tls-key key.pem -dry-run -require-package-signatures=false
...
```

Or maybe add a note/comment that this test should not be run in parallel?

.buildkite/scripts/publish.sh
teresaromero: 👍🏻 thanks! i removed it from the parameter mode and determine this from the tag_suffix :)

flags_test.go
teresaromero: i've created a helper to validateTLS and did the tests over this helper to avoid global changes

internal/storage/fakestorage.go
teresaromero: I've took a look to this... and i've finally changed so the returned values of the func is the new server a client updated for the server. but still the change needs to be done outside

.buildkite/scripts/publish-distribution.sh
```diff
@@ -22,6 +22,7 @@ DOCKER_IMAGE_TARGETS=("${DOCKER_IMAGE}")
 if [[ -n "${DOCKER_IMAGE_RENAMED:-""}" ]]; then
     DOCKER_IMAGE_TARGETS+=("${DOCKER_IMAGE_RENAMED}")
 fi
+IMAGE_SUFFIXES=("" "-ubi" "-fips")
```
mrodm: The suffix `-fips` cannot be added until there exists a distribution with that tag in the docker registry.

And for that, it would require first to do a EPR release so there is available a docker based on FIPS for the package-registry service.

main.go
```diff
@@ -781,3 +779,16 @@ func validateFlags() error {
...
+func validateTLSFlags(certFile, keyFile string, minVersion tlsVersionValue, fips bool) error {
+	if minVersion == 0 {
+		return nil
+	}
+	if certFile == "" || keyFile == "" {
+		return fmt.Errorf("-tls-min-version set but missing TLS cert and key files (-tls-cert and -tls-key)")
+	}
+	if fips && uint16(minVersion) < tls.VersionTLS12 {
+		return fmt.Errorf("FIPS 140-3 build: -tls-min-version %s is not permitted; minimum allowed version is 1.2", minVersion)
+	}
```
mrodm, line 791: I'm not sure this would consider all cases.

Checking the code, if `tlsMinVersion` parameter is not set by the user, then the `tls.Config.MinVersion` will get the default value (0):

https://github.com/elastic/package-registry/blob/e58a73645011a0a0449b38535464080d496d359d/main.go#L515-L518


And in order to run the server in TLS mode it is required to set "just" `tlsCertFile` and `tlsKeyFile`:

https://github.com/elastic/package-registry/blob/e58a73645011a0a0449b38535464080d496d359d/main.go#L523-L528

So, what I mean is that even if `-tls-min-version` is not set, the TLS mode of the server can be enabled if the flags `-tls-cert` and `-tls-key` are set. In that case, it is still needed to be checked the default minimum version ?

It's true that starting with Go 1.22 the default value for servers TLS is 1.2 https://github.com/golang/go/issues/62459

Wondering if a part from checking if the flags are correct in the `validateFlags` function, should we force via code the implicit value instead of relying in the default value depending on the Go version?

```go
var tlsConfig tls.Config
if tlsMinVersionValue > 0 {
    tlsConfig.MinVersion = uint16(tlsMinVersionValue)
} else if isFIPSBinary() {
    tlsConfig.MinVersion = tls.VersionTLS12 // FIPS 140-3 requires TLS 1.2+
}
```

So would it be enough to just rely on the default go version since we're using Go 1.26 here or set it explictitly?

docs/fips.md
```diff
@@ -0,0 +1,120 @@
...
+variant, for example `docker.elastic.co/package-registry/package-registry:v1.40.0-fips`.
+
+Package-filled distribution images are also published with the `-fips` suffix.
+For an Elastic Stack release such as `8.19.0`, the production and lite variants
+are available as:
+
+```text
+docker.elastic.co/package-registry/distribution:8.19.0-fips
+docker.elastic.co/package-registry/distribution:lite-8.19.0-fips
+```
```
mrodm: Should we use a newer version for this example ? 9.5.0 or 9.6.0 ?

.buildkite/scripts/test-fips.sh
```diff
@@ -0,0 +1,34 @@
...
+default_godebug=$(go version -m -json package-registry | jq -r '.Settings[] | select(.Key == "DefaultGODEBUG") | .Value')
+
+go version -m package-registry
+
+if [[ ! "${fips_version}" =~ ^v1\.0\.0(-[a-f0-9]{5,40})?$ ]]; then
+	echo "Expected GOFIPS140 to reference the certified v1.0.0 module, got: '${fips_version}'"
+	exit 1
+fi
+
+if [[ "${default_godebug}" != *"fips140=on"* ]]; then
```
mrodm: Should this be stricter ? 


```suggestion
if [[ "${default_godebug}" =~ ^fips140=on$ ]]; then
```

mrodm (commented):
Pulled the docker image locally created in this PR and checking the binary inside that image contain the expected build variables 👍 :

```
	build	DefaultGODEBUG=fips140=on
	build	GOFIPS140=v1.0.0-c2097c7c
```

Changes in `publish-distribution.sh` I think they should be postponed until the EPR release is published (along with the docker image). Or at least remove the new suffix from the array for now.

.buildkite/scripts/test-fips.sh
teresaromero: Updated the check to match fips140=on as a complete comma-separated DefaultGODEBUG entry. Exact whole-string equality would be too strict because Go may include additional defaults, while the previous substring check could accept values such as fips140=only.

main.go
teresaromero: Good point. I’ve updated the TLS configuration so FIPS binaries explicitly default to TLS 1.2 when -tls-min-version is omitted, rather than relying on Go’s default. Explicit values and non-FIPS behavior remain unchanged, with tests covering these cases.

.buildkite/scripts/publish-distribution.sh
teresaromero: i should move this change into an other PR that runs this when we do have the image?

mrodm: It could be kept the changes, but at least the third element of the array (`-fips`) should be moved out to another Pull Request.

As it is, I think it is going to fail to publish those, since it requires both:
- a new version of Package Registry to publish the FIPS EPR docker 
- a new distribution docker FIPS image that requires the EPR FIPS docker above.

teresaromero: i've pushed a change where we check if the source image exists during this transition, and conditionally triggers the copy and the fips. let me know what you think. i would like to keep it on this pr

.buildkite/release-package-registry-distribution.yml
```diff
@@ -68,6 +68,34 @@ steps:
...
+        steps:
+          - trigger: unified-release-copy-elastic-images-to-dockerhub
+            label: ":docker: Copy release $${tag_name} FIPS images to Docker Hub"
+            key: "$${step_key}"
+            async: false
+            build:
+              env:
+                IMAGES_NAMES: "package-registry/package-registry-distribution"
+                IMAGES_TAG: "$${image_tag}"
+        PIPELINE
```
mrodm, line 93: Just checking in the documentation how it works if multiple uploads are perform in the same step:

https://buildkite.com/docs/agent/cli/reference/pipeline#insertion-order

> Steps are inserted immediately following the job performing the pipeline upload. Note that if you perform multiple uploads from a single step, they can appear to be in reverse order, because the later uploads are inserted earlier in the pipeline.

So it looks like there should not be any problem.

This code could be simplified once the distribution fips image is available, to just add the new steps directly in the same `buildkite-agent pipeline upload` command. It would be good

mrodm: Another option to avoid those multiplo pipeline uploads and the new method is using `cat >>` instead.

Something like:

```sh
pipeline_file="$$(mktemp)"
trap "rm -f $${pipeline_file}" EXIT

cat >> "$${pipeline_file}" << PIPELINE
steps:
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release images to Docker Hub"
    key: copy-prod
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "$${DOCKER_TAG}"
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release UBI images to Docker Hub"
    key: copy-prod-ubi
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "$${DOCKER_TAG}-ubi"
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release lite images to Docker Hub"
    key: copy-lite
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "lite-$${DOCKER_TAG}"
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release lite UBI images to Docker Hub"
    key: copy-lite-ubi
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "lite-$${DOCKER_TAG}-ubi"
PIPELINE

if [[ "$$(buildkite-agent meta-data get "fips-distribution-production-available" --default="false")" == "true" ]]; then
    cat >> "$${pipeline_file}" << PIPELINE
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release FIPS images to Docker Hub"
    key: copy-prod-fips
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "$${DOCKER_TAG}-fips"
PIPELINE
fi

if [[ "$$(buildkite-agent meta-data get "fips-distribution-lite-available" --default="false")" == "true" ]]; then
    cat >> "$${pipeline_file}" << PIPELINE
  - trigger: unified-release-copy-elastic-images-to-dockerhub
    label: ":docker: Copy release lite FIPS images to Docker Hub"
    key: copy-lite-fips
    async: false
    build:
      env:
        IMAGES_NAMES: "package-registry/package-registry-distribution"
        IMAGES_TAG: "lite-$${DOCKER_TAG}-fips"
PIPELINE
fi

buildkite-agent pipeline upload "$${pipeline_file}"

```

.buildkite/scripts/test-fips.sh
```diff
@@ -0,0 +1,36 @@
...
+with_jq
+with_mage
+
+echo "--- Build FIPS 140-3 binary"
+mage -debug buildFIPS
+
+echo "--- Verify FIPS 140-3 build settings"
+# GOFIPS140 includes a commit hash suffix (e.g. v1.0.0-c2097c7c), not just v1.0.0.
+fips_version=$(go version -m -json package-registry | jq -r '.Settings[] | select(.Key == "GOFIPS140") | .Value')
+default_godebug=$(go version -m -json package-registry | jq -r '.Settings[] | select(.Key == "DefaultGODEBUG") | .Value')
```
mrodm: Optional: `go version` could be run just once:

```sh
build_info=$(go version -m -json package-registry)
fips_version=$(echo "${build_info}" | jq -r '.Settings[] | select(.Key == "GOFIPS140") | .Value')
default_godebug=$(echo "${build_info}" | jq -r '.Settings[] | select(.Key == "DefaultGODEBUG") | .Value')

```

internal/storage/fakestorage.go
```diff
@@ -74,17 +88,49 @@ func ClientNoAuth(server *fakestorage.Server) *storage.Client {
...
+	// preserves the stored Md5Hash, keeping subsequent InitialObjects
+	// insertion FIPS-safe (non-empty Md5Hash skips recomputation).
+	listResp, err := server.ListObjectsWithOptionsPaginated(FakePackageStorageBucketInternal, fakestorage.ListOptions{})
+	require.NoError(tb, err, "failed to list existing server objects")
+	existingAttrs := listResp.Objects
+	allObjects := make([]fakestorage.Object, 0, len(existingAttrs)+len(newObjects))
+	for _, attrs := range existingAttrs {
+		obj, err := server.GetObject(FakePackageStorageBucketInternal, attrs.Name)
+		require.NoError(tb, err, "failed to retrieve existing server object %q", attrs.Name)
+		allObjects = append(allObjects, obj)
```
mrodm: From the `Claude` analysis, maybe there could be some issues if there are more packages (objects) than what it is shown in the response. It would depend on the `page size` set at the server side since `ListOptions` has no explicit page size.

According to Claude, it should be continued until token is empty:

> If there are more objects than one page, `listResp.Objects` is a partial set and the remaining pages are silently dropped.
> The restarted server would be missing those objects. Should iterate until `listResp.Token` is empty:

```go
var existingAttrs []fakestorage.ObjectAttrs
var token string
for {
    listResp, err := server.ListObjectsWithOptionsPaginated(
        FakePackageStorageBucketInternal,
        fakestorage.ListOptions{Token: token},
    )
    require.NoError(tb, err)
    existingAttrs = append(existingAttrs, listResp.Objects...)
    if listResp.Token == "" {
        break
    }
    token = listResp.Token
}

```

.buildkite/scripts/publish.sh
mrodm: Nit. Just realized that this message could use the variables defined at lines 13 and 14:

```
	echo "Docker images pushed: ${docker_img_tag} ${docker_img_tag_branch}"
```

Even those variables are the ones used in the `docker buildx build` command

.buildkite/release-package-registry-distribution.yml
teresaromero: i've opened an issue to address this as a future cleanup fix https://github.com/elastic/package-registry/issues/1893 and left the code as is for now

mrodm (approved):
Thanks!

PR #1888, opened 2026-07-25
packages: expose iac_blueprints and provider_permissions in manifest structs

jeniawhite:
`iac_blueprints` and `provider_permissions`are being added to the package-spec
(https://github.com/elastic/package-spec/pull/1209, https://github.com/elastic/package-spec/pull/1180). Adding to the registry's Go manifest structs to include
them, so the fields would appear in the JSON API responses.

This adds the necessary Go types (`IaCBlueprint`, `ProviderPermission`,
`ProviderRole`, `PermissionEntry`) and wires them into all four levels where the
spec allows the fields:

- `Package` (package-level)
- `PolicyTemplate` (policy-template-level)
- `Input` (input-level, datastream.go)
- `DataStream` (data-stream-level, datastream.go)

All fields are omitempty so there is no impact on existing packages that don't
declare them.

### DIFF
--- a/packages/package.go
+++ b/packages/package.go
@@ -40,10 +40,10 @@ type Package struct {
 	BasePackage   `config:",inline" json:",inline" yaml:",inline"`
 	FormatVersion string `config:"format_version" json:"format_version" yaml:"format_version"`
 
-	Readme          *string               `config:"readme,omitempty" json:"readme,omitempty" yaml:"readme,omitempty"`
-	License         string                `config:"license,omitempty" json:"license,omitempty" yaml:"license,omitempty"`
-	Screenshots     []Image               `config:"screenshots,omitempty" json:"screenshots,omitempty" yaml:"screenshots,omitempty"`
-	Assets          []string              `config:"assets,omitempty" json:"assets,omitempty" yaml:"assets,omitempty"`
+	Readme              *string               `config:"readme,omitempty" json:"readme,omitempty" yaml:"readme,omitempty"`
+	License             string                `config:"license,omitempty" json:"license,omitempty" yaml:"license,omitempty"`
+	Screenshots         []Image               `config:"screenshots,omitempty" json:"screenshots,omitempty" yaml:"screenshots,omitempty"`
+	Assets              []string              `config:"assets,omitempty" json:"assets,omitempty" yaml:"assets,omitempty"`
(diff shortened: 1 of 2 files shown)


groman92 (approved):

packages/package.go
```diff
@@ -130,6 +132,40 @@ type PolicyTemplate struct {
...
+	ID          string `config:"id,omitempty" json:"id,omitempty" yaml:"id,omitempty"`
+	Description string `config:"description,omitempty" json:"description,omitempty" yaml:"description,omitempty"`
+}
+
+// PermissionEntry is an individual permission grant required by an integration unit.
+type PermissionEntry struct {
+	Name        string                 `config:"name" json:"name" yaml:"name" validate:"required"`
+	Description string                 `config:"description,omitempty" json:"description,omitempty" yaml:"description,omitempty"`
+	Resources   []string               `config:"resources,omitempty" json:"resources,omitempty" yaml:"resources,omitempty"`
+	Conditions  map[string]interface{} `config:"conditions,omitempty" json:"conditions,omitempty" yaml:"conditions,omitempty"`
```
bhapas, line 168: How do these Conditions look like? Can we type safe them? Like if spec defines them in a specific format then a typed struct probably is good.

bhapas (approved):
Minor question. LGTM otherwise

teresaromero (commented):
hi, thanks for the PR.

Couple comments:
- please add a changelog entry under [unreleased] and the relevant category
- as this is a new field optional, please provide some test fixtures with these fields, and regenerate the snapshots so the new fields are tested (testdata/packages)

you can find more here https://github.com/elastic/package-registry#development - let me know if you need further help

bhapas:
>please add a changelog entry under [unreleased] and the relevant category
as this is a new field optional, please provide some test fixtures with these fields, and regenerate the snapshots so the new fields are tested (testdata/packages)

@teresaromero 
Can the CI catch these stuff jsut so things don't get missed up in future?

teresaromero:
@bhapas 

>Can the CI catch these stuff jsut so things don't get missed up in future?

We do have some workflows in other repos to control this and we are working on improving contribution docs so external contributors have a better experience. For the time being, we catch this things via pr-review. Things get missed up all the time and this is not a bad thing, its expected.

PR #1923, opened 2026-08-10
feat: add incremental index updates for storage indexer

### LINKED ISSUE #1527: Add support to perform incremental updates of packages when updating the indexers
The goal of this issue is adding support to incremental updates by adding a new `delta` file that contains the latest changes introduced in the search index.

Currently, storage indexers read two JSON files (index and cursor) containing all the packages every time a package is added, delete or updated:
| File Path | Purpose |
| --- | --- |
| `v2/metadata/<timestamp>/search-index-all.json` | Full package list (used for every update in package-registry) |
| `v2/metadata/cursor.json` | Points to latest search index created |

The new implementation will add a new `delta` file:

| File Path | New file | Purpose |
| --- | --- | --- |
| `v2/metadata/<timestamp>/search-index-all.json` | No |  Full package list (for recovery/init) |
| `v2/metadata/<timestamp>/search-index-delta.json` | Yes | Delta changes (added/updated/deleted) |
| `v2/metadata/cursor.json` | No | Points to latest search index created |

The delta will contain packages that have been added, deleted, or updated in that operation. And the index in memory should be updated accordingly.

Example of this delta file:
```json
{
  "added": [
    { "name": "packageA", "version": "1.0.0", ... }
  ],
  "updated": [
    { "name": "packageB", "version": "2.1.0", ... }
  ],
  "deleted": [
    { "name": "packageC", "version": "1.0.0" }
  ]
}
```

The Package Registry must handle two scenarios:
- The service is (re)started and does not have any package from the remote index.
    - It must load all the packages from the search index (current logic).
- The service is already running with a set of packages loaded in memory.
    - It must read new delta files and perform only the required operations.

For the first scenario (initial service state as well as current approach):
1. Read `v2/metadata/cursor.json` to get the latest timestamp (cursor).
2. Read the search-index pointed by the cursor (`search-index-all.json`).
3. Load all the packages from that index.
4. Update its own cursor of last processed search index.

For the second scenario (incremental update):
- Service already has a cursor pointing to a search index from previous scenario.
- On each interval:
    1. Read `v2/metadata/cursor.json` to get the latest cursor updated.
        - If it is the same cursor, skip iteration and wait for the next interval execution.
    2. List all folders in `v2/metadata/` with timestamps newer than the last processed.
    3. For each folder (in order):
        1. Read and apply `search-index-delta.json` to update local state (packages).
        2. Optionally, use `search-index-all.json` if a full sync is required.
    4. Update its cursor with the last processed search index.

## Requirements
- Package Registry service should be able to keep reading the full search index to replace all packages:
    - option to ensure consistency
    - required when the service starts from scratch or it is restarted.
- When applying multiple delta files is important to apply them in the same order they were created (e.g. according to their timestamps/cursors). 
- This new feature should be behind a feature flag, so it keeps by default the current behavior (updating all packages).

## Checklist
- [ ] Package registry should be able to update the list of packages with the new packages.
- [ ] Package registry should be able to update the list of packages removing the required ones.
- [ ] Package registry should be able to read and apply all the delta files between the cursor stored and the one retrieved in the same order they were created.
- [ ] Package registry should be able to do a full synchronization of packages.
- [ ] Added feature flag to enable this feature. By default, it should keep the same logic as it has now.

teresaromero:
Resolves https://github.com/elastic/package-registry/issues/1527

## Summary

Adds support for incremental package index updates in the in-memory `storage.Indexer`.
Instead of downloading and replacing the full `search-index-all.json` on every poll
cycle, the indexer can now apply a series of small `search-index-delta.json` files
that describe only what changed (packages added, updated, or removed).

The feature is off by default. Existing behaviour is fully preserved.

### New flag

```
--feature-incremental-updates   (env: EPR_FEATURE_INCREMENTAL_UPDATES)
```

Default `false`. When `true`, the storage indexer uses the incremental path for every
poll cycle after the first full sync on startup. A warning is logged at startup when
the flag is enabled so it always appears in container logs.

### New GCS file: `v2/metadata/<revision>/search-index-delta.json`

```json
{
  "added":   [{ ...full package... }],
  "updated": [{ ...full package... }],
  "removed": [{ "name": "pkg", "version": "1.0.0" }]
}
```

`updated` entries must be the **complete** package representation, not a partial patch —
`applyDelta` does a full replacement by key, not a field merge.

### Poll flow

```
startup / cursor == ""     → full sync from search-index-all.json  (unchanged)
flag off                   → full sync from search-index-all.json  (unchanged)

flag on, cursor set:
  read cursor.json         → latest revision
  cursor unchanged         → skip (unchanged)
  list v2/metadata/ folders between current cursor and latest (ascending)
  for each revision:
    delta file present     → applyDelta (add / update / remove entries in memory)
    delta file missing     → warn + fall back to full sync for that revision
  update cursor
```

### `applyDelta` implementation

Rather than building a full `map[string]*Package` over all N in-memory packages
(the obvious approach, which causes a 3× transient memory peak), `applyDelta`:

1. Builds two small maps sized to the delta entries only (`removeKeys`, `updateMap`).
2. Does a single pass over the existing `packageList`, writing into a fresh backing
   array (allocated at the same capacity), filtering out removed packages and
   applying updates.
3. Appends added packages at the end of the new backing array.

Peak memory during a delta application is ~2× steady-state at peak (old array read,
new array being built), avoiding the full 3× (old slice + full map + new slice) of
the naïve approach.

### Out of scope

`SQLIndexer` is not changed — it uses a full DB-swap strategy and incremental updates
would require separate database-level operations. It continues to do full-swap loads
regardless of the flag.

### Unit tests

```bash
go test ./internal/storage/... ./storage/... -run TestIncrementalUpdate -v
```

10 sub-tests covering: add, remove, update, update is full replacement (not a field
merge), multiple deltas in order, flag-off falls back to full sync, same cursor skips,
startup always full syncs, missing delta falls back to full sync for that revision,
concurrent get-and-apply-delta (race detector clean).

### Benchmarks — full sync vs incremental, against the full 1139-package index

```bash
go test ./storage/... -run='^$' -bench='^BenchmarkIndexerUpdateIndex' -benchmem -benchtime=5s
```

Results on Apple M4 Pro:

```
BenchmarkIndexerUpdateIndex-12                28   202 ms/op   101 MB/op   482 K allocs/op
BenchmarkIndexerUpdateIndex_Incremental-12  25926   574 µs/op   137 KB/op     815 allocs/op
```

**~351× faster and ~738× less memory per poll cycle.**

The 815 allocs/op in the incremental case are dominated by the fake-GCS HTTP
round-trip and JSON decode. `applyDelta` itself contributes only a handful of
allocs (two small maps + the delta entries), scaling with delta size rather
than index size.

### End-to-end with Docker (fake GCS server)

Started the fake GCS server via `dev/docker-compose-gcs.yml` and ran the registry
locally with:

```bash
STORAGE_EMULATOR_HOST=localhost:4443 \
EPR_FEATURE_STORAGE_INDEXER=true \
EPR_FEATURE_INCREMENTAL_UPDATES=true \
EPR_STORAGE_INDEXER_BUCKET_INTERNAL="gs://fake-package-storage-internal" \
EPR_STORAGE_INDEXER_WATCH_INTERVAL=10s \
EPR_CONFIG=config.docker.yml \
go run .
```

Delta files were pushed via the fake GCS JSON API and verified against the `/search`
endpoint within the watch interval — no restart required. All three delta operations
(add, remove, update) produced the correct response.

### O(N²) → O(N) in `Filter.Apply` and `legacyApply`

The "keep only the latest version per package name" dedup pass inside `Filter.Apply`
(and `legacyApply` for legacy clients) previously scanned the growing result slice with
an inner loop for every incoming package — O(N²) over the number of distinct package
names, hit on every API request.

Replaced with a `map[string]int` (name → index in result slice) so each lookup is O(1)
and the full pass is O(N). Also pre-sized the result slices in `Apply`, `legacyApply`,
and `filterCategories` to eliminate grow-reallocations.

Benchmarks (Apple M4 Pro, `-count=6`, `benchstat` p=0.002):

| Scale | Before | After | Change |
|---|---|---|---|
| 100 package names × 5 versions | 42 µs | 11 µs | −74% |
| 500 × 5 | 1,059 µs | 57 µs | −95% |
| 1,000 × 5 | 4,359 µs | 110 µs | −97% |
| 2,000 × 5 | 18,754 µs | 220 µs | −99% |

The widening gap confirms the quadratic behaviour of the old code: 10× more package
names cost ~100× more CPU. The new code scales linearly.

## Known limitations / open questions

- **`SQLIndexer`**: incremental updates are out of scope for this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add incremental index updates for the storage indexer via `--feature-incremental-updates` / `EPR_FEATURE_INCREMENTAL_UPDATES`. When enabled, poll cycles after the initial full sync apply `search-index-delta.json` files instead of re-downloading the full index, significantly reducing per-cycle memory and CPU usage. [#1923](https://github.com/elastic/package-registry/pull/1923)
+* Improve performance of `Filter.Apply` and `legacyApply` for large package lists by replacing the latest-version dedup pass with a map-based lookup. [#1923](https://github.com/elastic/package-registry/pull/1923)
+
 ### Deprecated
 
 ### Known Issues
--- a/internal/storage/cursors.go
+++ b/internal/storage/cursors.go
@@ -1,0 +1,59 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package storage
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+// listCursorsBetween returns all timestamp folder names in v2/metadata/ that are
+// strictly greater than since and less than or equal to until, in ascending order.
+func listCursorsBetween(ctx context.Context, storageClient *storage.Client, bucketName, rootStoragePath, since, until string) ([]string, error) {
+	prefix := joinObjectPaths(rootStoragePath, v2MetadataStoragePath) + "/"
+
+	query := &storage.Query{
+		Prefix:    prefix,
+		Delimiter: "/",
+	}
+
+	it := storageClient.Bucket(bucketName).Objects(ctx, query)
+
+	var timestamps []string
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error listing cursor folders: %w", err)
+		}
+		if attrs.Prefix == "" {
+			continue
+		}
+		token := strings.TrimSuffix(strings.TrimPrefix(attrs.Prefix, prefix), "/")
+		if token > since && token <= until {
+			timestamps = append(timestamps, token)
+		}
+	}
+
+	sort.Strings(timestamps)
+	return timestamps, nil
+}
+
+// ListCursorsBetween returns all timestamp folder names in v2/metadata/ strictly
+// greater than since and less than or equal to until, in ascending order.
+func ListCursorsBetween(ctx context.Context, storageClient *storage.Client, storageBucketInternal, since, until string) ([]string, error) {
+	bucketName, rootStoragePath, err := extractBucketNameFromURL(storageBucketInternal)
+	if err != nil {
+		return nil, fmt.Errorf("can't extract bucket name from URL: %w", err)
+	}
+	return listCursorsBetween(ctx, storageClient, bucketName, rootStoragePath, since, until)
+}
--- a/internal/storage/cursors.go
+++ b/internal/storage/cursors.go
@@ -16,6 +16,9 @@ import (
 
 // listCursorsBetween returns all timestamp folder names in v2/metadata/ that are
 // strictly greater than since and less than or equal to until, in ascending order.
+// Cursor values must be lexicographically comparable in chronological order (e.g.
+// zero-padded or fixed-length strings such as Unix epoch seconds). Simple unpadded
+// integer strings break ordering once they reach a second digit ("9" > "10").
 func listCursorsBetween(ctx context.Context, storageClient *storage.Client, bucketName, rootStoragePath, since, until string) ([]string, error) {
 	prefix := joinObjectPaths(rootStoragePath, v2MetadataStoragePath) + "/"
 
--- a/internal/storage/cursors_test.go
+++ b/internal/storage/cursors_test.go
@@ -62,17 +62,6 @@ func TestListCursorsBetween_MultipleIntermediates(t *testing.T) {
 	assert.Equal(t, []string{"2", "3", "4"}, timestamps)
 }
 
-func TestListCursorsBetween_ExcludesSince(t *testing.T) {
-	server := newCursorsServer(t, "1", "2", "3")
-	client := ClientNoAuth(server)
-
-	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "3")
-	require.NoError(t, err)
-	assert.NotContains(t, timestamps, "1")
-	assert.Contains(t, timestamps, "2")
-	assert.Contains(t, timestamps, "3")
-}
-
 // TestListCursorsBetween_FixedLengthSort verifies that multi-character cursor values
 // sort correctly. listCursorsBetween uses lexicographic ordering, which requires
 // cursor values to be fixed-length strings (e.g. zero-padded integers or Unix
--- a/internal/storage/cursors_test.go
+++ b/internal/storage/cursors_test.go
@@ -81,3 +81,16 @@ func TestListCursorsBetween_IncludesUntil(t *testing.T) {
 	require.NoError(t, err)
 	assert.Contains(t, timestamps, "3")
 }
+
+// TestListCursorsBetween_FixedLengthSort verifies that multi-character cursor values
+// sort correctly. listCursorsBetween uses lexicographic ordering, which requires
+// cursor values to be fixed-length strings (e.g. zero-padded integers or Unix
+// epoch timestamps). This test uses zero-padded values to document that assumption.
+func TestListCursorsBetween_FixedLengthSort(t *testing.T) {
+	server := newCursorsServer(t, "09", "10", "11", "12")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "09", "12")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"10", "11", "12"}, timestamps)
+}
--- a/internal/storage/cursors_test.go
+++ b/internal/storage/cursors_test.go
@@ -1,0 +1,83 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package storage
+
+import (
+	"testing"
+
+	"github.com/fsouza/fake-gcs-server/fakestorage"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func revisionObject(revision string) fakestorage.Object {
+	return fakestorage.Object{
+		ObjectAttrs: fakestorage.ObjectAttrs{
+			BucketName: FakePackageStorageBucketInternal,
+			Name:       joinObjectPaths(v2MetadataStoragePath, revision, searchIndexDeltaFile),
+			Md5Hash:    fakeObjectMD5Hash,
+		},
+		Content: []byte(`{}`),
+	}
+}
+
+func newCursorsServer(t *testing.T, revisions ...string) *fakestorage.Server {
+	t.Helper()
+	objects := make([]fakestorage.Object, 0, len(revisions))
+	for _, r := range revisions {
+		objects = append(objects, revisionObject(r))
+	}
+	server := fakestorage.NewServer(objects)
+	t.Cleanup(server.Stop)
+	return server
+}
+
+func TestListCursorsBetween_EmptyRange(t *testing.T) {
+	server := newCursorsServer(t, "2", "3")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "3", "3")
+	require.NoError(t, err)
+	assert.Empty(t, timestamps)
+}
+
+func TestListCursorsBetween_SingleIntermediate(t *testing.T) {
+	server := newCursorsServer(t, "1", "2", "3")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "2")
+	require.NoError(t, err)
+	require.Len(t, timestamps, 1)
+	assert.Equal(t, "2", timestamps[0])
+}
+
+func TestListCursorsBetween_MultipleIntermediates(t *testing.T) {
+	server := newCursorsServer(t, "1", "2", "3", "4", "5")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "4")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"2", "3", "4"}, timestamps)
+}
+
+func TestListCursorsBetween_ExcludesSince(t *testing.T) {
+	server := newCursorsServer(t, "1", "2", "3")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "3")
+	require.NoError(t, err)
+	assert.NotContains(t, timestamps, "1")
+	assert.Contains(t, timestamps, "2")
+	assert.Contains(t, timestamps, "3")
+}
+
+func TestListCursorsBetween_IncludesUntil(t *testing.T) {
+	server := newCursorsServer(t, "1", "2", "3")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "3")
+	require.NoError(t, err)
+	assert.Contains(t, timestamps, "3")
+}
--- a/internal/storage/cursors_test.go
+++ b/internal/storage/cursors_test.go
@@ -73,15 +73,6 @@ func TestListCursorsBetween_ExcludesSince(t *testing.T) {
 	assert.Contains(t, timestamps, "3")
 }
 
-func TestListCursorsBetween_IncludesUntil(t *testing.T) {
-	server := newCursorsServer(t, "1", "2", "3")
-	client := ClientNoAuth(server)
-
-	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "3")
-	require.NoError(t, err)
-	assert.Contains(t, timestamps, "3")
-}
-
 // TestListCursorsBetween_FixedLengthSort verifies that multi-character cursor values
 // sort correctly. listCursorsBetween uses lexicographic ordering, which requires
 // cursor values to be fixed-length strings (e.g. zero-padded integers or Unix
--- a/internal/storage/delta.go
+++ b/internal/storage/delta.go
@@ -1,0 +1,68 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package storage
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"cloud.google.com/go/storage"
+	"go.elastic.co/apm/v2"
+	"go.uber.org/zap"
+
+	"github.com/elastic/package-registry/packages"
+)
+
+// packageIndexEntry mirrors the package_manifest wrapper used by package-storage-infra
+// when writing added/updated entries to the delta file.
+type packageIndexEntry struct {
+	PackageManifest packages.Package `json:"package_manifest"`
+}
+
+// SearchIndexDelta holds the changes between two consecutive index revisions.
+// Field names match the JSON produced by package-storage-infra:
+//   - added/updated entries are wrapped in {"package_manifest": {...}}
+//   - removed entries use the key "removed" (not "deleted")
+type SearchIndexDelta struct {
+	Added   []packageIndexEntry `json:"added"`
+	Updated []packageIndexEntry `json:"updated"`
+	Removed []RemovedPackageRef `json:"removed"`
+}
+
+// RemovedPackageRef identifies a package to remove by name and version.
+type RemovedPackageRef struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+func loadSearchIndexDelta(ctx context.Context, logger *zap.Logger, storageClient *storage.Client, bucketName, rootStoragePath string, aCursor cursor) (*SearchIndexDelta, error) {
+	span, ctx := apm.StartSpan(ctx, "LoadSearchIndexDelta", "app")
+	defer span.End()
+
+	logger.Debug("load search-index-delta", zap.String("delta.file", searchIndexDeltaFile))
+
+	rootedPath := buildIndexStoragePath(rootStoragePath, aCursor, searchIndexDeltaFile)
+	objectReader, err := storageClient.Bucket(bucketName).Object(rootedPath).NewReader(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("can't read the delta file (path: %s): %w", rootedPath, err)
+	}
+	defer objectReader.Close()
+
+	var delta SearchIndexDelta
+	if err := json.NewDecoder(objectReader).Decode(&delta); err != nil {
+		return nil, fmt.Errorf("can't decode the delta file: %w", err)
+	}
+	return &delta, nil
+}
+
+// LoadSearchIndexDelta reads and parses the delta file for the given cursor value.
+func LoadSearchIndexDelta(ctx context.Context, logger *zap.Logger, storageClient *storage.Client, storageBucketInternal, cursorValue string) (*SearchIndexDelta, error) {
+	bucketName, rootStoragePath, err := extractBucketNameFromURL(storageBucketInternal)
+	if err != nil {
+		return nil, fmt.Errorf("can't extract bucket name from URL: %w", err)
+	}
+	return loadSearchIndexDelta(ctx, logger, storageClient, bucketName, rootStoragePath, cursor{Current: cursorValue})
+}
--- a/internal/storage/delta.go
+++ b/internal/storage/delta.go
@@ -29,11 +29,11 @@ type packageIndexEntry struct {
 type SearchIndexDelta struct {
 	Added   []packageIndexEntry `json:"added"`
 	Updated []packageIndexEntry `json:"updated"`
-	Removed []RemovedPackageRef `json:"removed"`
+	Removed []removedPackageRef `json:"removed"`
 }
 
-// RemovedPackageRef identifies a package to remove by name and version.
-type RemovedPackageRef struct {
+// removedPackageRef identifies a package to remove by name and version.
+type removedPackageRef struct {
 	Name    string `json:"name"`
 	Version string `json:"version"`
 }
--- a/internal/storage/fakestorage.go
+++ b/internal/storage/fakestorage.go
@@ -172,9 +172,9 @@ func PrepareServerObjects(revision string, indexContent []byte) ([]fakestorage.O
 	return serverObjects, len(index.Packages), nil
 }
 
-// PrepareDeltaServerObjects creates GCS objects for a delta file revision:
+// prepareDeltaServerObjects creates GCS objects for a delta file revision:
 // a cursor.json pointing to revision and a search-index-delta.json at that revision's path.
-func PrepareDeltaServerObjects(revision string, deltaContent []byte) []fakestorage.Object {
+func prepareDeltaServerObjects(revision string, deltaContent []byte) []fakestorage.Object {
 	return []fakestorage.Object{
 		{
 			ObjectAttrs: fakestorage.ObjectAttrs{
@@ -195,7 +195,7 @@ func PrepareDeltaServerObjects(revision string, deltaContent []byte) []fakestora
 // server and returning a new one with the previous objects plus the new revision's
 // delta file (no search-index-all.json for the new revision).
 func UpdateFakeServerWithDelta(tb testing.TB, server *fakestorage.Server, revision string, deltaContent []byte) (*fakestorage.Server, *storage.Client) {
-	newObjects := PrepareDeltaServerObjects(revision, deltaContent)
+	newObjects := prepareDeltaServerObjects(revision, deltaContent)
 
 	var existingAttrs []fakestorage.ObjectAttrs
 	var pageToken string
--- a/internal/storage/fakestorage.go
+++ b/internal/storage/fakestorage.go
@@ -171,3 +171,52 @@ func PrepareServerObjects(revision string, indexContent []byte) ([]fakestorage.O
 	})
 	return serverObjects, len(index.Packages), nil
 }
+
+// PrepareDeltaServerObjects creates GCS objects for a delta file revision:
+// a cursor.json pointing to revision and a search-index-delta.json at that revision's path.
+func PrepareDeltaServerObjects(revision string, deltaContent []byte) []fakestorage.Object {
+	return []fakestorage.Object{
+		{
+			ObjectAttrs: fakestorage.ObjectAttrs{
+				BucketName: FakePackageStorageBucketInternal, Name: cursorStoragePath, Md5Hash: fakeObjectMD5Hash,
+			},
+			Content: []byte(`{"current":"` + revision + `"}`),
+		},
+		{
+			ObjectAttrs: fakestorage.ObjectAttrs{
+				BucketName: FakePackageStorageBucketInternal, Name: joinObjectPaths(v2MetadataStoragePath, revision, searchIndexDeltaFile), Md5Hash: fakeObjectMD5Hash,
+			},
+			Content: deltaContent,
+		},
+	}
+}
+
+// UpdateFakeServerWithDelta simulates a delta update by stopping the given fake
+// server and returning a new one with the previous objects plus the new revision's
+// delta file (no search-index-all.json for the new revision).
+func UpdateFakeServerWithDelta(tb testing.TB, server *fakestorage.Server, revision string, deltaContent []byte) (*fakestorage.Server, *storage.Client) {
+	newObjects := PrepareDeltaServerObjects(revision, deltaContent)
+
+	var existingAttrs []fakestorage.ObjectAttrs
+	var pageToken string
+	for {
+		listResp, err := server.ListObjectsWithOptionsPaginated(FakePackageStorageBucketInternal, fakestorage.ListOptions{PageToken: pageToken})
+		require.NoError(tb, err, "failed to list existing server objects")
+		existingAttrs = append(existingAttrs, listResp.Objects...)
+		if listResp.NextPageToken == "" {
+			break
+		}
+		pageToken = listResp.NextPageToken
+	}
+	allObjects := make([]fakestorage.Object, 0, len(existingAttrs)+len(newObjects))
+	for _, attrs := range existingAttrs {
+		obj, err := server.GetObject(FakePackageStorageBucketInternal, attrs.Name)
+		require.NoError(tb, err, "failed to retrieve existing server object %q", attrs.Name)
+		allObjects = append(allObjects, obj)
+	}
+	allObjects = append(allObjects, newObjects...)
+
+	server.Stop()
+	newServer := fakestorage.NewServer(allObjects)
+	return newServer, ClientNoAuth(newServer)
+}
--- a/main.go
+++ b/main.go
@@ -400,6 +400,9 @@ func initIndexer(ctx context.Context, logger *zap.Logger, options serverOptions)
 		}
 		combined = append(combined, indexer)
 	case featureStorageIndexer:
+		if featureIncrementalUpdates {
+			logger.Warn("Technical preview: Incremental updates feature is enabled.")
+		}
 		indexer, err := initStorageIndexer(ctx, logger, options)
 		if err != nil {
 			logger.Fatal("failed to initialize storage indexer", zap.Error(err))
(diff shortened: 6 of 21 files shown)


internal/storage/cursors_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestListCursorsBetween_IncludesUntil(t *testing.T) {
+	server := newCursorsServer(t, "1", "2", "3")
+	client := ClientNoAuth(server)
+
+	timestamps, err := listCursorsBetween(t.Context(), client, FakePackageStorageBucketInternal, "", "1", "3")
+	require.NoError(t, err)
+	assert.Contains(t, timestamps, "3")
+}
```
mrodm: Would these two tests equivalent (or already tested by) to `TestListCursorsBetween_MultipleIntermediates` ?

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+		delta, err := internalStorage.LoadSearchIndexDelta(ctx, i.logger, i.storageClient, i.options.PackageStorageBucketInternal, ts)
+		if err != nil {
+			if errors.Is(err, storage.ErrObjectNotExist) {
+				i.logger.Warn("delta file missing, falling back to full sync for timestamp", zap.String("cursor", ts))
+				anIndex, err := internalStorage.LoadSearchIndexAllForCursor(ctx, i.logger, i.storageClient, i.options.PackageStorageBucketInternal, ts)
+				if err != nil {
+					metrics.StorageIndexerUpdateIndexErrorsTotal.Inc()
+					return fmt.Errorf("can't load search-index-all for cursor %s: %w", ts, err)
+				}
+				revisions = append(revisions, revision{timestamp: ts, fullIndex: anIndex})
```
mrodm: Would it be ok to just set one revision here with the full index ? 

```go
revisions = revisions[:0] // reset the slice to reuse the memory
revisions = append(revisions, revision{timestamp: ts, fullIndex: anIndex})
```

This will help also to avoid having in memory more than once the full index if several iterations of this loop fail reading the index-delta (I hope it should not happen).

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+	removeKeys := make(map[string]struct{}, len(delta.Removed))
+	for _, ref := range delta.Removed {
+		removeKeys[ref.Name+"-"+ref.Version] = struct{}{}
+	}
+
+	updateMap := make(map[string]*packages.Package, len(delta.Updated))
+	for _, entry := range delta.Updated {
+		p := entry.PackageManifest
+		p.BasePath = fmt.Sprintf("%s-%s.zip", p.Name, p.Version)
+		p.SetRemoteResolver(i.resolver)
```
mrodm, line 294: Could these fields be set before acquiring the lock ? to try to reduce the time spent in the lock.

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+			delete(updateMap, key)
+		} else {
+			out = append(out, p)
+		}
+	}
+
+	for _, entry := range delta.Added {
+		p := entry.PackageManifest
+		p.BasePath = fmt.Sprintf("%s-%s.zip", p.Name, p.Version)
+		p.SetRemoteResolver(i.resolver)
```
mrodm: Could these fields be set before acquiring the lock ? to try to reduce the time spent in the lock.

storage/indexer_test.go
```diff
@@ -403,3 +447,236 @@ func TestGet_IndexUpdated(t *testing.T) {
...
+
+		// revision "2": add 0.3.0
+		fs, indexer.storageClient = internalStorage.UpdateFakeServerWithDelta(t, fs, "2", readDeltaFile(t, "testdata/search-index-delta-add.json"))
+		// revision "3": remove 0.1.1
+		fs, indexer.storageClient = internalStorage.UpdateFakeServerWithDelta(t, fs, "3", readDeltaFile(t, "testdata/search-index-delta-remove.json"))
+		// revision "4": update 0.2.0 title — cursor.json now points to "4"
+		_, indexer.storageClient = internalStorage.UpdateFakeServerWithDelta(t, fs, "4", readDeltaFile(t, "testdata/search-index-delta-update.json"))
+
+		// single updateIndex call should apply all three deltas
+		err = indexer.updateIndex(t.Context())
```
mrodm, line 662: It would be interesting to add a new test just for `delta-update` where the same package contains different fields. To be sure that the package at the end contains the expected fields (fields that should be removed plus fields that should be added).

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
 	return nil
 }
 
+func (i *Indexer) incrementalSync(ctx context.Context, latestCursorValue string) error {
+	timestamps, err := internalStorage.ListCursorsBetween(ctx, i.storageClient, i.options.PackageStorageBucketInternal, i.cursor, latestCursorValue)
+	if err != nil {
+		metrics.StorageIndexerUpdateIndexErrorsTotal.Inc()
+		return fmt.Errorf("can't list cursors between %s and %s: %w", i.cursor, latestCursorValue, err)
+	}
+
```
mrodm, line 213: It could be added here an `if` to return early when there are no new `revisions` found. IIUC this would happen when `revisions` is empty. 

If this can be applied, no need to run this calls:
```
	packages.UpdateLatestDeprecatedPackagesMapByName(i.packageList, i.deprecatedPackages)
	packages.PropagateLatestDeprecatedInfoToPackageList(i.packageList, i.deprecatedPackages)
```
since there should not be any changes there. Is that right?

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+			revisions = append(revisions, revision{timestamp: ts, delta: delta})
+		}
+	}
+
+	i.m.Lock()
+	defer i.m.Unlock()
+
+	for _, r := range revisions {
+		if r.fullIndex != nil {
+			i.transformSearchIndexAllToPackages(r.fullIndex)
```
mrodm: Could this be done before acquiring the lock ?

storage/indexer.go
mrodm: Here it must also be checked in case the index is nil as we do for the full sync currently:

https://github.com/elastic/package-registry/blob/2a50e2d26507d6f610bae38aabbd19f87986de45/storage/indexer.go#L172-L175

storage/indexer.go
```diff
@@ -179,18 +193,117 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+
+	updateMap := make(map[string]*packages.Package, len(delta.Updated))
+	for _, entry := range delta.Updated {
+		p := entry.PackageManifest
+		p.BasePath = fmt.Sprintf("%s-%s.zip", p.Name, p.Version)
+		p.SetRemoteResolver(i.resolver)
+		updateMap[p.Name+"-"+p.Version] = &p
+	}
+
+	out := i.packageList[:0]
```
mrodm: It looks like (according to Claude) that this reuse of slice here, could cause some race between reads and writes.

---
```
 Goroutine A — caller                 Goroutine B — watcher tick
 ─────────────────────────────────    ──────────────────────────────────
 Get() acquires RLock
 returns i.packageList  →  caller
   now holds slice {ptr=0xABCD,
   len=50} pointing at array X
 RLock released (defer runs)
                                      Lock() acquired  ← succeeds, A
                                                          no longer holds RLock
                                      applyDelta():
                                        out := i.packageList[:0]
                                          same ptr=0xABCD, array X
 for _, pkg := range result {           out[0] = newer  ← writes array X
     pkg.Name  ← reads array X    ←──── RACE: both touch array X
 }
```
The mutex only protects `i.packageList` — the slice header (pointer + len + cap) stored in the struct. Once Get returns, it hands the caller a copy of that header. The caller now has their own `{ptr, len, cap}` pointing at the same backing array, with no lock held.

Goroutine B's `Lock()` successfully excludes other goroutines from touching `i.packageList` (the struct field), but it does nothing to prevent Goroutine A from reading the raw memory at ptr+offset that `applyDelta` is simultaneously writing.


The lock used wrapping these methods (Read+Write) would protect you only if the caller read everything it needed while still holding `RLock`. Since `Get` releases it before the caller can iterate the result, the window is open.


In `main`, this is not happening since every update swaps the slice header to point at a completely different backing array. So any caller holding the old slice from a previous Get call is reading from memory that is never touched again — no writes, no race.


--- 

Claude suggests to change this line by:
```go
out := make(packages.Packages, 0, len(i.packageList))
```

It would be good to benchmark with this change (to avoid this kind of issues) and see the numbers using if possible indices (JSON files) with the real number of packages similar to production.

As `i.packagesList` is a slice of pointers, probably (I hope) there should not be too much difference, but worth to run benchmarks.

Depending on the results, we would need to check other approaches here.

internal/storage/cursors.go
```diff
@@ -0,0 +1,59 @@
...
+		if attrs.Prefix == "" {
+			continue
+		}
+		token := strings.TrimSuffix(strings.TrimPrefix(attrs.Prefix, prefix), "/")
+		if token > since && token <= until {
+			timestamps = append(timestamps, token)
+		}
+	}
+
+	sort.Strings(timestamps)
```
mrodm, line 50: Are all the timestamp values of the same length ?

If they are not , there could be the case that this happens `"9" > "10"`

it would be good to double check this.

internal/storage/delta.go
```diff
@@ -0,0 +1,68 @@
...
+	Added   []packageIndexEntry `json:"added"`
+	Updated []packageIndexEntry `json:"updated"`
+	Removed []RemovedPackageRef `json:"removed"`
+}
+
+// RemovedPackageRef identifies a package to remove by name and version.
+type RemovedPackageRef struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
```
mrodm: Could this struct be set as unexported? It looks like it is not used outside of `internal/storage` but its fields (Name and Version).

internal/storage/fakestorage.go
```diff
@@ -171,3 +171,52 @@ func PrepareServerObjects(revision string, indexContent []byte) ([]fakestorage.O
 	})
 	return serverObjects, len(index.Packages), nil
 }
+
+// PrepareDeltaServerObjects creates GCS objects for a delta file revision:
+// a cursor.json pointing to revision and a search-index-delta.json at that revision's path.
+func PrepareDeltaServerObjects(revision string, deltaContent []byte) []fakestorage.Object {
```
mrodm: Could this function be set as private/unexported?

It looks like it is just used here in this file.

storage/indexer_test.go
```diff
@@ -403,3 +447,236 @@ func TestGet_IndexUpdated(t *testing.T) {
...
+		err := indexer.Init(t.Context())
+		require.NoError(t, err)
+
+		_, indexer.storageClient = internalStorage.UpdateFakeServer(t, fs, "2", "testdata/search-index-all-full.json")
+		err = indexer.updateIndex(t.Context())
+		require.NoError(t, err)
+
+		foundPackages, err := indexer.Get(t.Context(), &packages.GetOptions{})
+		require.NoError(t, err)
+		assert.Greater(t, len(foundPackages), 2, "full sync should have loaded all packages")
```
mrodm: Why not using here `assert.Equal` to validate the exact number of packages loaded? or even `assert.Len(t, foundPackages, ..)` ?

storage/indexer_test.go
```diff
@@ -403,3 +447,236 @@ func TestGet_IndexUpdated(t *testing.T) {
...
+		t.Cleanup(func() { indexer.Close(context.Background()) })
+
+		err := indexer.Init(t.Context())
+		require.NoError(t, err)
+
+		// revision "2": add 0.3.0
+		fs, indexer.storageClient = internalStorage.UpdateFakeServerWithDelta(t, fs, "2", readDeltaFile(t, "testdata/search-index-delta-add.json"))
+		// revision "3": remove 0.1.1
+		fs, indexer.storageClient = internalStorage.UpdateFakeServerWithDelta(t, fs, "3", readDeltaFile(t, "testdata/search-index-delta-remove.json"))
+		// revision "4": update 0.2.0 title — cursor.json now points to "4"
```
mrodm, line 658: Should it be added a check in each of tests to ensure that the cursor is pointing to the right value ? It could be tested via `LoadLatestCursorValue` ?

main.go
```diff
@@ -129,6 +130,8 @@ func init() {
 	flag.DurationVar(&storageIndexerWatchInterval, "storage-indexer-watch-interval", 1*time.Minute, "Address of the package-registry service.")
 	// The following storage related flags are technical preview and might be removed in the future or renamed
 	flag.BoolVar(&featureSQLStorageIndexer, "feature-sql-storage-indexer", false, "Enable SQL storage indexer to include packages from Package Storage v2 (technical preview).")
+	flag.BoolVar(&featureIncrementalUpdates, "feature-incremental-updates", false,
+		"Enable incremental index updates using delta files (technical preview).")
```
mrodm: Add a reference here about being is not supported in the SQL indexer.

mrodm: It would be good to add some warning log line while the storage indexer is being initialized indicating that this new feature is enabled.

Something like this: 
https://github.com/elastic/package-registry/blob/2a50e2d26507d6f610bae38aabbd19f87986de45/main.go#L393

Or maybe like these ones (but they are for the `fsOptions`), but I would set it as Warning level to show always that message in the container logs:

https://github.com/elastic/package-registry/blob/2a50e2d26507d6f610bae38aabbd19f87986de45/main.go#L414-L415

Unrelated to this PR: Maybe should those debug lines be updated to be Warning too ? And just show when `fsOptions.EnablePathsWtcher` is true ?

mrodm (commented):
The approach looks really good!
I added a first round of comments, I want to also spend some time by doing some local tests with these changes.

A part from these comments, this change also requires a new changelog entry.

storage/indexer.go
teresaromero: Fixed. Changed `out := i.packageList[:0]` to `out := make(packages.Packages, 0, len(i.packageList))` so `applyDelta` builds into a fresh backing array. The caller's slice from `Get()` no longer shares the same backing memory. Added a `concurrent_get_and_apply_delta` sub-test that runs `Get()` and `UpdateIndex()` concurrently and passes cleanly under `-race`.

storage/indexer.go
teresaromero: Fixed. Added `revisions = revisions[:0]` before appending the full-sync fallback result, so if multiple revisions fall back in the same loop only the latest full index is held in memory. There's also a comment explaining why: "Full sync supersedes all prior deltas — reset to avoid holding multiple full copies in memory."

internal/storage/cursors.go
teresaromero: Added a doc comment to `listCursorsBetween` documenting the fixed-length requirement. GCS revision timestamps are Unix epoch seconds (always 10 digits), so the constraint holds in practice. Also added `TestListCursorsBetween_FixedLengthSort` with zero-padded two-digit values to guard the assumption and make it explicit for future readers.

storage/indexer.go
teresaromero: Added a nil guard after `LoadSearchIndexAllForCursor` in the fallback branch, mirroring the pattern at lines 172–175 in `fullSync`. Also changed `else { applyDelta(r.delta) }` to `else if r.delta != nil { ... }` as a defensive guard — `loadSearchIndexAll` currently always returns a non-nil pointer so the path is unreachable in practice, but the check is cheap and prevents a panic if that ever changes.

storage/indexer.go
teresaromero: Added an early return immediately after the `revisions` loop when `len(revisions) == 0`, before the `UpdateLatestDeprecatedPackagesMapByName` / `PropagateLatestDeprecatedInfoToPackageList` calls. An empty `timestamps` from `listCursorsBetween` produces an empty `revisions` slice, so the guard covers both the "nothing to do" and "cursor unchanged" fast paths.

storage/indexer_test.go
teresaromero: Added `update_is_full_replacement` sub-test. It seeds the indexer with a package at version 1.0.0 with description "original", then applies a delta with the same package/version but description "updated". The assertion checks the result contains exactly the new description, confirming full replacement rather than a field merge.

storage/indexer_test.go
teresaromero: After each delta operation in the affected sub-tests, the test now asserts `indexer.cursor` equals the expected revision timestamp directly (the cursor field is accessible within the package). The assertions cover all delta sub-tests including the new ones added in this round.

main.go
teresaromero: Added `logger.Warn("Technical preview: Incremental updates feature is enabled.")` inside the `featureStorageIndexer` case in `main.go`, gated on `featureIncrementalUpdates`.

storage/indexer.go
teresaromero: Moved the `transformSearchIndexAllToPackages` calls outside the lock in `incrementalSync`. The function only reads `i.resolver`, which is set once at init and never mutated, so it's safe to call without holding the mutex. The lock now only covers the final field assignments (`i.packageList`, `i.cursor`).

storage/indexer.go
teresaromero: same https://github.com/elastic/package-registry/pull/1923#discussion_r3765218044

storage/indexer.go
teresaromero: same https://github.com/elastic/package-registry/pull/1923#discussion_r3765218044

internal/storage/cursors_test.go
teresaromero: Looked at all three. `TestListCursorsBetween_ExcludesSince` was the redundant one: `MultipleIntermediates` already proves since-exclusion via exact equality (`assert.Equal(t, []string{"2","3","4"}, timestamps)`), and `ExcludesSince` only added weaker `Contains`/`NotContains` checks over the same invariant. Deleted `ExcludesSince`. `IncludesUntil` had been removed in a prior commit; the remaining tests — `EmptyRange`, `SingleIntermediate`, `MultipleIntermediates`, `FixedLengthSort` — each cover a distinct scenario.

storage/indexer.go
```diff
@@ -228,6 +228,11 @@ func (i *Indexer) incrementalSync(ctx context.Context, latestCursorValue string)
 					metrics.StorageIndexerUpdateIndexErrorsTotal.Inc()
 					return fmt.Errorf("can't load search-index-all for cursor %s: %w", ts, err)
 				}
+				if anIndex == nil {
+					i.logger.Info("Delta fallback: full sync returned no packages for revision.", zap.String("cursor", ts))
+				}
```
mrodm: Should it be added a `continue` inside this `if` when `anIndex` is nil?

If it is allowed to continue, all the previous deltas will be removed and the result will not be correct.

storage/indexer.go
mrodm: These loops are still inside the lock. Could they be moved outside of the lock?

storage/indexer.go
mrodm: These assignments for `BasePath` and `SetRemoteResolver` have not been moved outside of the lock as was done for the `fullSync`. 

Maybe it does not provide much improvement?

storage/indexer_test.go
```diff
@@ -550,6 +554,50 @@ func TestIncrementalUpdate(t *testing.T) {
...
+		require.Contains(t, byVersion, "0.2.0")
+		require.Contains(t, byVersion, "0.1.1")
+
+		updated := byVersion["0.2.0"]
+		require.NotNil(t, updated.Title)
+		assert.Equal(t, "1Password Events Reporting UPDATED DELTA", *updated.Title, "0.2.0 title must reflect the delta, not the original")
+
+		untouched := byVersion["0.1.1"]
+		require.NotNil(t, untouched.Title)
+		assert.Equal(t, "1Password Events Reporting", *untouched.Title, "0.1.1 must not be affected by the update delta")
```
mrodm, line 598: Nice!

If possible, I was also expecting to validate or test that given those packages (or other fields):
- A: without group field, with categories field
- B: with group field, without categories field

In the delta is set that A is going to be replaced by B.

Then the result should be exactly the fields present in B:
- no categories field
- group field is present

storage/indexer.go
teresaromero: Before (prior branch):
```
BenchmarkIndexerUpdateIndex-12                28   202 ms/op   101 MB/op   482 K allocs/op
BenchmarkIndexerUpdateIndex_Incremental-12  25926   574 µs/op   137 KB/op     815 allocs/op
```

After (this branch):
```
BenchmarkIndexerUpdateIndex-12                31   185 ms/op   101 MB/op   482 K allocs/op
BenchmarkIndexerUpdateIndex_Incremental-12  27216   496 µs/op   412 KB/op     817 allocs/op
```

Note: the incremental B/op is higher (137 → 412 KB) — expected, since make now allocates a fresh 1139-pointer backing array per call instead of reusing the existing one. That's the intentional trade-off for correctness. Alloc count is essentially unchanged (815 → 817), and throughput improved slightly (574 → 496 µs).

storage/indexer.go
```diff
@@ -280,7 +295,7 @@ func (i *Indexer) applyDelta(delta *internalStorage.SearchIndexDelta) {
 		updateMap[p.Name+"-"+p.Version] = &p
 	}
 
-	out := i.packageList[:0]
+	out := make(packages.Packages, 0, len(i.packageList))
```
mrodm: Thinking that here it could be considered also `len(delta.Added)` to set the capacity of the slice considering the ones that are going to be added.

Probably, it should be considered to remove from the capacity the packages that are going to be removed.

storage/indexer.go
mrodm, line 270: Just thought about this function `UpdateLatestDeprecatedPackagesMapByName`.

If a package that was marked as deprecated is removed from the index (in both full sync and incremental sync), those packages are kept in `i.deprecatedPackages` forever.

Let's consider this scenario:
- package A is published as deprecated
- package A is removed from the registry

In that scenario, all the packages with the same package name as A would be still be deprecated, even if the one that was marked as deprecated was removed.

Is my assumption correct?

mrodm: And according to claude there could be also a race condition here:

https://github.com/elastic/package-registry/blob/9657de1cfc12670f8dfc57493fa1b26f3d57a7db/packages/deprecated.go#L49-L53

Similar to what happened in https://github.com/elastic/package-registry/pull/1923#discussion_r3759100099

mrodm: One of the options mentioned by Claude was this one

```go
func PropagateLatestDeprecatedInfoToPackageList(packageList Packages, deprecatedPackages DeprecatedPackages) {
    for i, pkg := range packageList {
        if deprecatedInfo, found := deprecatedPackages.Deprecated(pkg.Name); found {
            pkgCopy := *pkg                  // shallow copy of the struct
            pkgCopy.Deprecated = deprecatedInfo
            packageList[i] = &pkgCopy       // new pointer, old one untouched
        }
    }
}
```


Please, could you double check this too ?

storage/indexer.go
mrodm: I stand correct on this, in your Pull Request this code was refactored and I think this `LoadSearchIndexAllForCursor` function could not return `anIndex` as nil. Is that right? I assumed that the same logic remain to calculate the index.

Maybe it can be removed this `if` completely.


Given that, it looks like that this other function `LoadPackagesAndCursorFromIndex` (from internal/storage/index.go) would have no callers , should it be deleted it or at least marked as deprecated?

storage/indexer.go
teresaromero: You're right — the loops for both `Updated` and `Added` were still running inside the lock. Both only read `delta.*` and `i.resolver`, which is read-only after init, so they can safely run without holding the write mutex.

Addressed by splitting `applyDelta` into two stages.

storage/indexer.go
teresaromero: Good catch — addressed together with https://github.com/elastic/package-registry/pull/1923#discussion_r3765393220. The `BasePath` assignment and `SetRemoteResolver` call for `Added` entries are now part of `prepareDelta`, which runs before the lock is acquired, matching the pattern already in place for full-index revisions.

storage/indexer.go
teresaromero: You're right on both counts.

`loadSearchIndexAll` (the internal function called by `LoadSearchIndexAllForCursor`) has two return paths: `(nil, err)` on error, and `(&packages, nil)` on success — the non-error path always returns a non-nil pointer. So the `if anIndex == nil` guard is unreachable and has been removed.

`LoadPackagesAndCursorFromIndex` had no callers: `sqlindexer.go` uses `LoadPackagesAndCursorFromIndexBatches` (the batched variant) exclusively. The function has been deleted.

storage/indexer_test.go
teresaromero: Good call. The existing `update_is_full_replacement` sub-test only verified that a field changes value — it didn't catch the case where a field present in the original is absent from the delta entry.

Added `update_removes_absent_fields` sub-test: seeds 1password 0.2.0 with `categories: ["security"]` and no `group`, then applies a delta that has `group: "observability"` and no `categories`. The assertions check that `group` is set to `"observability"` and `categories` is empty — confirming that fields absent from the delta entry do not survive in the result.

storage/indexer.go
teresaromero: The capacity hint is now `max(0, len(i.packageList)-len(pd.removeKeys)) + len(pd.added)`, which accounts for both removals and additions. (`removeKeys` and `added` are the pre-processed equivalents of `delta.Removed` and `delta.Added` after the refactor in https://github.com/elastic/package-registry/pull/1923#discussion_r3765393220 and https://github.com/elastic/package-registry/pull/1923#discussion_r3765425417.) The `max(0, ...)` guard handles the edge case where the removal count exceeds the current list size.

storage/indexer.go
teresaromero: Your assumption is technically correct — the stale entry does persist in `i.deprecatedPackages`. However, after reviewing the deprecation RFC ([internal doc](https://docs.google.com/document/d/1daDzfoxlsPM-s9sMy0kizUcB2Ly91mUXORqw_uYy1Ys/edit)) and the tracking issue ([elastic/package-spec#227](https://github.com/elastic/package-spec/issues/227)), I believe this is actually the intended behaviour rather than a bug.

The RFC defines the deprecation lifecycle as follows (quoting the "Lifecycle of package deprecation" section directly):

- A package or feature is marked as deprecated.
- It remains available in the registry for installation and use.
- A maintenance period (usually one year) follows, during which the package remains available.
- After this period, **the package may be removed from the registry**.

Deprecation is a package-family signal, not a version-specific one. The RFC's Package Registry section explicitly states that the registry should "return the latest known deprecations in the registry response **of all versions** for each package, if any" — which is precisely what `UpdateLatestDeprecatedPackagesMapByName` + `PropagateLatestDeprecatedInfoToPackageList` implement.

When a deprecated version is removed from the registry while other versions remain, those remaining versions **should still appear as deprecated** — the removal of a specific version does not un-deprecate the package family. The RFC has no lifecycle step for reversing a deprecation; once a package is marked deprecated, it remains deprecated until the whole package is eventually removed.

When all versions of a deprecated package are removed, the stale entry in `deprecatedPackages` becomes harmless: `PropagateLatestDeprecatedInfoToPackageList` only iterates `i.packageList`, so if no packages with that name remain, the entry is never applied to anything.

storage/indexer.go
```diff
@@ -260,8 +262,8 @@ func (i *Indexer) incrementalSync(ctx context.Context, latestCursorValue string)
 	for _, r := range revisions {
 		if r.fullIndex != nil {
 			i.packageList = *r.fullIndex
-		} else if r.delta != nil {
-			i.applyDelta(r.delta)
+		} else if r.prepared != nil {
+			i.applyDelta(*r.prepared)
 		}
 		i.cursor = r.timestamp
 	}
```
mrodm, line 265: It would be interesting to add a final debug message as in the full sync like this (https://github.com/elastic/package-registry/blob/9657de1cfc12670f8dfc57493fa1b26f3d57a7db/internal/storage/sqlindexer.go#L242):

```
i.logger.Info("Downloaded new search-index-all index", zap.String("index.packages.size", fmt.Sprintf("%d", len(*anIndex))))
```

storage/indexer.go
```diff
@@ -295,27 +299,34 @@ func (i *Indexer) applyDelta(delta *internalStorage.SearchIndexDelta) {
...
-
-	for _, entry := range delta.Added {
-		p := entry.PackageManifest
-		p.BasePath = fmt.Sprintf("%s-%s.zip", p.Name, p.Version)
-		p.SetRemoteResolver(i.resolver)
-		out = append(out, &p)
-	}
-
+	out = append(out, pd.added...)
```
mrodm: Should we add here checks to ensure that the added package is not already present in the package list ?

It should not happen, but in one of my tests I did add two consecutive cursos adding the same package and the registry add the same package/version twice (and also in the responses).

storage/indexer.go
```diff
@@ -179,18 +193,143 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+		delta     *internalStorage.SearchIndexDelta
+		prepared  *preparedDelta
+		fullIndex *packages.Packages
+	}
+
+	revisions := make([]revision, 0, len(timestamps))
+	for _, ts := range timestamps {
+		delta, err := internalStorage.LoadSearchIndexDelta(ctx, i.logger, i.storageClient, i.options.PackageStorageBucketInternal, ts)
+		if err != nil {
+			if errors.Is(err, storage.ErrObjectNotExist) {
```
mrodm: In one of my local test, I did a mistake and one of the cursos had a JSON syntax error and that caused that no package was able to be updated.

This scenario moreover did not have any `search-index-all.json` file, so the go routine could never update the indices nor update the cursor.

What do you about doing this process of using the full synchronization for any error that could happen when reading/loading the index delta ?

storage/indexer.go
```diff
@@ -179,18 +193,143 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+		}
+		if newer, ok := pd.updateMap[key]; ok {
+			out = append(out, newer)
+			delete(pd.updateMap, key)
+		} else {
+			out = append(out, p)
+		}
+	}
+	out = append(out, pd.added...)
+	i.packageList = out
```
mrodm, line 336: would it be too verbose to show a debug line with the removed/updated/added packages?

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Add incremental index updates for the storage indexer via `--feature-incremental-updates` / `EPR_FEATURE_INCREMENTAL_UPDATES`. When enabled, poll cycles after the initial full sync apply `search-index-delta.json` files instead of re-downloading the full index, significantly reducing per-cycle memory and CPU usage. [#1923](https://github.com/elastic/package-registry/pull/1923)
+* Improve performance of `Filter.Apply` and `legacyApply` for large package lists by replacing the latest-version dedup pass with a map-based lookup. [#1923](https://github.com/elastic/package-registry/pull/1923)
```
mrodm, line 16: Please check these changes doing some requests with your local package-registry (using the latest index from production) and with the production one.

To be sure there is no regression there.

mrodm (commented):
Thanks for addressing all the comments!

I did a few tests locally, and added a few more comments related to some scenarios that I was checking. And some other tests that could be performed.

storage/indexer.go
```diff
@@ -313,19 +314,24 @@ func (i *Indexer) applyDelta(pd preparedDelta) {
...
+			i.logger.Debug("removed package", zap.String("package", key))
 			continue
 		}
 		if newer, ok := pd.updateMap[key]; ok {
 			out = append(out, newer)
 			delete(pd.updateMap, key)
+			i.logger.Debug("updated package", zap.String("package", key))
 		} else {
 			out = append(out, p)
+			i.logger.Debug("kept package", zap.String("package", key))
```
mrodm: I think this message could be dismissed (`kept package`). Currently, it would show in the output around 14000 messages when debug is enabled for the usual operations.

WDYT ?

CHANGELOG.md
teresaromero: I've done a test based on the ones we do when release, same query for both envs and i did not catch any difference. I noted some time reduced but i cant track how much is because of the improvement, or how much is because the local setup.

storage/indexer.go
```diff
@@ -179,18 +193,150 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
 
-	// set the deprecated notice information once the package list is updated
 	packages.UpdateLatestDeprecatedPackagesMapByName(i.packageList, i.deprecatedPackages)
 	packages.PropagateLatestDeprecatedInfoToPackageList(i.packageList, i.deprecatedPackages)
 
 	return nil
 }
 
+func (i *Indexer) incrementalSync(ctx context.Context, latestCursorValue string) error {
+	timestamps, err := internalStorage.ListCursorsBetween(ctx, i.storageClient, i.options.PackageStorageBucketInternal, i.cursor, latestCursorValue)
```
mrodm: I would rename this variable to be `cursors`.

Right now they are timestamps, and probably it is not changed in the future. But we could keep that name separated from the implementation.

storage/indexer.go
```diff
@@ -179,18 +193,150 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+	if err != nil {
+		metrics.StorageIndexerUpdateIndexErrorsTotal.Inc()
+		return fmt.Errorf("can't list cursors between %s and %s: %w", i.cursor, latestCursorValue, err)
+	}
+
+	type revision struct {
+		timestamp string
```
mrodm: Same as above.
I would rename this field as `cursor`.

storage/indexer.go
```diff
@@ -179,18 +193,150 @@ func (i *Indexer) updateIndex(ctx context.Context) error {
...
+		delta     *internalStorage.SearchIndexDelta
+		prepared  *preparedDelta
+		fullIndex *packages.Packages
+	}
+
+	revisions := make([]revision, 0, len(timestamps))
+	for _, ts := range timestamps {
```
mrodm: Same here

```go
for _, cursor := range cursors
```

mrodm (approved):
LGTM 🚀 

As it is under a parameter (feature flag), we could test this safely 👍 

Added some suggestions

PR #2040, opened 2026-09-01
Update go runtime to 1.26.7

mrodm:


teresaromero (approved):
👍🏻  does this need a patch release?

mrodm:
> 👍🏻 does this need a patch release?

It could be released in the next release, I think there is no need for a specific patch release now.

PR #2067, opened 2026-09-08
fix: keep Dockerfile Go version in sync with .go-version

### LINKED ISSUE #1906: fix: add default value for ARG GO_VERSION in Dockerfile and a mage target to keep it in sync with .go-version
## Problem

The Dockerfile declares `ARG GO_VERSION` without a default value. Because the `FROM` line on line 9 uses `${BUILDER_IMAGE}:${GO_VERSION}`, Docker's linter emits the following warning when running `mage dockerBuild` or in the CI ([Buildkite build](https://buildkite.com/elastic/package-registry/builds/2350#019fc656-d9a5-4cdf-af7e-9093fdbc5202/L1151)):

```
1 warning found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG ${BUILDER_IMAGE}:${GO_VERSION} results in empty or invalid base image name (line 9)
```

The build itself works because `magefile.go` always injects `--build-arg GO_VERSION=<value>` read from `.go-version`. The warning is purely about the ARG default evaluation.

## Proposed solution

1. **Fix the lint warning** — give `ARG GO_VERSION` a default value in the Dockerfile that mirrors the current content of `.go-version` (currently `1.26.4`).

2. **Keep them in sync** — add a `mage` target (e.g. `UpdateGoVersion`) that:
   - Writes the new version to `.go-version`
   - Updates the `ARG GO_VERSION=<version>` default in `Dockerfile`

   This ensures both files are always updated together and the Dockerfile default never drifts from the actual intended version.

## Why this matters

- Eliminates a persistent build warning that may surface in CI or Docker Scout scans.
- Makes the Dockerfile self-documenting: a reader can see the intended Go version without needing to know that `.go-version` is the source of truth.
- Reduces human error when bumping the Go version (currently requires manual edits in at least two places).

IamPritamAcharya:
## What does this PR do?

Adds a valid default value for `GO_VERSION` in the Dockerfile and introduces a Mage target to keep it synchronized with `.go-version`:

    mage updateGoVersion <version>

The target updates both:

- `.go-version`
- `ARG GO_VERSION=<version>` in `Dockerfile`

It validates the supplied Go version, requires exactly one matching Dockerfile declaration, preserves unrelated Dockerfile content, and is idempotent when the requested version is already configured.

A test also verifies that the committed `.go-version` and Dockerfile default remain synchronized, preventing future Go version updates from introducing drift.

## Why is this needed?

The Dockerfile previously declared:

    ARG GO_VERSION

`GO_VERSION` is used to construct the builder image in a `FROM` instruction. Although `mage dockerBuild` explicitly supplies the version from `.go-version`, Docker still validates the default expansion and reports:

    InvalidDefaultArgInFrom: Default value for ARG ${BUILDER_IMAGE}:${GO_VERSION} results in empty or invalid base image name

Providing a valid default removes this warning while keeping `.go-version` as the source used by the existing Docker build tooling.

The existing explicit `GO_VERSION` build argument and FIPS build behavior remain unchanged.

## Validation

- `go test -tags=mage -run '^TestUpdateGoVersion' .`
- `mage updateGoVersion 1.26.7`
- `mage format`
- `mage check`
- `mage test`
- `mage dockerBuild issue-1906-local`
- `git diff --check`

Fixes #1906

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # This image contains the package-registry binary.
 # It expects packages to be mounted under /packages/package-registry or have a config file loaded into /package-registry/config.yml
 
+# Keep in sync with .go-version.
 ARG GO_VERSION=1.26.7
 ARG BUILDER_IMAGE=golang
 ARG RUNNER_IMAGE=cgr.dev/chainguard/wolfi-base
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # This image contains the package-registry binary.
 # It expects packages to be mounted under /packages/package-registry or have a config file loaded into /package-registry/config.yml
 
-ARG GO_VERSION
+ARG GO_VERSION=1.26.7
 ARG BUILDER_IMAGE=golang
 ARG RUNNER_IMAGE=cgr.dev/chainguard/wolfi-base
 
--- a/magefile.go
+++ b/magefile.go
@@ -7,10 +7,8 @@
 package main
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
-	goversion "go/version"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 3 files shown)


magefile.go
```diff
@@ -135,6 +142,64 @@ func dockerBuild(tag string, fips bool) error {
 	return nil
 }
 
+// UpdateGoVersion updates the Go version used by development tooling and the
+// Docker builder image.
+func UpdateGoVersion(version string) error {
```
teresaromero: i think updating the file could be done manually; we usually do this manually when prepping releases or if we just want to update the version. having a mage command is handy but i don't see the benefts of having all this code here for just that.

magefile_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateGoVersionFilesInSync(t *testing.T) {
```
teresaromero: this will trigger unit tests and if the values are not aligned if will fail as part of a test run. what about adding a function under mage Check so this fails fast as part of the static check?

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This image contains the package-registry binary.
 # It expects packages to be mounted under /packages/package-registry or have a config file loaded into /package-registry/config.yml
 
-ARG GO_VERSION
+ARG GO_VERSION=1.26.7
```
teresaromero, line 5: can we add a small comment indicating this must be aligned with the current .go-version, to act as a reminder

teresaromero (commented):
its aligned with https://github.com/elastic/package-registry/issues/1906; i left a comment for the update mage command, as i am not sure its really needed. 

@mrodm wdyt about removing the mage command?

magefile.go
IamPritamAcharya: Thanks for the review. I’ve left the updater as-is for now since #1906 explicitly proposes a Mage target to keep both files synchronized. Happy to simplify or remove it if @mrodm or you prefer the manual + check-only approach.

magefile_test.go
IamPritamAcharya: Done. I added a read-only consistency check at the beginning of `mage Check()` so a mismatch fails before the rest of the checks. The existing test now calls the same helper to avoid duplicating the validation logic.

Dockerfile
IamPritamAcharya: Done. Added `# Keep in sync with .go-version.` above the `GO_VERSION` declaration.

magefile.go
mrodm: As there is now a check that validates that both `.go-version` and the value in `ARG GO_VERSION` are the same, I think it is not needed to have another target in mage for this.

For now, it would be easier to just update manually the file when it is required.

Please @IamPritamAcharya remove that `UpdateGoVersion` target, and let's keep just the changes to include the validation when it is executed `mage check`.

magefile.go
```diff
@@ -153,6 +248,9 @@ func Check() error {
 }
 
 func Test() error {
+	if err := sh.RunV("go", "test", "-tags=mage", "-run", "^TestUpdateGoVersion", "."); err != nil {
```
mrodm: Should it be removed the filter and run all tests ? In case more tests are added, if that filter is present they won't be executed.

magefile.go
IamPritamAcharya: Done. Removed the `UpdateGoVersion` target and the updater-specific code/tests. The PR now keeps only the Dockerfile default and the consistency validation through `mage check`.

magefile.go
IamPritamAcharya: Done. Removed the test-name filter, so the Mage test step now runs all Mage-tagged tests with `go test -tags=mage .`.

magefile.go
```diff
@@ -135,7 +140,37 @@ func dockerBuild(tag string, fips bool) error {
...
+
+	expected := "ARG GO_VERSION=" + goVersion
+	if string(matches[0]) != expected {
+		return fmt.Errorf("Go version in %s does not match %s: got %q, expected %q", dockerfile, goVersionFile, string(matches[0]), expected)
+	}
+	return nil
+}
+
 func Check() error {
+	if err := checkGoVersionFilesInSync(); err != nil {
```
mrodm: This function could accept both path files as parameters (`.go-version` and `Dockerfile` paths):

```go
func checkGoVersionFilesInSync(goVersionPath, dockerfilePath string) error {
```

Doing so, it would allow us to unit test this function by creating some different temporal files in `TestGoVersionFilesInSync`. You could create some different cases with different contents for both files to test that the function works as expected (whether or not versions are equal or not).

Just an idea here:

```go
func TestGoVersionFilesInSync(t *testing.T) {
	tests := []struct {
		name        string
		goVersion   string
		dockerfile  string
		expectError bool
	}{
		{
			name:       "matching versions",
			goVersion:  "1.26.7",
			dockerfile: "ARG GO_VERSION=1.26.7\n",
		},
		{
			name:        "different versions",
			goVersion:   "1.26.7",
			dockerfile:  "ARG GO_VERSION=1.25.0\n",
			expectError: true,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			dir := t.TempDir()
			goVersionPath := filepath.Join(dir, ".go-version")
			dockerfilePath := filepath.Join(dir, "Dockerfile")

			err := os.WriteFile(goVersionPath, []byte(tt.goVersion+"\n"), 0o644)
			require.NoError(t, err)
			err = os.WriteFile(dockerfilePath, []byte(tt.dockerfile), 0o644)
			require.NoError(t, err)

			err = checkGoVersionFilesInSync(goVersionPath, dockerfilePath)
			if tt.expectError {
				assert.Error(t, err)
				return
			}
			assert.NoError(t, err)
		})
	}
}
```

IamPritamAcharya: Done. The helper now accepts both file paths, and `TestGoVersionFilesInSync` uses temporary files to cover matching, mismatched, empty, missing, and duplicate declaration cases. `mage check` continues to call it with the repository's actual `.go-version` and `Dockerfile` paths.

teresaromero (approved):
thanks!!

mrodm (approved):

Issue #1693, opened 2026-04-29
Make package ordering consistent on the search endpoint


nchaulet:
## Description 

Following https://github.com/elastic/package-registry/pull/1689

Depending of filtering with `?all` or not, the packages are ordered in reverse order. We should avoid this inconsistent behaviour.

Before making this change, we should validate the impact on Kibana.

PR #1335, opened 2025-07-05
Tool to help creating distributions of packages

jsoriano:
The tool makes queries to package registry to discover and download packages.

Users provide a configuration file that includes the set of queries to perform, each query receives parameters that map with query parameters. A matrix can be defined with parameters that should be included in all queries.
Actions can be also defined for each package.

Some small examples:

Download all available packages for 9.1.0, with format versions between 2.3 and 3.4.
```
actions:
  - download:
      destination: ./build/distribution
queries:
  - kibana.version: 9.1.0
    spec.min: 2.3
    spec.max: 3.4
```

Download all available packages for 9.1.0, with format versions between 2.3 and 3.4, including prereleases
```
actions:
  - download:
      destination: ./build/distribution
matrix:
  - {}
  - prerelease: true
queries:
  - kibana.version: 9.1.0
    spec.min: 2.3
    spec.max: 3.4
```

A couple of more complex examples can be found in the PR:
- `lite.yml` downloads the latest version of the packages included in the lite distribution, for each one of the given versions of kibana.
- `sample.yml` downloads the latest version of all packages, including their prereleases.

PR #2109, opened 2026-09-21
feat(distribution): enable standalone cmd/distribution

teresaromero:
Fix https://github.com/elastic/ingest-dev/issues/9119

## Summary

- **Retry + adaptive pacing**: the search phase survives transient registry
  errors. A retrying transport wraps the existing HTTP client and retries
  429/5xx up to 4 times with full-jitter backoff. An adaptive pacer throttles
  all workers when the server signals distress, reducing peak origin memory
  pressure without penalising a healthy registry.
- **Fail-fast collect**: a cancellable context stops remaining search requests
  the moment one fails terminally, preventing retry amplification (80 URLs × 4
  attempts = 320 requests aimed at a server that is already down).
- **Split concurrency**: `searchConcurrency=2`, `downloadConcurrency=4`. Each
  `all=true` response costs ~22 MB at the origin; 2 concurrent requests peak at
  ~44 MB instead of ~90 MB. Downloads are static files and stay at 4.
- **`keep` mechanism**: per-response version window so the tool can fetch the N
  newest versions per package instead of all or one.
- **URL deduplication**: identical search URLs from matrix/query combinations
  are collapsed before dispatch.
- **Standalone module**: `cmd/distribution` can be installed via `go install`
  with no dependency on the parent module.
- **`lite.yaml` updated**: reflects the current `distribution:lite` image
  content (Kibana 8.15–9.5, 18 core packages).
- **`examples/production-slim.yaml`**: documents the `keep:3` production config
  and explains how it differs from an unrestricted all-versions fetch (~2 500
  entries vs ~13 700).

### Why the search phase was failing

Running against `prod.yaml` (103 matrix entries × 1 query, all with `all=true`)
produced ~22 MB response buffers per request at the origin. At `concurrency=4`,
that is ~90 MB of live response data simultaneously — consistent with the origin
hitting its memory ceiling and returning 502s. Three client-side flaws turned
a transient overload into a total build failure:

1. **No retry**: any non-200 was a terminal error; a 502 was indistinguishable
   from a 404.
2. **No fail-fast**: once the registry started failing, all remaining requests
   kept firing at it. Adding retries without fixing this makes it worse: 80
   failing URLs × 4 attempts = 320 additional requests.
3. **No backpressure**: concurrency was the only throttle, shared between search
   and download.

### Resilience changes

**`retry.go`** (new): `retryTransport` wraps `userAgentTransport` using the
transport-composition pattern in `client.go`. Retries on
429/500/502/503/504 and transport errors; never retries `context.Canceled` or
`context.DeadlineExceeded` — this keeps the fail-fast mechanism effective.
Wait uses `Retry-After` when present (seconds or HTTP-date, clamped to 30 s),
otherwise full-jitter exponential backoff. A shared `pacer` starts at zero
interval (no overhead on a healthy registry), doubles per error up to 5 s, and
decays ×3/4 per success. A single 502 slows every worker, not just the one that
saw the error. Retry notices go to stderr so stdout remains pipeable.

**`config.go`**: `collect()` creates a `context.WithCancel` context. The first
terminal failure calls `cancel()`; remaining tasks short-circuit; the URL queue
loop breaks. `context.Canceled` errors are swallowed so only the root-cause
error surfaces. `searchURLs()` is now eager and deduplicates identical URLs using
`mergeKeep`, where unlimited (`keep=0`) wins over any bounded window.

**`client.go`** (new): extracted from `main.go`. Defines `searchConcurrency=2`,
`downloadConcurrency=4`, `searchTimeout=5m`, `downloadTimeout=15m`. Wires the
retry transport into the shared `httpClient`. `searchTimeout` covers up to 4
attempts (30 s `ResponseHeaderTimeout` each) plus 3 backoff sleeps capped at
30 s. Includes `drainAndClose` for connection reuse and `buildUserAgent` for
request attribution.

**`download.go`**: uses the shared `httpClient`; adds per-download context with
`downloadTimeout`; guards against missing `SignaturePath`.

**`main.go`**: action/download pool uses `downloadConcurrency`. Not fail-fast:
a bad download should not abort the remaining ~3 900 files.

**`AGENTS.md`**: documents the two-module layout and the constraint that
`cmd/distribution` must remain independent of the parent module.

### `keep` mechanism

The registry `/search` endpoint returns either one version per package (default)
or all versions (`all=true`). The `keep` field fills the gap:

- `keep: N` limits results to the N newest versions per package per search
  response; `N > 1` automatically adds `all=true` on the wire.
- Applied per response (one matrix entry × one query) so each Kibana release
  gets its own N newest installable versions before results are merged.
- `keep: 0` (default) keeps everything, preserving existing behaviour.

### Standalone module

`cmd/distribution` previously imported the parent module via a `replace`
directive, breaking `go install`. The only dependency was `workers.TaskPool`,
now copied into `cmd/distribution/internal/workers`. `go install` from any
commit or `@latest` works correctly.

### `lite.yaml`

Updated to match `docker.elastic.co/package-registry/distribution:lite`:
removed stale Kibana versions, added new patch releases across 8.15–8.19 and
9.0–9.5, corrected missing `spec.max` constraints, kept only the 18 packages
matching the current lite image. Old 7.x entries are commented out (preserved
for reference). A small set of old versions required by Kibana's Fleet API
integration tests is pinned via the `packages:` block.

## Changes

| File | What changed |
|---|---|
| `cmd/distribution/retry.go` | New: `retryTransport`, `pacer`, backoff logic, `parseRetryAfter`, `sleepCtx` |
| `cmd/distribution/retry_test.go` | New: full retry/pacer/backoff test suite |
| `cmd/distribution/client.go` | New: extracted HTTP client setup; `searchConcurrency=2`, `downloadConcurrency=4`, timeouts, retry transport wiring, `drainAndClose`, `buildUserAgent` |
| `cmd/distribution/client_test.go` | New: user-agent and connection-reuse tests |
| `cmd/distribution/config.go` | `collect()` fail-fast context; `searchURLs()` deduplication; `mergeKeep`; `keep` field; `truncateVersions`; `keepFor`; `comparePackageInfo`/`compareVersions` factored out |
| `cmd/distribution/download.go` | Uses shared `httpClient`; per-request `downloadTimeout` context; `SignaturePath` guard |
| `cmd/distribution/main.go` | Action pool uses `downloadConcurrency`; uses shared `httpClient` |
| `cmd/distribution/main_test.go` | Added `TestConfigCollectFailFast`, `TestConfigCollectRetriesTransientError`, `TestConfigSearchURLsDeduplicates`, and keep-related tests |
| `cmd/distribution/examples/production-slim.yaml` | New: production config with `keep:3` annotation explaining difference from unrestricted all-versions fetch |
| `cmd/distribution/examples/lite.yaml` | Updated Kibana matrix and packages; added Fleet pinned-versions block |
| `cmd/distribution/internal/workers/taskpool.go` | Copied from parent module; no functional change |
| `cmd/distribution/internal/workers/taskpool_test.go` | Copied from parent module |
| `cmd/distribution/go.mod` | Removed parent-module require and replace |
| `cmd/distribution/README.md` | Documents `keep`, standalone install, and examples |
| `AGENTS.md` | Documents two-module layout and `cmd/distribution` independence constraint |
| `.github/workflows/post-dependabot.yml` | Removed stale workflow |

## Test plan

- [x] `go test ./...` inside `cmd/distribution` passes (51 test functions).
- [ ] `go install github.com/elastic/package-registry/cmd/distribution@<commit>` succeeds from a clean module cache.
- [ ] `go run . examples/prod.yaml 2>&1 | tee /tmp/prod-run.txt` completes without `failed to collect packages`; retry notices visible on stderr if EPR wobbles.
- [ ] `go run . examples/lite.yaml` and `go run . examples/sample.yaml` produce expected request volumes and results.
- [x] End-to-end CI test via elastic/package-storage-infra#1280: builds both `prod` and `lite` distributions against the live registry and pushes results to a testing repository.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/post-dependabot.yml
teresaromero, line 1: This file is removed as i removed the dependency between this command and package-registry

mrodm, line 1: Why removing this workflow ? 
This workflow allows to update PRs created by dependabot that require changes in `cmd/distribution`.

mrodm: Ok, seen that the `replace` directive has been removed.

cmd/distribution/retry.go
mrodm, line 1: Could it be used here the same retry library that uses package-registry or elastic-package ? 

In package-registry and elastic-package it is used: [github.com/hashicorp/go-retryablehttp](https://github.com/hashicorp/go-retryablehttp)

- https://github.com/elastic/package-registry/blob/a83e687bb76c19bddb2f3938a6f95ee993909729/proxymode/proxymode.go#L57-L72

- https://github.com/elastic/elastic-package/blob/446b7cd0efac4acc7d9764e60d9bd4daaaf68001/internal/retry/http.go#L35

cmd/distribution/client.go
```diff
@@ -0,0 +1,95 @@
...
+// sustained distress the adaptive pacer reduces this further automatically.
+// Deliberately a constant and not a config knob: it exists to protect the
+// registry, not to be tuned per run.
+const searchConcurrency = 2
+
+// downloadConcurrency bounds how many package downloads are in flight at once.
+// Downloads are static files served from object storage, so a higher
+// concurrency than search is safe.
+// Deliberately a constant and not a config knob.
+const downloadConcurrency = 4
```
mrodm, line 26: Why not using the same number of workers as before ? `runtime.GOMAXPROCS(0)` ? 

Maybe it could be divided by 2 to not use all the processors if needed?

cmd/distribution/go.mod
```diff
@@ -25,5 +24,3 @@ require (
 	golang.org/x/sys v0.48.0 // indirect
 	golang.org/x/telemetry v0.0.0-20260908163034-4bcc4b2ee518 // indirect
 )
-
-replace github.com/elastic/package-registry => ../..
```
mrodm, line 29: 👍

cmd/distribution/config.go
```diff
@@ -144,29 +222,65 @@ func (c config) collect(client *http.Client) ([]packageInfo, error) {
...
+
+// truncateVersions keeps at most keep newest versions of each package in
+// packages, reordering and compacting it in place. A keep of zero or less
+// keeps everything. Versions that are not valid semantic versions sort oldest
+// and are dropped first.
+//
+// The window is per search response on purpose: each matrix entry is a Kibana
+// version that needs its own installable versions, so the limit is applied
+// before responses are merged.
+func truncateVersions(packages []packageInfo, keep int) []packageInfo {
```
mrodm, line 265: Keep in mind that [epr.elastic.co](https://epr.elastic.co) is returning the packages in the search responses with the older versions first. To validate that this is working as expected 

As example:
```shell
 $ curl -s "https://epr.elastic.co/search?package=apache&all=true" | jq -r '.[] | .version'
1.1.0
1.3.0
1.3.2
1.3.5
1.4.1
1.5.0
...
2.0.0
2.0.1
2.1.0
2.1.1
3.0.0
3.0.1
3.0.2
```


Related to:
- (semver) https://github.com/elastic/package-registry/pull/1585
- (name vs title) https://github.com/elastic/package-registry/pull/1689

mrodm: There is an open issue that could affect this https://github.com/elastic/package-registry/issues/1693

If the logic in `cmd/distribution` orders the packages with their own ordering, there should be no dependency on that. 

In case that could affect here, it would be good at least to add a note about this issue so the order of the packages in the response could change if that issue is fixed/solved.

cmd/distribution/examples/production-slim.yaml
```diff
@@ -0,0 +1,290 @@
...
+    spec.max: 3.6
+  - kibana.version: 9.5.2
+    spec.min: 2.3
+    spec.max: 3.6
+  - kibana.version: 9.5.3
+    spec.min: 2.3
+    spec.max: 3.6
+  - kibana.version: 9.5.4
+    spec.min: 2.3
+    spec.max: 3.6
```
mrodm, line 278: Currently, 9.6.0-SNAPSHOT (and others like 9.4.8-SNAPSHOT, 9.5.5-SNAPSHOT) should be included here.

For instance, there could be packages that already contain their kibana constraint to be `^9.6.0`.

Some examples:
- https://github.com/elastic/integrations/blob/83e255bb8824028fd50c1b63378b8605bf5ed388/packages/aws/manifest.yml#L24

- https://github.com/elastic/integrations/blob/83e255bb8824028fd50c1b63378b8605bf5ed388/packages/auditd/manifest.yml#L17

cmd/distribution/README.md
```diff
@@ -54,16 +70,40 @@ queries:
...
+file are downloaded unconditionally — signature verification runs on every package.
+
+See `examples/pinned.yaml` for a self-contained example.
+
 ## Actions
 
 - **print**: Output package names and versions to console
 - **download**: Download package ZIP files and signatures
   - `destination`: Target directory for downloads
-  - `validate`: Verify package signatures using GPG
```
mrodm, line 65: Was this not implemented at all?
I don't see any logic removed in this PR about this.

cmd/distribution/README.md
```diff
@@ -54,16 +70,40 @@ queries:
...
 
 ## Dependencies
 
-Managed via Go modules. Run `go mod tidy` to update dependencies.
+`cmd/distribution` is a self-contained Go module with no dependency on the
+parent `github.com/elastic/package-registry` module. The concurrency primitive
+it needs (`internal/workers`) is a local copy, allowing `go install` to work
+without any `replace` directives. Run `go mod tidy` inside this directory to
+update dependencies.
```
mrodm, line 109: There are also some mage targets for this (defined in the root of the working copy):
- mage modTidy
- mage buildDistribution
- mage check
- mage test

cmd/distribution/config.go
```diff
@@ -22,19 +22,40 @@ import (
...
+	"github.com/elastic/package-registry/cmd/distribution/internal/workers"
 )
 
 type config struct {
-	Address  string          `yaml:"address"`
+	Address string `yaml:"address"`
+	// Keep is the default newest-N window per package per search response.
+	// Zero keeps everything. Values > 1 force all=true on the wire.
+	// Individual matrix entries and queries can override this.
+	Keep     int             `yaml:"keep"`
```
mrodm, line 33: It looks like that this `keep` field is not used in the YAML example files.

```yaml
# No package filter — all packages compatible with each matrix entry are fetched.
# keep: 3 retains the 3 newest versions per package per matrix entry before results are merged.
queries:
  - keep: 3
```

Is it needed here too? or should it be just kept the one from `configQuery` struct ? 

Could that be replaced by the following ? 
```yaml
keep: 3
```

mrodm (commented):
Looks interesting to add the `keep` option even if it is not part of the package-registry API.

cmd/distribution/config.go
```diff
@@ -53,30 +74,64 @@ func (c config) searchURLs() (iter.Seq[*url.URL], error) {
...
+			keep := c.keepFor(m, q)
+			values := m.Build()
+			for k, v := range q.Build() {
+				values[k] = v
+			}
+			if keep > 1 {
+				// The registry returns a single version per package unless all is set,
+				// and it has no way to limit the result size.
+				values.Set("all", "true")
+			}
```
mrodm, line 96: should keep== 0 add also `all=true` as part of the query ?

cmd/distribution/internal/workers/taskpool.go
```diff
@@ -0,0 +1,55 @@
...
+
+	p.errLock.Lock()
+	p.errors = append(p.errors, err)
+	p.errLock.Unlock()
+}
+
+// Wait waits for all the tasks to finish, and joins the errors found. The pool cannot be used after calling Wait.
+func (p *taskPool) Wait() error {
+	close(p.pool)
+	p.wg.Wait()
```
mrodm, line 53: Just got this from `claude`:

> Wait()` closes the semaphore channel before `wg.Wait()` completes; any task goroutine that calls `Do()` on the same pool will `send on closed channel` and panic.

Could you double check this ? Maybe it is required to update workers in package-registry too?

cmd/distribution/config.go
teresaromero: the command has its own sorting from "all" https://github.com/elastic/package-registry/pull/2109/changes#diff-d713c9b59cf81717693f8f49c8a3f96a7a6ca658d5293cef45597066314b5863R272

cmd/distribution/client.go
teresaromero: This value is used to determine how many workers we use in the pool. Not sure why this was elegible for it in the first place as i went back to the pr and did not see any discussion around it. In practice this value is dependent on available CPU; what we need to control is how many requests are handled by the server (epr) which is not related to cpu of the machine performing the request. 

I prefer to have this determined and tuned as convenient to control how many request the server can handle. 

From claude:

> ## What `runtime.GOMAXPROCS(0)` actually is
> 
> `GOMAXPROCS(0)` is a **read** (argument ≤ 0 returns the current value without changing it).
> It returns the Go scheduler's cap on simultaneously-running OS threads.
> 
> **Default value at startup:**
> 1. `GOMAXPROCS=N` env var, if set
> 2. Otherwise: `min(sched_getaffinity CPU count, cgroup CPU quota/period)`, with a floor of 2 for small quotas (Linux + Go 1.25+)
> 3. On macOS and everything else: just the logical CPU count
> 
> **Practical values for this tool:**
> 
> | Environment | GOMAXPROCS | Workers (old) |
> |---|---|---|
> | Dev laptop, 10-core M1 | 10 | 10 |
> | Buildkite agent, 4 vCPU | 4 | 4 |
> | Container, 2 CPU limit | 2 | 2 |
> | Container, 0.5 CPU quota | 2 (cgroup floor) | 2 |
> | Single-vCPU VM / `taskset -c 0` | 1 | 1 |
> | `GOMAXPROCS=2` env set | 2 | 2 |
> 
> **What makes it change:** It's per-process and knows nothing about other processes. Go 1.25+ also updates it dynamically when cgroup CPU limits change, so it is not even a stable number within one invocation.

cmd/distribution/config.go
teresaromero: The intention is to have both availabe, a per query and a global one. I am updating the production lite example so it uses the global value

cmd/distribution/config.go
teresaromero: 🤔 if keep is "false" (empty, or 0 value) i don't think we should trigger the all query under the hood; as is not explicit to keep a given number

cmd/distribution/README.md
teresaromero: This was initially planned to be part of the config; documenting it on the readme. There was a fix after initial implementation which included the verification on all downloads; so this is not opt-in setting. 

I am removing it here to aling what the code is capable to do


- **PR #1335** — `500b635c` "Tool to help creating distributions of packages" — initial implementation, included the README's `validate` doc but no actual config field in code; newly downloaded packages were not verified.
- **PR #1645** — `fde252fe` "Check signature of packages after downloading them in distribution tool" — the bug fix that made verification run on freshly downloaded packages, not just cached ones.

cmd/distribution/README.md
teresaromero: Updated the doc using this commands

cmd/distribution/retry.go
teresaromero: I've used this suggestion and used the library for retries

teresaromero (commented):
updated branch and comments. Thanks!!

cmd/distribution/internal/workers/taskpool.go
teresaromero: fixed and also added an entry on changelog as it involved package-registry too

cmd/distribution/examples/production-slim.yaml
```diff
@@ -0,0 +1,291 @@
...
+#   still guaranteeing that every Kibana release in the matrix has at least 3
+#   installable versions of each package available offline.
+#
+
+# Base address of the Package Registry
+address: "https://epr.elastic.co"
+
+# Document-wide default: keep the 3 newest versions of every package per matrix entry.
+# Individual queries can override this with their own keep: N; 0 (the default) keeps everything.
+keep: 3
```
mrodm: Currently EPR does not support this as a query parameter. But if at some point we would need to add this parameter to the search query in EPR, I think it would be better to rename it. `keep` field name for instance does not show the concept of returning the latest N versions of each package.

I think in some other options:
- versions , e.g. `/search?package=apache&versions=3&prerelease=true`
- version.limit , e.g. `/search?package=apache&version.limit=3&prerelease=true`
- package.versions , e.g. `/search?package=apache&package.versions=3&prerelease=true`

cmd/distribution/internal/workers/taskpool_test.go
```diff
@@ -0,0 +1,103 @@
...
+		return err2
+	})
+
+	err := pool.Wait()
+	if err != nil {
+		// At least one error should be present
+		errMsg := err.Error()
+		hasErr1 := strings.Contains(errMsg, "error 1")
+		hasErr2 := strings.Contains(errMsg, "error 2")
+		require.True(t, hasErr1 || hasErr2, "should contain at least one error")
```
mrodm: Should this require validate that `hasErr1 && hasErr2` ?

cmd/distribution/download.go
```diff
@@ -85,10 +97,18 @@ func (a *downloadAction) download(urlPath string) error {
 	if err != nil {
 		return fmt.Errorf("failed to open %s in %s: %w", path.Base(urlPath), a.Destination, err)
 	}
-	defer f.Close()
+	if _, err = io.Copy(f, resp.Body); err != nil {
+		f.Close()
+		return err
+	}
```
mrodm: Should it be added `os.Remove(f.Name())` here in case of error ? 

To avoid leaving partial files on disk ?

cmd/distribution/retry.go
```diff
@@ -0,0 +1,89 @@
...
+					d.Seconds(), resp.StatusCode)
+				return d
+			}
+		}
+	}
+	cap := min(minWait<<attemptNum, maxWait)
+	var wait time.Duration
+	if cap > 0 {
+		wait = rand.N(cap)
+	}
```
mrodm: From claude review, 

> `time.Duration` is `int64`; at `attemptNum >= 34`, `minWait << attemptNum` overflows to negative, `min(negative, maxWait)` returns negative, the `> 0` guard sets `wait = 0`, and retries become a tight loop with no delay.

Currently, the number of attempts is 7 and it will not reach that. But it could be added a guard here to avoid this in the future if `maxAttempts` is ever raised above 34.

cmd/distribution/client.go
mrodm: Agree that it is interesting to have those values as default ones to be able to manage how many requests could handle the server.

But it also looks interesting to add configuration or settings for those in case we want to test new values without re-building the tool.

No need to do it now, but I think it could be interesting to allow overwriting those values.

